PR #20, opened 2025-08-08
Fix client performance.

Yury-Fridlyand:
See discussion in https://github.com/valkey-io/valkey-glide/issues/3563#issuecomment-2788053881

Cluster client was failing `ConcurrentOperationsWork` test having only 1 thread in tokio thread pool (request timeouts). After increasing them to 1 sec (default = 250ms) timeouts disappeared, but it is not a fix.
With bigger thread pool no tests fail with default timeout. See full matrix CI run: https://github.com/valkey-io/valkey-glide/actions/runs/16629098022


yipin-chen (approved):

jbrinkman (approved):
:shipit:

PR #12, opened 2025-08-05
CD pipeline - phase 1

Yury-Fridlyand:
Replacement for https://github.com/valkey-io/valkey-glide/pull/4459

Partially fixes #1

PR #21, opened 2025-08-08
API compatibility workflow (dotnet-api-diff): net8.0 HTML report

jbrinkman:
Adds a GitHub Actions workflow to compare Valkey.Glide vs StackExchange.Redis using dotnet-api-diff.\n\n- Installs dotnet-api-diff via curl-based script\n- Uses config at .github/api-compat/api-diff-config.json (namespace/type mappings, breaking-change rules)\n- Builds and compares net8.0 only\n- Emits clean HTML report and uploads as artifact\n- Fails on breaking changes


.github/workflows/api-compatibility.yml
```diff
@@ -0,0 +1,127 @@
...
+                  fi
+
+                  set +e
+                  dotnetapidiff compare \
+                    "$SER_DLL" \
+                    "$VALKEY_DLL" \
+                    --config ".github/api-compat/api-diff-config.json" \
+                    --output html \
+                    --output-file "$OUTFILE"
+                  CODE=$?
```
Yury-Fridlyand, line 102: if you remove `set +e`, GHA will fail the pipeline if exit code is not 0. So you can remove lines 102-106 and the last step too

.github/workflows/api-compatibility.yml
```diff
@@ -0,0 +1,127 @@
...
+
+                  echo "Using StackExchange.Redis version: ${SER_VERSION}"
+                  echo "SER DLL: ${SER_DLL}"
+                  echo "Valkey DLL: ${VALKEY_DLL}"
+
+                  if [[ ! -f "$SER_DLL" ]]; then
+                    echo "# API Diff Error (net8.0)" > "$OUTFILE"
+                    echo "Could not find StackExchange.Redis DLL at $SER_DLL" >> "$OUTFILE"
+                    echo "exit_code=6" >> "$GITHUB_OUTPUT"
+                    exit 0
```
Yury-Fridlyand, line 86:
```suggestion
                    exit 1
```
pipeline failed - don't suppress it

.github/workflows/api-compatibility.yml
```diff
@@ -0,0 +1,127 @@
...
+                  fi
+                  if [[ ! -f "$VALKEY_DLL" ]]; then
+                    echo "# API Diff Error (net8.0)" > "$OUTFILE"
+                    echo "Could not find Valkey.Glide DLL at $VALKEY_DLL" >> "$OUTFILE"
+                    echo "exit_code=6" >> "$GITHUB_OUTPUT"
```
Yury-Fridlyand, line 91: same

.github/workflows/api-compatibility.yml
```diff
@@ -0,0 +1,127 @@
...
+
+            - name: Download StackExchange.Redis package (${{ env.SER_VERSION }})
+              shell: pwsh
+              run: |
+                  nuget install StackExchange.Redis -Version "$env:SER_VERSION" -OutputDirectory ./packages
+
+            - name: Build Valkey.Glide (net8.0)
+              shell: pwsh
+              run: |
+                  dotnet build --configuration Release --framework net8.0 --nologo
```
Yury-Fridlyand, line 65:
```suggestion
                  dotnet build --configuration Lint --framework net8.0 --nologo
```
to skip building rust part
(or set `SkipCargo` env var after merging #12)

.github/workflows/api-compatibility.yml
```diff
@@ -0,0 +1,127 @@
...
+              with:
+                  submodules: true
+
+            - name: Setup .NET
+              uses: actions/setup-dotnet@v4
+              with:
+                  dotnet-version: |
+                      8.x
+
+            # Install Rust toolchain on Windows for cargo prebuild
```
Yury-Fridlyand, line 35: uncomment or delete?

.github/workflows/api-compatibility.yml
jbrinkman: I do not want lint build here. We have separate builds that focus on linting. This only exists to give me the assembly to test.

.github/workflows/api-compatibility.yml
jbrinkman: Was in the middle of testing as the pipeline was failing. Will remove now that everything is working and is def not needed.

.github/workflows/api-compatibility.yml
jbrinkman: Lint did exactly what I thought it might do, it failed because of linting errors. Reverting back to a release build.

PR #22, opened 2025-08-08
Update FFI delegates.

Yury-Fridlyand:
Read about `LibraryImport`: [one](https://learn.microsoft.com/en-us/dotnet/standard/native-interop/pinvoke-source-generation) [two](https://learn.microsoft.com/en-us/dotnet/standard/native-interop/calling-conventions) [three](https://learn.microsoft.com/en-us/dotnet/standard/native-interop/best-practices) and [examples](https://github.com/search?q=%5BLibraryImport&type=code)

UPD
This was mostly done to address a code style analyzer notice. As far I understand things, it can provide better performance in a multi-thread application. Performance wasn't measured.

### DIFF
--- a/sources/Valkey.Glide/Internals/FFI.methods.cs
+++ b/sources/Valkey.Glide/Internals/FFI.methods.cs
@@ -1,0 +1,47 @@
+﻿// Copyright Valkey GLIDE Project Contributors - SPDX Identifier: Apache-2.0
+
+// check https://stackoverflow.com/a/77455034 if you're getting analyzer error (using is unnecessary)
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Valkey.Glide.Internals;
+
+internal partial class FFI
+{
+#if NET8_0_OR_GREATER
+    [LibraryImport("libglide_rs", EntryPoint = "command")]
+    [UnmanagedCallConv(CallConvs = new[] { typeof(CallConvCdecl) })]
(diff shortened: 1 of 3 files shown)


sources/Valkey.Glide/Internals/FFI.methods.cs
```diff
@@ -0,0 +1,47 @@
...
+
+// check https://stackoverflow.com/a/77455034 if you're getting analyzer error (using is unnecessary)
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Valkey.Glide.Internals;
+
+internal partial class FFI
+{
+#if NET8_0_OR_GREATER
```
jbrinkman, line 11: Why do we not create these FFI methods for .Net 6?

jduo:
This description needs more detail (why it's done this way, does this only work with .NET 8.0+, is there an alternative implementation that needs to be done for older), what this does.

Yury-Fridlyand:
This was mostly done to address a code style analyzer notice. As far I understand things, it can provide better performance in a multi-thread application. Performance wasn't measured.

jduo:
Thanks. 

Let's include the info about the static analyzer warning in the PR description and commit message.

Lgtm otherwise.

jduo (approved):

alexr-bq (approved):

jbrinkman (approved):
:shipit:

PR #63, opened 2025-08-28
HFE Commands

edlng:
This PR adds HFE commands. I only got HFE Commands to an all tests passed state.

**WARNING: I have not reviewed it yet so there may be inconsistencies or mistakes**

### DIFF
--- a/sources/Valkey.Glide/Commands/IHashCommands.cs
+++ b/sources/Valkey.Glide/Commands/IHashCommands.cs
@@ -1,5 +1,7 @@
 // Copyright Valkey GLIDE Project Contributors - SPDX Identifier: Apache-2.0
 
+using Valkey.Glide.Commands.Options;
+
 namespace Valkey.Glide.Commands;
 
 /// <summary>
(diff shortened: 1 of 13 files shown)


sources/Valkey.Glide/Commands/IHashCommands.cs
```diff
@@ -366,4 +368,327 @@ public interface IHashCommands
     /// </example>
     /// </remarks>
     Task<HashEntry[]> HashRandomFieldsWithValuesAsync(ValkeyKey key, long count, CommandFlags flags = CommandFlags.None);
+
+    // TODO: ALIGN HASH FIELD EXPIRE COMMANDS WITH SER AFTER SER IMPLEMENTS THEM
```
affonsov, line 372: If the verification was done, remove the comment

sources/Valkey.Glide/Commands/Options/HashSetExOptions.cs
```diff
@@ -0,0 +1,123 @@
...
+/// </remarks>
+public class HashSetExOptions
+{
+    /// <summary>
+    /// Expiration settings for the hash fields.
+    /// </summary>
+    public ExpirySet? Expiry { get; set; }
+
+    /// <summary>
+    /// When true, only set fields if none of them exist.
```
affonsov, line 19: nit: For me it is difficult to track if this is the FNX | FXX. It would be good if the acronym be in the comment as well. This apply for all the other comments, but this it is your call.

jbrinkman: We have not mapped option sets to abbreviated commandline arguments anywhere else in documentation. Client libraries intentionally hide the implementation details of the underlying protocols as an individual API call may involve multiple Valkey/Redis commands.

sources/Valkey.Glide/Internals/Request.HashCommands.cs
```diff
@@ -167,4 +168,317 @@ public static Cmd<object[], HashEntry[]> HashRandomFieldsWithValuesAsync(ValkeyK
...
+    public static Cmd<object[], ValkeyValue[]?> HashGetExAsync(ValkeyKey key, ValkeyValue[] fields, HashGetExOptions options)
+    {
+        List<GlideString> args = [key.ToGlideString()];
+
+        // Add expiry options before FIELDS keyword
+        if (options.Expiry != null)
+        {
+            switch (options.Expiry.Type)
+            {
+                case HGetExExpiryType.Seconds:
```
xShinnRyuu, line 180: Quick question: What are the criteria for the mapping of the case `HGetExExpiryType` to the Constant key values when adding to the `args` list?

sources/Valkey.Glide/Commands/Options/HashFieldExpirationConditionOptions.cs
```diff
@@ -0,0 +1,54 @@
...
+
+/// <summary>
+/// Expiration condition options for hash field expiration commands.
+/// </summary>
+public enum ExpireOptions
+{
+    /// <summary>
+    /// Set expiration only if the field has no expiration.
+    /// </summary>
+    HAS_NO_EXPIRY,
```
jeremyprime, line 38: Enums in C# usually use upper camel case (and a quick check of other enums in this repo shows the same).

Issue #50, opened 2025-08-27
C#: Generic Commands


yipin-chen:
### Description

Generic commands support for C# 

Full list is listed in https://github.com/valkey-io/valkey-glide/wiki/ValKey-Commands-Implementation-Progress

## Commands

- [x] `expireat` -
- [x] `expiretime`
- [x] `object encoding`
- [x] `object freq`
- [x] `object idletime`
- [x] `object refcount`
- [x] `pexpire`
- [x] `pexpireat`
- [x] `pexpiretime`
- [x] `pttl`
- [x] `randomkey`
- [x] `renamenx`
- [x] `scan`
- [x] `sort`
- [x] `sort_ro`
- [x] `unlink`
- [x] `wait`

PR #89, opened 2025-09-26
Generic Commands

alexr-bq:
Issue: https://github.com/valkey-io/valkey-glide-csharp/issues/50

Adds support for commands while :
expireat
expiretime
object encoding
object freq
object idletime
object refcount
pexpire
pexpireat
pexpiretime
pttl
randomkey
scan
sort
sort_ro
wait

Support already existed:
unlink
renamenx


jbrinkman (approved):
:Shipit:

PR #70, opened 2025-09-08
Temporary workaround: disable SelectAsync and ClientSetNameAsync (see valkey-glide#4691)

jbrinkman:
This PR applies a temporary solution to disable SelectAsync and ClientSetNameAsync as discussed in https://github.com/valkey-io/valkey-glide/issues/4691. See issue #69 in this repo for tracking the workaround.

PR #90, opened 2025-10-02
Valkey-9 MultiDB support

alexr-bq:
- Adds support for database ID with cluster client
- Moves Move/Copy commands from standalone to shared client


alexr-bq:
> It's missing the select command, which is basically reverting this pr for the select #70. The only difference is that select should only be available for standalone batch and not cluster batch

It seems like Select functionality is already in. Pushed a commit that removes it from cluster batch.

affonsov (approved):

yipin-chen (approved):

PR #107, opened 2025-10-14
Geospatial Commands

alexr-bq:
Implements Geospatial commands

### DIFF
--- a/tests/Valkey.Glide.IntegrationTests/GeospatialCommandTests.cs
+++ b/tests/Valkey.Glide.IntegrationTests/GeospatialCommandTests.cs
@@ -247,8 +247,7 @@ public class GeospatialCommandTests(TestConfiguration config)
     public async Task GeoSearch_AllUnits_ReturnsConsistentResults(BaseClient client)
     {
         string key = Guid.NewGuid().ToString();
-        GeoEntry[] entries =
-        [
+        GeoEntry[] entries = [
             new GeoEntry(13.361389, 38.115556, "Palermo"),
             new GeoEntry(15.087269, 37.502669, "Catania"),
             new GeoEntry(12.758489, 38.788135, "Trapani")
@@ -295,14 +294,13 @@ public class GeospatialCommandTests(TestConfiguration config)
     public async Task GeoHash_NonExistentMembers_ReturnsNulls(BaseClient client)
     {
         string key = Guid.NewGuid().ToString();
-        GeoEntry[] entries =
-        [
+        GeoEntry[] entries = [
             new GeoEntry(13.361389, 38.115556, "Palermo"),
             new GeoEntry(15.087269, 37.502669, "Catania")
         ];
         await client.GeoAddAsync(key, entries);
 
-        string?[] hashes = await client.GeoHashAsync(key, new ValkeyValue[] { "Palermo", "Catania", "NonExistent" });
+        string?[] hashes = await client.GeoHashAsync(key, ["Palermo", "Catania", "NonExistent"]);
         Assert.NotNull(hashes[0]); // Palermo
         Assert.NotNull(hashes[1]); // Catania
         Assert.Null(hashes[2]); // NonExistent
@@ -315,7 +313,7 @@ public class GeospatialCommandTests(TestConfiguration config)
         string key = Guid.NewGuid().ToString();
         await client.GeoAddAsync(key, 13.361389, 38.115556, "Palermo");
 
-        GeoPosition?[] positions = await client.GeoPositionAsync(key, new ValkeyValue[] { "Palermo", "NonExistent" });
+        GeoPosition?[] positions = await client.GeoPositionAsync(key, ["Palermo", "NonExistent"]);
         Assert.NotNull(positions[0]); // Palermo
         Assert.Null(positions[1]); // NonExistent
     }
@@ -409,11 +407,11 @@ public class GeospatialCommandTests(TestConfiguration config)
         ];
         await client.GeoAddAsync(key, entries);
 
-        string?[] hashes = await client.GeoHashAsync(key, new ValkeyValue[] { "Palermo", "Catania" });
+        string?[] hashes = await client.GeoHashAsync(key, ["Palermo", "Catania"]);
         Assert.NotNull(hashes[0]);
         Assert.NotNull(hashes[1]);
-        Assert.NotEmpty(hashes[0]);
-        Assert.NotEmpty(hashes[1]);
+        Assert.NotEmpty(hashes[0]!);
+        Assert.NotEmpty(hashes[1]!);
     }
 
     [Theory(DisableDiscoveryEnumeration = true)]
@@ -445,7 +443,7 @@ public class GeospatialCommandTests(TestConfiguration config)
         string key = Guid.NewGuid().ToString();
         await client.GeoAddAsync(key, 13.361389, 38.115556, "Palermo");
 
-        string?[] hashes = await client.GeoHashAsync(key, new ValkeyValue[] { });
+        string?[] hashes = await client.GeoHashAsync(key, []);
         Assert.Empty(hashes);
     }
 
@@ -476,7 +474,7 @@ public class GeospatialCommandTests(TestConfiguration config)
         ];
         await client.GeoAddAsync(key, entries);
 
-        GeoPosition?[] positions = await client.GeoPositionAsync(key, new ValkeyValue[] { "Palermo", "Catania" });
+        GeoPosition?[] positions = await client.GeoPositionAsync(key, ["Palermo", "Catania"]);
         Assert.NotNull(positions[0]);
         Assert.NotNull(positions[1]);
         Assert.True(Math.Abs(positions[0]!.Value.Longitude - 13.361389) < 0.001);
@@ -694,7 +692,6 @@ public class GeospatialCommandTests(TestConfiguration config)
         Assert.NotEmpty(distResults);
 
         var palermoDist = distResults.FirstOrDefault(r => r.Member.ToString() == "Palermo");
-        Assert.NotNull(palermoDist);
         Assert.Equal("Palermo", palermoDist.Member.ToString());
         Assert.NotNull(palermoDist.Distance); // Should have distance
         Assert.Equal(0.0, palermoDist.Distance.Value, 1); // Distance from itself should be ~0
@@ -705,7 +702,6 @@ public class GeospatialCommandTests(TestConfiguration config)
         Assert.NotEmpty(coordResults);
 
         var palermoCoord = coordResults.FirstOrDefault(r => r.Member.ToString() == "Palermo");
-        Assert.NotNull(palermoCoord);
         Assert.Equal("Palermo", palermoCoord.Member.ToString());
         Assert.True(palermoCoord.Position.HasValue); // Should have coordinates
         Assert.Null(palermoCoord.Distance); // Should be null without WithDistance
@@ -729,8 +725,6 @@ public class GeospatialCommandTests(TestConfiguration config)
         var palermoResult = results.FirstOrDefault(r => r.Member.ToString() == "Palermo");
         var cataniaResult = results.FirstOrDefault(r => r.Member.ToString() == "Catania");
 
-        Assert.NotNull(palermoResult);
-        Assert.NotNull(cataniaResult);
         Assert.NotNull(palermoResult.Distance);
         Assert.NotNull(cataniaResult.Distance);
 
@@ -812,13 +806,11 @@ public class GeospatialCommandTests(TestConfiguration config)
         var results = await client.SortedSetRangeByRankWithScoresAsync(destinationKey, 0, -1);
         Assert.Equal(2, results.Length);
 
-        var palermoResult = results.FirstOrDefault(r => r.Key.ToString() == "Palermo");
-        var cataniaResult = results.FirstOrDefault(r => r.Key.ToString() == "Catania");
+        var palermoResult = results.FirstOrDefault(r => r.Element.ToString() == "Palermo");
+        var cataniaResult = results.FirstOrDefault(r => r.Element.ToString() == "Catania");
 
-        Assert.NotNull(palermoResult);
-        Assert.NotNull(cataniaResult);
-        Assert.Equal(0.0, palermoResult.Value, 0.1);
-        Assert.Equal(166.2742, cataniaResult.Value, 0.1);
+        Assert.Equal(0.0, palermoResult.Score, 0.1);
+        Assert.Equal(166.2742, cataniaResult.Score, 0.1);
     }
 
     [Theory(DisableDiscoveryEnumeration = true)]
@@ -899,8 +891,7 @@ public class GeospatialCommandTests(TestConfiguration config)
         string sourceKey = keyPrefix + ":source";
         string destinationKey = keyPrefix + ":dest";
 
-        GeoEntry[] entries =
-        [
+        GeoEntry[] entries = [
             new GeoEntry(13.361389, 38.115556, "Palermo"),
             new GeoEntry(15.087269, 37.502669, "Catania"),
             new GeoEntry(12.758489, 38.788135, "Trapani"),
@@ -932,7 +923,7 @@ public class GeospatialCommandTests(TestConfiguration config)
         ];
         await client.GeoAddAsync(sourceKey, entries);
 
-        await client.SortedSetAddAsync(destinationKey, new SortedSetEntry[] { new SortedSetEntry("OldMember", 100) });
+        await client.SortedSetAddAsync(destinationKey, [new SortedSetEntry("OldMember", 100)]);
         Assert.Equal(1, await client.SortedSetCardAsync(destinationKey));
 
         var shape = new GeoSearchCircle(200, GeoUnit.Kilometers);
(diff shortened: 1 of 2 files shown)


sources/Valkey.Glide/BaseClient.GeospatialCommands.cs
```diff
@@ -0,0 +1,106 @@
...
+        return await Command(Request.GeoSearchAndStoreAsync(sourceKey, destinationKey, fromMember, shape, count, demandClosest, order, storeDistances));
+    }
+
+    /// <inheritdoc/>
+    public async Task<long> GeoSearchAndStoreAsync(ValkeyKey sourceKey, ValkeyKey destinationKey, GeoPosition fromPosition, GeoSearchShape shape, long count = -1, bool demandClosest = true, Order? order = null, bool storeDistances = false, CommandFlags flags = CommandFlags.None)
+    {
+        Utils.Requires<NotImplementedException>(flags == CommandFlags.None, "Command flags are not supported by GLIDE");
+        return await Command(Request.GeoSearchAndStoreAsync(sourceKey, destinationKey, fromPosition, shape, count, demandClosest, order, storeDistances));
+    }
+}
```
jeremyprime: Nit: The new geo classes are all missing a trailing newline, which goes against rules in `.editorconfig`.  If you install an EditorConfig plugin in your IDE it should automatically apply these rules when you add/update files.

sources/Valkey.Glide/Commands/IGeospatialCommands.cs
```diff
@@ -0,0 +1,279 @@
...
+    /// <param name="flags">The flags to use for this operation. Currently flags are ignored.</param>
+    /// <returns><see langword="true"/> if the member was added. <see langword="false"/> if the member was already a member of the sorted set and the score was updated.</returns>
+    /// <remarks>
+    /// <example>
+    /// <code>
+    /// bool wasAdded = await client.GeoAddAsync("mygeo", 13.361389, 38.115556, "Palermo");
+    /// </code>
+    /// </example>
+    /// </remarks>
+    Task<bool> GeoAddAsync(ValkeyKey key, double longitude, double latitude, ValkeyValue member, CommandFlags flags = CommandFlags.None);
```
currantw: The core GEOADD command says that it returns "the number of elements added to the sorted set". For consistency, does it make sense to return `Task<long>` here as well (0 or 1, obviously) -- or is this consistent with how other Glide clients implement this right now?

sources/Valkey.Glide/Commands/IGeospatialCommands.cs
```diff
@@ -0,0 +1,279 @@
...
+    /// <summary>
+    /// Adds the specified geospatial items to the specified key with options.
+    /// </summary>
+    /// <seealso href="https://valkey.io/commands/geoadd"/>
+    /// <param name="key">The key of the sorted set.</param>
+    /// <param name="values">The geospatial items to add.</param>
+    /// <param name="options">The options for the GEOADD command.</param>
+    /// <param name="flags">The flags to use for this operation. Currently flags are ignored.</param>
+    /// <returns>The number of elements added or changed, depending on options.</returns>
+    Task<long> GeoAddAsync(ValkeyKey key, GeoEntry[] values, GeoAddOptions options, CommandFlags flags = CommandFlags.None);
```
currantw: Why do we have separate methods that add options, instead of just making `options` an optional argument? I initially thought that this was to avoid a user having to provide an `options` value of `null` of they also wanted to provide a `flags` argument in the future ... but, if that is the case, that seems like it would also be the same problem with the `GeoDistanceAsync` and the `units` parameter below?

sources/Valkey.Glide/Commands/IGeospatialCommands.cs
```diff
@@ -0,0 +1,279 @@
...
+    /// <example>
+    /// <code>
+    /// GeoPosition?[] positions = await client.GeoPositionAsync("mygeo", new ValkeyValue[] { "Palermo", "Catania" });
+    /// </code>
+    /// </example>
+    /// </remarks>
+    Task<GeoPosition?[]> GeoPositionAsync(ValkeyKey key, ValkeyValue[] members, CommandFlags flags = CommandFlags.None);
+
+    /// <summary>
+    /// Returns the members of a geospatial index which are within the borders of the area specified by a given shape.
```
currantw: Maybe worth specifying that the shape is centred on the specified member in the summary? Is the member itself also returned (by definition, it is within the shape, so I assume so)? Last, I assume that it returns an empty array if the member does exist, but perhaps this is worth clarifying (or perhaps not)?

sources/Valkey.Glide/GeoAddOptions.cs
```diff
@@ -0,0 +1,50 @@
...
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="GeoAddOptions"/> class with changed option.
+    /// </summary>
+    /// <param name="changed">Whether to return the count of changed elements.</param>
+    public GeoAddOptions(bool changed)
+    {
+        ConditionalChange = null;
+        Changed = changed;
+    }
```
currantw: Maybe property initializers are a bit cleaner? 🤷 
Not really super familiar with C#...

```suggestion
    /// <summary>
    /// Gets the conditional change option.
    /// </summary>
    public ConditionalChange? ConditionalChange { get; } = null;

    /// <summary>
    /// Gets whether to return the count of changed elements instead of added elements.
    /// </summary>
    public bool Changed { get; } = false;

    /// <summary>
    /// Initializes a new instance of the <see cref="GeoAddOptions"/> class with conditional change option.
    /// </summary>
    /// <param name="conditionalChange">The conditional change option.</param>
    public GeoAddOptions(ConditionalChange conditionalChange)
    {
        ConditionalChange = conditionalChange;
    }

    /// <summary>
    /// Initializes a new instance of the <see cref="GeoAddOptions"/> class with changed option.
    /// </summary>
    /// <param name="changed">Whether to return the count of changed elements.</param>
    public GeoAddOptions(bool changed)
    {
        ConditionalChange = null;
    }
```

sources/Valkey.Glide/Internals/Request.GeospatialCommands.cs
```diff
@@ -0,0 +1,431 @@
...
+    }
+
+    /// <summary>
+    /// Creates a request for GEOADD command with multiple values and options.
+    /// </summary>
+    /// <param name="key">The key of the sorted set.</param>
+    /// <param name="values">The geospatial items to add.</param>
+    /// <param name="options">The options for the GEOADD command.</param>
+    /// <returns>A <see cref="Cmd{T, R}"/> with the request.</returns>
+    public static Cmd<long, long> GeoAddAsync(ValkeyKey key, GeoEntry[] values, GeoAddOptions options)
```
currantw: If we supported `options = null`, wouldn't it be possible remove all of these other internal `GeoAddAsync` methods, and have them all just call into this one method? Would remove a decent amount of duplication. 🤷

sources/Valkey.Glide/Internals/Request.GeospatialCommands.cs
```diff
@@ -0,0 +1,431 @@
...
+        {
+            return response.Select(item => 
+            {
+                if (item == null) return (GeoPosition?)null;
+                var posArray = (object[])item;
+                if (posArray.Length < 2 || posArray[0] == null || posArray[1] == null) return (GeoPosition?)null;
+                return (GeoPosition?)new GeoPosition(double.Parse(posArray[0].ToString()!), double.Parse(posArray[1].ToString()!));
+            }).ToArray();
+        });
+    }
```
currantw: Could we also combine these into one method, and then the caller could be responsible for inpacking the `Geoposition?[]` as needed? At the very least, maybe we should extract the common response parsing logic?

sources/Valkey.Glide/Internals/Request.GeospatialCommands.cs
```diff
@@ -0,0 +1,431 @@
...
+    /// Creates a request for GEOSEARCH command with position origin.
+    /// </summary>
+    /// <param name="key">The key of the sorted set.</param>
+    /// <param name="fromPosition">The position to search from.</param>
+    /// <param name="shape">The search area shape.</param>
+    /// <returns>A <see cref="Cmd{T, R}"/> with the request.</returns>
+    public static Cmd<object[], GeoRadiusResult[]> GeoSearchAsync(ValkeyKey key, GeoPosition fromPosition, GeoSearchShape shape)
+    {
+        return GeoSearchAsync(key, fromPosition, shape, -1, true, null, GeoRadiusOptions.None);
+    }
```
currantw: Rather than having this extra method, why not just have the caller call the correct `GeoSearchAsync` method itself directly? Same comment for other similar ones.

tests/Valkey.Glide.IntegrationTests/GeospatialCommandTests.cs
```diff
@@ -0,0 +1,974 @@
...
+        ];
+        await client.GeoAddAsync(key, entries);
+        
+        // Test default unit (should be meters)
+        double? distanceDefault = await client.GeoDistanceAsync(key, "Palermo", "Catania");
+        double? distanceMeters = await client.GeoDistanceAsync(key, "Palermo", "Catania", GeoUnit.Meters);
+        
+        Assert.NotNull(distanceDefault);
+        Assert.NotNull(distanceMeters);
+        Assert.Equal(distanceMeters.Value, distanceDefault.Value, 1e-9);
```
currantw: Is this considered best practice in C#? Won't the test fail even if we just have the `AssertEqual` test if one of the value is `null`?

```suggestion
        Assert.Equal(distanceMeters.Value, distanceDefault.Value, 1e-9);
```

sources/Valkey.Glide/Commands/IGeospatialCommands.cs
alexr-bq: Return type is as it is for consistency with StackExchangeRedis interface.

sources/Valkey.Glide/Commands/IGeospatialCommands.cs
```diff
@@ -0,0 +1,279 @@
...
+    /// <summary>
+    /// Adds the specified geospatial item to the specified key with options.
+    /// </summary>
+    /// <seealso href="https://valkey.io/commands/geoadd"/>
+    /// <param name="key">The key of the sorted set.</param>
+    /// <param name="value">The geospatial item to add.</param>
+    /// <param name="options">The options for the GEOADD command.</param>
+    /// <param name="flags">The flags to use for this operation. Currently flags are ignored.</param>
+    /// <returns>The number of elements added or changed, depending on options.</returns>
+    Task<long> GeoAddAsync(ValkeyKey key, GeoEntry value, GeoAddOptions options, CommandFlags flags = CommandFlags.None);
```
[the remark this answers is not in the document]
alexr-bq: Fixing

tests/Valkey.Glide.IntegrationTests/GeospatialCommandTests.cs
```diff
@@ -0,0 +1,974 @@
...
+        Assert.True(distanceFeet > 545000 && distanceFeet < 546000); // ~545,518 feet
+        
+        // Verify unit conversions are consistent
+        double metersToKm = distanceMeters.Value / 1000;
+        double metersToMiles = distanceMeters.Value / 1609.344;
+        double metersToFeet = distanceMeters.Value * 3.28084;
+        
+        Assert.True(Math.Abs(metersToKm - distanceKilometers.Value) < 0.001);
+        Assert.True(Math.Abs(metersToMiles - distanceMiles.Value) < 0.001);
+        Assert.True(Math.Abs(metersToFeet - distanceFeet.Value) < 1); // Allow 1 foot tolerance
```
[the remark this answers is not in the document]
alexr-bq: Following what seem to be general patterns but it's a good larger conversation we should have about the scope of testing for our wrappers.

sources/Valkey.Glide/Internals/Request.GeospatialCommands.cs
```diff
@@ -0,0 +1,431 @@
+// Copyright Valkey GLIDE Project Contributors - SPDX Identifier: Apache-2.0
```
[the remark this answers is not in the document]
alexr-bq: I went the route of just factoring out common logic from within the Request class, though I see the benefit of what you're saying here

sources/Valkey.Glide/Pipeline/BaseBatch.GeospatialCommands.cs
```diff
@@ -0,0 +1,64 @@
...
+
+    /// <inheritdoc cref="IBatchGeospatialCommands.GeoSearch(ValkeyKey, ValkeyValue, GeoSearchShape, long, bool, Order?, GeoRadiusOptions)" />
+    public T GeoSearch(ValkeyKey key, ValkeyValue fromMember, GeoSearchShape shape, long count = -1, bool demandClosest = true, Order? order = null, GeoRadiusOptions options = GeoRadiusOptions.Default) => AddCmd(GeoSearchAsync(key, fromMember, shape, count, demandClosest, order, options));
+
+    /// <inheritdoc cref="IBatchGeospatialCommands.GeoSearch(ValkeyKey, GeoPosition, GeoSearchShape, long, bool, Order?, GeoRadiusOptions)" />
+    public T GeoSearch(ValkeyKey key, GeoPosition fromPosition, GeoSearchShape shape, long count = -1, bool demandClosest = true, Order? order = null, GeoRadiusOptions options = GeoRadiusOptions.Default) => AddCmd(GeoSearchAsync(key, fromPosition, shape, count, demandClosest, order, options));
+
+
+
+    // Explicit interface implementations for IBatchGeospatialCommands
```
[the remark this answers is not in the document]
alexr-bq: Not completely sure, just following existing pattern here.

sources/Valkey.Glide/Commands/IGeospatialCommands.cs
jbrinkman: @currantw As Alex says, we need to ensure our public API is compatible with the StackExchange.Redis API. This will often result in differences from how we handle commands in other languages, but it is consistent with what StackExchange does.

sources/Valkey.Glide/BaseClient.GeospatialCommands.cs
```diff
@@ -0,0 +1,106 @@
...
+    /// <inheritdoc/>
+    public async Task<bool> GeoAddAsync(ValkeyKey key, double longitude, double latitude, ValkeyValue member, CommandFlags flags = CommandFlags.None)
+    {
+        return await GeoAddAsync(key, new GeoEntry(longitude, latitude, member), flags).ConfigureAwait(false);
+    }
+
+    /// <inheritdoc/>
+    public async Task<bool> GeoAddAsync(ValkeyKey key, GeoEntry value, CommandFlags flags = CommandFlags.None)
+    {
+        Utils.Requires<NotImplementedException>(flags == CommandFlags.None, "Command flags are not supported by GLIDE");
```
jbrinkman: These exception messages should be implemented as a const. Should be able to use the same const across all command implementations not just Geo commands.

sources/Valkey.Glide/ConditionalChange.cs
```diff
@@ -0,0 +1,19 @@
...
+    /// <summary>
+    /// Only add new elements. Don't update already existing elements.
+    /// </summary>
+    ONLY_IF_DOES_NOT_EXIST,
+
+    /// <summary>
+    /// Only update elements that already exist. Don't add new elements.
+    /// </summary>
+    ONLY_IF_EXISTS
+}
```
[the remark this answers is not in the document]
currantw: Gotcha. Guess we are stuck with that then... 😢

sources/Valkey.Glide/Commands/IGeospatialCommands.cs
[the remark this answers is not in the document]
currantw: Looks good, thanks

sources/Valkey.Glide/Commands/IGeospatialCommands.cs
currantw: Looks like you updated this to say "The shape is centred on the specified member or longlat." What case is the "or longlat" referencing here - doesn't this method only accept a member as the centre point?

tests/Valkey.Glide.IntegrationTests/GeospatialCommandTests.cs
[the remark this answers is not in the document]
currantw: Thanks. If its consistent with the pattern elsewhere, then I'm fine with leaving it. Not need to get into the "larger conversation" now. 😉

tests/Valkey.Glide.IntegrationTests/GeospatialCommandTests.cs
```diff
@@ -0,0 +1,974 @@
...
+            new GeoEntry(13.361389, 38.115556, "Palermo"),
+            new GeoEntry(15.087269, 37.502669, "Catania")
+        ];
+        await client.GeoAddAsync(key, entries);
+        
+        string?[] hashes = await client.GeoHashAsync(key, new ValkeyValue[] { "Palermo", "Catania", "NonExistent" });
+        Assert.Equal(3, hashes.Length);
+        Assert.NotNull(hashes[0]); // Palermo
+        Assert.NotNull(hashes[1]); // Catania
+        Assert.Null(hashes[2]); // NonExistent
```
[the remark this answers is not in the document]
currantw: Whoops! Looks like I didn't read this well enough. 🤦 
Looks like you've removed the check for length though, which I think we want?

sources/Valkey.Glide/Internals/Request.GeospatialCommands.cs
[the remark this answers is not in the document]
currantw: Sure, I'm happy with that.

sources/Valkey.Glide/BaseClient.GeospatialCommands.cs
alexr-bq: https://github.com/valkey-io/valkey-glide-csharp/issues/112

PR #113, opened 2025-10-22
Implement Cluster Scan

currantw:
# Implement Cluster Scan

## Overview

Adds `ScanAsync` methods to the standalone and cluster clients.

## Key Changes
- Added `ScanAsync` to `IGenericClusterCommands` and implemented it in `GlideClusterClient`, which calls new `BaseClient.ClusterScanCommand` method.
- Added `ScanAsync` to `IGenericCommands` (standalone client) and implemented it in `GlideClusterClient`, which calls the existing Glide SCAN command. Provides consistent scan interface for standalone and cluster clients.
- Cluster scan is implemented with `ClusterScanCursor`, while both standalone and cluster scanning use `ScanOptions`, which supports match pattern, count, and type arguments. Refactor existing `Request.GenericCommands.ScanAsync` and `KeysAsync` method to use `ScanOptions`.
- Added `RequestClusterScanFfi` and `RemoveClusterScanCursorFfi` methods to "FFI.methods.rs", as well as Add `request_cluster_scan` and `remove_cluster_scan_cursor` to "rust/src/lib.rs" that call into the existing core methods (implementation based on equivalent methods [here](https://github.com/valkey-io/valkey-glide/blob/main/ffi/src/lib.rs)).

## Testing Coverage

- Added new `ScanTests` for both standalone and cluster `ScanAsync` methods.
- Updated `CommandTests` for change to internal `Request.ScanAsync` interface. 

## Breaking Changes

None.

### DIFF
--- a/sources/Valkey.Glide/GlideClusterClient.cs
+++ b/sources/Valkey.Glide/GlideClusterClient.cs
@@ -324,6 +324,21 @@ public sealed class GlideClusterClient : BaseClient, IGenericClusterCommands, IS
     }
 
     /// <summary>
+    /// Iterates incrementally over keys in the cluster.
+    /// </summary>
+    /// <param name="cursor">The cursor to use for this iteration.</param>
+    /// <param name="options">Optional scan options to filter results.</param>
+    /// <returns>A tuple containing the next cursor and the keys found in this iteration.</returns>
+    /// <seealso cref="ClusterScanCursor"/>
+    /// <seealso cref="ScanOptions"/>
+    public async Task<(ClusterScanCursor cursor, ValkeyKey[] keys)> ScanAsync(ClusterScanCursor cursor, ScanOptions? options = null)
+    {
+        string[] args = options?.ToArgs() ?? [];
+        var (nextCursorId, keys) = await ClusterScanCommand(cursor.CursorId, args);
+        return (new ClusterScanCursor(nextCursorId), keys);
+    }
+
+    /// <summary>
     /// Executes a cluster scan command with the given cursor and arguments.
     /// </summary>
     /// <param name="cursor">The cursor for the scan iteration.</param>
@@ -398,20 +413,5 @@ public sealed class GlideClusterClient : BaseClient, IGenericClusterCommands, IS
             RemoveClusterScanCursorFfi(cursorPtr);
             Marshal.FreeHGlobal(cursorPtr);
         }
-    }
-
-    /// <summary>
-    /// Iterates incrementally over keys in the cluster.
-    /// </summary>
-    /// <param name="cursor">The cursor to use for this iteration.</param>
-    /// <param name="options">Optional scan options to filter results.</param>
-    /// <returns>A tuple containing the next cursor and the keys found in this iteration.</returns>
-    /// <seealso cref="ClusterScanCursor"/>
-    /// <seealso cref="ScanOptions"/>
-    public async Task<(ClusterScanCursor cursor, ValkeyKey[] keys)> ScanAsync(ClusterScanCursor cursor, ScanOptions? options = null)
-    {
-        string[] args = options?.ToArgs() ?? [];
-        var (nextCursorId, keys) = await ClusterScanCommand(cursor.CursorId, args);
-        return (new ClusterScanCursor(nextCursorId), keys);
     }
 }


rust/src/lib.rs
```diff
@@ -438,3 +439,288 @@ pub unsafe extern "C" fn init(level: Option<Level>, file_name: *const c_char) ->
...
+
+/// Execute a cluster scan request.
+///
+/// # Safety
+/// * `client_ptr` must be a valid Client pointer from create_client
+/// * `cursor` must be "0" for initial scan or a valid cursor ID from previous scan
+/// * `args` and `arg_lengths` must be valid arrays of length `arg_count`
+/// * `args` format: [b"MATCH", pattern, b"COUNT", count, b"TYPE", type] (all optional)
+#[unsafe(no_mangle)]
+pub unsafe extern "C-unwind" fn request_cluster_scan(
```
currantw: Don't have any experience with Rust. Implementation modelled on existing cluster scan methods at here](https://github.com/valkey-io/valkey-glide/blob/main/ffi/src/lib.rs). Any feedback appreciated!

rust/src/lib.rs
```diff
@@ -438,3 +439,288 @@ pub unsafe extern "C" fn init(level: Option<Level>, file_name: *const c_char) ->
...
+                    (core.failure_callback)(
+                        callback_index,
+                        format!("Invalid cursor ID: {}", cursor_id).as_ptr() as *const c_char,
+                        RequestErrorType::Unspecified,
+                    );
+                }
+                return;
+            }
+        }
+    };
```
currantw: Updated to handle the case (raise an error) where an invalid cursor is provided, as described in [this pull request](https://github.com/valkey-io/valkey-glide/pull/4823).

sources/Valkey.Glide/BaseClient.cs
```diff
@@ -132,6 +132,11 @@ internal virtual async Task<T> Command<R, T>(Cmd<R, T> command, Route? route = n
     }
     #endregion protected methods
 
+    #region protected fields
+    protected Version? _serverVersion; // cached server version
+    protected static readonly Version DefaultServerVersion = new(8, 0, 0);
+    #endregion protected fields
```
currantw: Not new. Just moved from "private fields" region below (since these obviously are _not_ private...)

sources/Valkey.Glide/Internals/Request.GenericCommands.cs
```diff
@@ -318,25 +318,19 @@ public static Cmd<long, long> SortAndStoreAsync(ValkeyKey destination, ValkeyKey
     public static Cmd<bool, bool> KeyMoveAsync(ValkeyKey key, int database)
         => Simple<bool>(RequestType.Move, [key.ToGlideString(), database.ToGlideString()]);
 
-    public static Cmd<object[], (long, ValkeyKey[])> ScanAsync(long cursor, ValkeyValue pattern = default, long pageSize = 0)
+    public static Cmd<object[], (string, ValkeyKey[])> ScanAsync(string cursor, ScanOptions? options = null)
```
currantw: Updated to use `ScanOptions` and for consistency with the cluster interface.

sources/Valkey.Glide/BaseClient.cs
```diff
@@ -150,6 +155,77 @@ private void FailureCallback(ulong index, IntPtr strPtr, RequestErrorType errTyp
 
     protected abstract Task<Version> GetServerVersionAsync();
 
+    protected async Task<(string cursor, ValkeyKey[] keys)> ClusterScanCommand(string cursor, string[] args)
```
currantw: I would have liked to put this in `GlideClusterClient`, since it is actually only applicable to the cluster client (obviously). But that would require changing the access modifier for `_messageContainer`, which I'm not sure is advisable? @jbrinkman / @alexr-bq any thoughts?

sources/Valkey.Glide/Commands/IGenericCommands.cs
```diff
@@ -167,6 +168,33 @@ public interface IGenericCommands
...
+    ///
+    /// do
+    /// {
+    ///     (cursor, var keys) = await client.ScanAsync(cursor);
+    ///     allKeys.AddRange(keys);
+    /// } while (cursor != "0");
+    /// </code>
+    /// </example>
+    /// <seealso href="https://valkey.io/commands/scan/">SCAN command</seealso>
+    Task<(string cursor, ValkeyKey[] keys)> ScanAsync(string cursor, ScanOptions? options = null);
```
currantw: Standalone `ScanAsync` added for consistency with other clients. There is still the existing `KeysAsync`, which matches the existing StackExchange.Redis interface, but that does not provide support for the filtering keys by type.

sources/Valkey.Glide/BaseClient.cs
jduo: Would it be possible to give _messageContainer an internal access level? That way arbitrary user code that extends BaseClient doesn't have access to this method.

sources/Valkey.Glide/GlideClient.cs
```diff
@@ -177,25 +177,34 @@ public async IAsyncEnumerable<ValkeyKey> KeysAsync(int database = -1, ValkeyValu
...
         Utils.Requires<NotImplementedException>(flags == CommandFlags.None, "Command flags are not supported by GLIDE");
 
-        long currentCursor = cursor;
+        string currentCursor = cursor.ToString();
         int currentOffset = pageOffset;
 
         do
         {
-            (long nextCursor, ValkeyKey[] keys) = await Command(Request.ScanAsync(currentCursor, pattern, pageSize));
+            var options = new ScanOptions();
```
jduo: Nit: let's allocate this outside the loop and reuse

jduo (approved):

sources/Valkey.Glide/GlideClusterClient.cs
```diff
@@ -318,4 +322,96 @@ protected override async Task<Version> GetServerVersionAsync()
...
+    /// <param name="options">Optional scan options to filter results.</param>
+    /// <returns>A tuple containing the next cursor and the keys found in this iteration.</returns>
+    /// <seealso cref="ClusterScanCursor"/>
+    /// <seealso cref="ScanOptions"/>
+    public async Task<(ClusterScanCursor cursor, ValkeyKey[] keys)> ScanAsync(ClusterScanCursor cursor, ScanOptions? options = null)
+    {
+        string[] args = options?.ToArgs() ?? [];
+        var (nextCursorId, keys) = await ClusterScanCommand(cursor.CursorId, args);
+        return (new ClusterScanCursor(nextCursorId), keys);
+    }
```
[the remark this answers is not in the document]
currantw: Hmmmm. `KeysAsync` is intended to match the corresponding method in StackExchange.Redis, which does not support cluster mode, so adding it for cluster clients this isn't necessary from this perspective We _could_, of course, implement that method for cluster clients now that we support cluster scan, personally, but I would he hesitant to do that without significantly changing the interface: I think the interface for the `ScanAsync` is much cleaner and less clunky, and that adding `KeysAsync` provides pretty limited additional benefit.

To summarize:
- I don't think it's necessary to implement `KeysAsync` for cluster client.
- If we do want to implement it, I think we should probably not follow the interface used for standalone clients.
- If we do want to implement it, it might make sense to do as a separate issue.

Thoughts?

sources/Valkey.Glide/GlideClient.cs
```diff
@@ -177,25 +177,35 @@ public async IAsyncEnumerable<ValkeyKey> KeysAsync(int database = -1, ValkeyValu
     {
```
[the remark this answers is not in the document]
currantw: As mentioned in my [comment above](https://github.com/valkey-io/valkey-glide-csharp/pull/113#discussion_r2473761490), I'm not sure we want to make this method work with cluster clients. In particular, the `long cursor = 0` argument doesn't really have any applicability to cluster scan, where the cursor ID is hidden behind `ClusterScanCursor`. Happy to discuss further.

alexr-bq (approved):

PR #135, opened 2025-11-07
feat: Support password update

currantw:
## Summary

This pull request implements [dynamic password update functionality](https://github.com/valkey-io/valkey-glide/wiki/General-Concepts#dynamic-password-management).

## Key Changes

- **Rust FFI**: Added `update_connection_password` function.
- **C# Bindings**: Added FFI method declarations for both .NET 8+ and legacy frameworks.
- **Public API**: Added `UpdateConnectionPasswordAsync` and `ClearConnectionPasswordAsync` methods to `BaseClient`
- **Integration Tests**: Added comprehensive tests for both standalone and cluster clients

## API Usage

```csharp
// Update password without immediate authentication
await client.UpdateConnectionPasswordAsync("new_password");

// Update password with immediate authentication  
await client.UpdateConnectionPasswordAsync("new_password", immediateAuth: true);

// Remove password from connection
await client.ClearConnectionPasswordAsync();
```

_Note_: There is some variation in the dynamic password update interface between clients. In Java and Node, a single update method is provided; if the password is null or the empty string, the password is cleared. In Go, there is a separate method that "resets" the password. I chose to follow Go's approach, with two separate methods, because I think that it is more explicit. (Of course, I though that "clear" made the effects of the method \*ahem\* _clearer_ than "reset").

### DIFF
--- a/tests/Valkey.Glide.IntegrationTests/UpdateConnectionPasswordTests.cs
+++ b/tests/Valkey.Glide.IntegrationTests/UpdateConnectionPasswordTests.cs
@@ -28,32 +28,32 @@ public class UpdateConnectionPasswordTests(TestConfiguration config)
             using var client = await GlideClient.CreateClient(config);
             using var adminClient = await GlideClient.CreateClient(config);
 
-            VerifyConnection(client);
-            VerifyConnection(adminClient);
+            await VerifyConnection(client);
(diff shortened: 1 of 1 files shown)


alexr-bq (approved):

yipin-chen (approved):
LGTM

PR #137, opened 2025-11-07
Transaction commands

alexr-bq:
Adds support for WATCH and UNWATCH commands. These are used in conjunction with existing Batch/Transaction functionality, with tests of this behaviour.

### DIFF
--- a/tests/Valkey.Glide.UnitTests/CommandTests.cs
+++ b/tests/Valkey.Glide.UnitTests/CommandTests.cs
@@ -809,7 +809,7 @@ public class CommandTests
             // Test HashGetAsync with multiple fields
             () =>
             {
-                ValkeyValue[] result = Request.HashGetAsync("key", ["field1", "field2", "field3"]).Converter([.. testList]);
+                ValkeyValue[] result = Request.HashGetAsync("key", ["field1", "field2", "field3"]).Converter((object[])testList.ToArray()!);
                 Assert.Equal(3, result.Length);
                 Assert.Equal("value1", result[0]);
                 Assert.Equal("value2", result[1]);


sources/Valkey.Glide/GlideClient.cs
```diff
@@ -202,6 +202,18 @@ public async IAsyncEnumerable<ValkeyKey> KeysAsync(int database = -1, ValkeyValu
     public async Task<(string cursor, ValkeyKey[] keys)> ScanAsync(string cursor, ScanOptions? options = null)
         => await Command(Request.ScanAsync(cursor, options));
 
+    public async Task<string> WatchAsync(ValkeyKey[] keys, CommandFlags flags = CommandFlags.None)
+    {
+        Utils.Requires<NotImplementedException>(flags == CommandFlags.None, "Command flags are not supported by GLIDE");
```
currantw: I've merged a helper method for this logic that you can use here: `GuardClauses.ThrowIfCommandFlags`

sources/Valkey.Glide/Commands/ITransactionBaseCommands.cs
```diff
@@ -0,0 +1,53 @@
...
+    ///     .StringSetAsync("sampleKey", "foobar");
+    /// // Modify the watched key from another client/connection
+    /// await client.StringSetAsync("sampleKey", "hello world");
+    /// object[] transactionResult2 = await client.Exec(batch2, true);
+    /// // transactionResult2 is null because the watched key was modified
+    /// </code>
+    /// </example>
+    /// </remarks>
+    /// <seealso href="https://valkey.io/commands/watch/"/>
+    Task<string> WatchAsync(ValkeyKey[] keys, CommandFlags flags = CommandFlags.None);
```
currantw: Two questions about this interface:
- What is the error behaviour here, if any? Do we need to specify any exceptions that might be thrown? For example, what happens if the key doesn't exist? What is the list contains duplicate keys?
- What is the value in returning "OK" here? If that is the only possible return type? If so, would it just be possible to return a blank `Task` on success? (It doesn't _seem_ to me like we are matching a StackExchange.Redis API here?)

alexr-bq: Fixed to not return "OK", other similar commands return blank Task like you suggest

Added doc of exception, but command works fine with with non-existant keys, won't error.

sources/Valkey.Glide/Commands/ITransactionBaseCommands.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// will be split across these slots and executed separately for each. This means the command
+    /// is atomic only at the slot level. If one or more slot-specific requests fail, the entire
+    /// call will return the first encountered error, even though some requests may have succeeded
+    /// while others did not. If this behavior impacts your application logic, consider splitting
+    /// the request into sub-requests per slot to ensure atomicity.
+    /// </para>
+    /// <example>
+    /// <code>
+    /// bool result = await client.WatchAsync(["sampleKey"]);
+    /// // result is "OK"
```
currantw: Small doc update needed here.

sources/Valkey.Glide/Commands/ITransactionCommands.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// Flushes all the previously watched keys for a transaction. Executing a transaction will
+    /// automatically flush all previously watched keys.
+    /// </summary>
+    /// <param name="flags">The flags to use for this operation. Currently flags are ignored.</param>
+    /// <exception cref="RequestException">Thrown if the command fails to execute on the server.</exception>
+    /// <example>
+    /// <code>
+    /// await client.WatchAsync(["sampleKey"]);
+    /// bool result = await client.UnwatchAsync();
+    /// // result is "OK", "sampleKey" is no longer watched
```
currantw: Small update also needed here.

currantw (approved):
LGTM. Just a couple documentation updates from previous changes.

PR #151, opened 2025-11-23
Stream Commands

alexr-bq:
Implements stream commands.

Basic Operations: XADD, XREAD, XRANGE/XREVRANGE, XLEN, XDEL, XTRIM
Consumer Groups: XGROUP CREATE/DESTROY/SETID, XGROUP DELCONSUMER, XREADGROUP, XACK
Pending Messages: XPENDING (summary and detailed)
Message Claiming: XCLAIM, XAUTOCLAIM (with JUSTID variants)
Stream Info: XINFO STREAM (basic and FULL), XINFO GROUPS, XINFO CONSUMERS

### DIFF
--- a/sources/Valkey.Glide/OpenTelemetry.cs
+++ b/sources/Valkey.Glide/OpenTelemetry.cs
@@ -16,6 +16,13 @@ public static class OpenTelemetry
     private static OpenTelemetryConfig? s_config;
 
     /// <summary>
+    /// Internal delegate for FFI initialization.
+    /// For testing purposes only.
+    /// See <see cref="T:Valkey.Glide.Tests.OpenTelemetryTests"/>.
+    /// </summary>
+    internal static Func<IntPtr, IntPtr> InitOpenTelemetryFfi = FFI.InitOpenTelemetryFfi;
+
+    /// <summary>
     /// Initialize OpenTelemetry with the provided configuration.
     /// Can only be called once per process. Subsequent calls will be ignored.
     /// </summary>
@@ -112,6 +119,11 @@ public static class OpenTelemetry
     /// <summary>
     /// Clear the OpenTelemetry configuration.
     /// For testing purposes only.
+    /// See <see cref="T:Valkey.Glide.Tests.OpenTelemetryTests"/>.
+    ///
+    /// Note that this method only clear OpenTelemetry in the C#
+    /// client: the underlying Rust OpenTelemetry initialization
+    /// remains active for the process lifetime.
     /// </summary>
     internal static void Clear()
     {
@@ -139,7 +151,7 @@ public static class OpenTelemetry
         try
         {
             Marshal.StructureToPtr(ffiConfig, configPtr, false);
-            var errorPtr = FFI.InitOpenTelemetryFfi(configPtr);
+            var errorPtr = InitOpenTelemetryFfi(configPtr);
 
             if (errorPtr != IntPtr.Zero)
             {
--- a/tests/Valkey.Glide.IntegrationTests/OpenTelemetryTests.cs
+++ b/tests/Valkey.Glide.IntegrationTests/OpenTelemetryTests.cs
@@ -18,16 +18,33 @@ public class OpenTelemetryTests : IDisposable
     private TracesFile Traces { get; }
     private static readonly string TracesFilePath = System.IO.Path.GetTempFileName();
 
-    // Before each test, initialize traces file.
-    public OpenTelemetryTests()
+    static OpenTelemetryTests()
     {
-        Traces = new TracesFile(TracesFilePath);
+        // Before any tests, initialize OpenTelemetry.
+        Assert.False(OpenTelemetry.IsInitialized());
+
+        var tracesConfig = TracesConfig.CreateBuilder()
+            .WithEndpoint($"file://{TracesFilePath}")
+            .Build();
+
+        var config = OpenTelemetryConfig.CreateBuilder()
+            .WithTraces(tracesConfig)
+            .WithFlushInterval(FlushInterval)
+            .Build();
+
+        OpenTelemetry.Init(config);
     }
 
-    // After each test, clear OpenTelemetry and dispose of traces file.
+    public OpenTelemetryTests()
+    {
+        // Before each test, initialize traces file.
+        Traces = new TracesFile();
+    }
+
     public void Dispose()
     {
-        OpenTelemetry.Clear();
+        // After each test, turn off tracing and dispose of traces file.
+        OpenTelemetry.SetSamplePercentage(SamplePercentageNone);
         Traces.Dispose();
     }
 
@@ -35,7 +52,7 @@ public class OpenTelemetryTests : IDisposable
     [MemberData(nameof(TestConfiguration.TestClients), MemberType = typeof(TestConfiguration))]
     public async Task Commands_WhenSamplingNone_NoSpans(BaseClient client)
     {
-        InitializeOpenTelemetry(SamplePercentageNone);
+        OpenTelemetry.SetSamplePercentage(SamplePercentageNone);
         await ExecuteSetGetDelete(client);
         Traces!.AssertSpanNames([]);
     }
@@ -45,7 +62,7 @@ public class OpenTelemetryTests : IDisposable
     [MemberData(nameof(TestConfiguration.TestClients), MemberType = typeof(TestConfiguration))]
     public async Task Commands_WhenSamplingAll_CreateSpans(BaseClient client)
     {
-        InitializeOpenTelemetry(SamplePercentageAll);
+        OpenTelemetry.SetSamplePercentage(SamplePercentageAll);
         await ExecuteSetGetDelete(client);
         Traces!.AssertSpanNames(["SET", "GET", "DEL"]);
     }
@@ -54,7 +71,7 @@ public class OpenTelemetryTests : IDisposable
     [MemberData(nameof(TestConfiguration.TestClients), MemberType = typeof(TestConfiguration))]
     public async Task Batch_WhenSamplingNone_NoSpans(BaseClient client)
     {
-        InitializeOpenTelemetry(SamplePercentageNone);
+        OpenTelemetry.SetSamplePercentage(SamplePercentageNone);
         await ExecuteBatchSetGetDelete(client);
         Traces!.AssertSpanNames([]);
     }
@@ -63,24 +80,9 @@ public class OpenTelemetryTests : IDisposable
     [MemberData(nameof(TestConfiguration.TestClients), MemberType = typeof(TestConfiguration))]
     public async Task Batch_WhenSamplingAll_CreateSpans(BaseClient client)
     {
-        InitializeOpenTelemetry(SamplePercentageAll);
+        OpenTelemetry.SetSamplePercentage(SamplePercentageAll);
         await ExecuteBatchSetGetDelete(client);
         Traces!.AssertSpanNames(["Batch"]);
-    }
-
-    private static void InitializeOpenTelemetry(uint samplePercentage)
-    {
-        var tracesConfig = TracesConfig.CreateBuilder()
-            .WithEndpoint($"file://{TracesFilePath}")
-            .WithSamplePercentage(samplePercentage)
-            .Build();
-
-        var config = OpenTelemetryConfig.CreateBuilder()
-            .WithTraces(tracesConfig)
-            .WithFlushInterval(FlushInterval)
-            .Build();
-
-        OpenTelemetry.Init(config);
     }
 
     // Executes SET, GET, and DEL commands on the given client.
@@ -124,24 +126,20 @@ public class OpenTelemetryTests : IDisposable
     // Temporary file for storing traces.
     private readonly struct TracesFile : IDisposable
     {
-        private readonly string _path;
-
-        public TracesFile(string path)
+        public TracesFile()
         {
-            _path = path;
-
             // Ensure file exists and is initially empty.
-            File.WriteAllText(path, string.Empty);
+            File.WriteAllText(TracesFilePath, string.Empty);
         }
 
         public readonly void Dispose()
         {
-            File.Delete(_path);
+            File.Delete(TracesFilePath);
         }
 
         public readonly void AssertSpanNames(List<string> expectedSpanNames)
         {
-            var lines = File.ReadAllLines(_path);
+            var lines = File.ReadAllLines(TracesFilePath);
 
             var actualSpanNames = new List<string>();
             foreach (var line in lines)
--- a/tests/Valkey.Glide.UnitTests/OpenTelemetryTests.cs
+++ b/tests/Valkey.Glide.UnitTests/OpenTelemetryTests.cs
@@ -10,11 +10,18 @@ public class OpenTelemetryTests : IDisposable
     private const uint SamplePercentageLow = 10u;
     private const uint SamplePercentageHigh = 90u;
     private const uint SamplePercentageMax = 100u;
-    private const string TestEndpoint = "http://localhost:4317";
+    private const string Endpoint = "ENDPOINT";
 
-    // After each test, clear OpenTelemetry.
+    public OpenTelemetryTests()
+    {
+        // Before each test, mock the FFI delegate to prevent actual initialization.
+        OpenTelemetry.InitOpenTelemetryFfi = MockInitOpenTelemetryFfi;
+    }
+
     public void Dispose()
     {
+        // After each test, clear configuration and reset the FFI delegate to the original.
+        OpenTelemetry.InitOpenTelemetryFfi = FFI.InitOpenTelemetryFfi;
         OpenTelemetry.Clear();
     }
 
@@ -28,7 +35,7 @@ public class OpenTelemetryTests : IDisposable
     public void Init_WithValidConfig_Succeeds()
     {
         var traces = TracesConfig.CreateBuilder()
-            .WithEndpoint(TestEndpoint)
+            .WithEndpoint(Endpoint)
             .WithSamplePercentage(SamplePercentageLow)
             .Build();
         var config = OpenTelemetryConfig.CreateBuilder()
@@ -45,7 +52,7 @@ public class OpenTelemetryTests : IDisposable
     public void Init_CalledTwice_IgnoresSecondCall()
     {
         var traces1 = TracesConfig.CreateBuilder()
-            .WithEndpoint(TestEndpoint)
+            .WithEndpoint(Endpoint)
             .WithSamplePercentage(SamplePercentageLow)
             .Build();
         var config1 = OpenTelemetryConfig.CreateBuilder()
@@ -53,7 +60,7 @@ public class OpenTelemetryTests : IDisposable
             .Build();
 
         var traces2 = TracesConfig.CreateBuilder()
-            .WithEndpoint(TestEndpoint)
+            .WithEndpoint(Endpoint)
             .WithSamplePercentage(SamplePercentageHigh)
             .Build();
         var config2 = OpenTelemetryConfig.CreateBuilder()
@@ -83,7 +90,7 @@ public class OpenTelemetryTests : IDisposable
     public void GetSamplePercentage_WithTracesConfigured_ReturnsDefault()
     {
         var traces = TracesConfig.CreateBuilder()
-            .WithEndpoint(TestEndpoint)
+            .WithEndpoint(Endpoint)
             .Build();
         var config = OpenTelemetryConfig.CreateBuilder()
             .WithTraces(traces)
@@ -98,7 +105,7 @@ public class OpenTelemetryTests : IDisposable
     public void GetSamplePercentage_WithSamplePercentageConfigured_ReturnsPercentage()
     {
         var traces = TracesConfig.CreateBuilder()
-            .WithEndpoint(TestEndpoint)
+            .WithEndpoint(Endpoint)
             .WithSamplePercentage(SamplePercentageLow)
             .Build();
         var config = OpenTelemetryConfig.CreateBuilder()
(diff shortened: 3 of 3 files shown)


sources/Valkey.Glide/Commands/IStreamCommands.cs
```diff
@@ -0,0 +1,463 @@
...
+    Task<ValkeyValue> StreamAddAsync(ValkeyKey key, NameValueEntry[] streamPairs, ValkeyValue? messageId, int? maxLength, bool useApproximateMaxLength, CommandFlags flags);
+
+    /// <summary>
+    /// Appends a new entry to a stream with a single field-value pair.
+    /// </summary>
+    /// <seealso href="https://valkey.io/commands/xadd"/>
+    /// <param name="key">The key of the stream.</param>
+    /// <param name="streamField">The field name.</param>
+    /// <param name="streamValue">The field value.</param>
+    /// <param name="messageId">The message ID. Use null or "*" for auto-generation, or "&lt;ms&gt;-*" to auto-generate sequence number for a specific timestamp.</param>
```
currantw: Do the `StreamAddAsync` methods not support auto-generating a sequence number for a specific timestamp?

sources/Valkey.Glide/Commands/IStreamCommands.cs
```diff
@@ -0,0 +1,463 @@
...
+    /// <param name="maxLength">The maximum length of the stream. If specified, old entries will be trimmed. Mutually exclusive with minId.</param>
+    /// <param name="useApproximateTrimming">If true, uses approximate trimming (~) for better performance.</param>
```
currantw: Should this be `useApproximateMaxLength` for consistency (both internally and with `StackExchange.Redis`)?

sources/Valkey.Glide/Commands/IStreamCommands.cs
```diff
@@ -0,0 +1,463 @@
...
+    /// </summary>
+    /// <seealso href="https://valkey.io/commands/xadd"/>
+    /// <param name="key">The key of the stream.</param>
+    /// <param name="streamPairs">The field-value pairs to add.</param>
+    /// <param name="messageId">The message ID. Use null or "*" for auto-generation.</param>
+    /// <param name="maxLength">The maximum length of the stream. If specified, old entries will be trimmed.</param>
+    /// <param name="useApproximateMaxLength">If true, uses approximate trimming (~) for better performance.</param>
+    /// <param name="flags">The flags to use for this operation. Currently flags are ignored.</param>
+    /// <returns>The ID of the added entry.</returns>
+    Task<ValkeyValue> StreamAddAsync(ValkeyKey key, NameValueEntry[] streamPairs, ValkeyValue? messageId, int? maxLength, bool useApproximateMaxLength, CommandFlags flags);
```
currantw: You've got `Errors.RequestException` in the documentation for the remaining functions. Might as well add it to these two for consistency? 🤷

sources/Valkey.Glide/Commands/IStreamCommands.cs
```diff
@@ -0,0 +1,463 @@
...
+    /// <returns>An array of stream entries, or an empty array if no entries are available.</returns>
+    /// <exception cref="Errors.RequestException">Thrown if the command fails to execute on the server.</exception>
+    Task<StreamEntry[]> StreamReadAsync(ValkeyKey key, ValkeyValue position, int? count = null, int? block = null, CommandFlags flags = CommandFlags.None);
+
+    /// <summary>
+    /// Reads entries from multiple streams starting from given positions.
+    /// </summary>
+    /// <seealso href="https://valkey.io/commands/xread"/>
+    /// <param name="streamPositions">Array of stream keys and their starting positions.</param>
+    /// <param name="count">The maximum number of entries to return per stream.</param>
```
currantw: `StackExchange.Redis` uses `countPerStream`. Perhaps a bit more clear? 🤷

sources/Valkey.Glide/Commands/IStreamCommands.cs
```diff
@@ -0,0 +1,463 @@
...
+    /// <returns>True if the group was created, false otherwise.</returns>
+    Task<bool> StreamCreateConsumerGroupAsync(ValkeyKey key, ValkeyValue groupName, ValkeyValue? position, CommandFlags flags);
+
+    /// <summary>
+    /// Creates a consumer group for a stream.
+    /// </summary>
+    /// <seealso href="https://valkey.io/commands/xgroup-create"/>
+    /// <param name="key">The key of the stream.</param>
+    /// <param name="groupName">The consumer group name.</param>
+    /// <param name="position">The position from which to start reading. Use "$" for new messages, "0" for all messages.</param>
```
currantw: Probably want to specify the default behaviour?

sources/Valkey.Glide/Commands/IStreamCommands.cs
```diff
@@ -0,0 +1,463 @@
...
+    /// <seealso href="https://valkey.io/commands/xreadgroup"/>
+    /// <param name="key">The key of the stream.</param>
+    /// <param name="groupName">The consumer group name.</param>
+    /// <param name="consumerName">The consumer name.</param>
+    /// <param name="position">The position from which to read. Use ">" for new messages only.</param>
+    /// <param name="count">The maximum number of entries to return.</param>
+    /// <param name="noAck">If true, messages are not added to the pending entries list.</param>
+    /// <param name="flags">The flags to use for this operation. Currently flags are ignored.</param>
+    /// <returns>An array of stream entries.</returns>
+    Task<StreamEntry[]> StreamReadGroupAsync(ValkeyKey key, ValkeyValue groupName, ValkeyValue consumerName, ValkeyValue? position = null, int? count = null, bool noAck = false, CommandFlags flags = CommandFlags.None);
```
currantw: `StackExchange.Redis` has a StreamReadGroupAsync` method with parameter `claimMinIdleTime`. Why do we not (seem) to have an equivalent one?

sources/Valkey.Glide/Commands/IStreamCommands.cs
```diff
@@ -0,0 +1,463 @@
...
+    /// Trims the stream by minimum ID.
+    /// </summary>
+    /// <seealso href="https://valkey.io/commands/xtrim"/>
+    /// <param name="key">The key of the stream.</param>
+    /// <param name="minId">Trim entries with IDs lower than this.</param>
+    /// <param name="useApproximateMaxLength">If true, uses approximate trimming (~) for better performance.</param>
+    /// <param name="limit">The maximum number of entries to trim.</param>
+    /// <param name="flags">The flags to use for this operation. Currently flags are ignored.</param>
+    /// <returns>The number of entries deleted.</returns>
+    Task<long> StreamTrimByMinIdAsync(ValkeyKey key, ValkeyValue minId, bool useApproximateMaxLength = false, long? limit = null, CommandFlags flags = CommandFlags.None);
```
currantw: `StackExchange.Redis` has a `StreamTrimMode mode` argument. Are we not supporting this?

sources/Valkey.Glide/Internals/Request.StreamCommands.cs
```diff
@@ -0,0 +1,868 @@
...
+        List<GlideString> args = [key.ToGlideString()];
+
+        // Add NOMKSTREAM if specified
+        if (noMakeStream)
+        {
+            args.Add("NOMKSTREAM".ToGlideString());
+        }
+
+        // Add trimming options (MAXLEN or MINID)
+        if (maxLength.HasValue)
```
currantw: Are we actually enforcing the fact that `maxLength` and `minId`? It seems like, if both were provided under the current implementation, we would just silently continue with `maxLength`?

sources/Valkey.Glide/Internals/Request.StreamCommands.cs
```diff
@@ -0,0 +1,868 @@
...
+
+        args.Add("STREAMS".ToGlideString());
+        foreach (var sp in streamPositions)
+        {
+            args.Add(sp.Key.ToGlideString());
+        }
+        foreach (var sp in streamPositions)
+        {
+            args.Add(sp.Position.ToGlideString());
+        }
```
currantw: Can we combine these into a single method (that takes `StreamPosition[]`) to reduce duplication?

sources/Valkey.Glide/BaseClient.StreamCommands.cs
```diff
@@ -0,0 +1,236 @@
...
+    public async Task<StreamEntry[]> StreamRangeAsync(ValkeyKey key, ValkeyValue? start = null, ValkeyValue? end = null, int? count = null, Order order = Order.Ascending, CommandFlags flags = CommandFlags.None)
+    {
+        GuardClauses.ThrowIfCommandFlags(flags);
+        return await Command(Request.StreamRangeAsync(key, start ?? "-", end ?? "+", count, order));
+    }
+
+    public async Task<bool> StreamCreateConsumerGroupAsync(ValkeyKey key, ValkeyValue groupName, ValkeyValue? position, CommandFlags flags)
+    {
+        GuardClauses.ThrowIfCommandFlags(flags);
+        return await StreamCreateConsumerGroupAsync(key, groupName, position, false, null, flags);
```
currantw: Same comment about duplicate `ThrowIfCommandFlags` calls.

sources/Valkey.Glide/BaseClient.StreamCommands.cs
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    public async Task<long> StreamDeleteConsumerAsync(ValkeyKey key, ValkeyValue groupName, ValkeyValue consumerName, CommandFlags flags = CommandFlags.None)
+    {
+        GuardClauses.ThrowIfCommandFlags(flags);
+        return await Command(Request.StreamDeleteConsumerAsync(key, groupName, consumerName));
+    }
+    public async Task<StreamEntry[]> StreamReadGroupAsync(ValkeyKey key, ValkeyValue groupName, ValkeyValue consumerName, ValkeyValue? position, int? count, CommandFlags flags)
+    {
+        GuardClauses.ThrowIfCommandFlags(flags);
```
currantw: Nit. Same comment about `ThrowIfCommandFlags` plus a missing newline. A few more of both in this file. Might as well cleanup while we can 🤷 .

```suggestion
    }
    
    public async Task<StreamEntry[]> StreamReadGroupAsync(ValkeyKey key, ValkeyValue groupName, ValkeyValue consumerName, ValkeyValue? position, int? count, CommandFlags flags)
    {
```

sources/Valkey.Glide/abstract_Enums/StreamTrimMode.cs
```diff
@@ -0,0 +1,26 @@
+// Copyright Valkey GLIDE Project Contributors - SPDX Identifier: Apache-2.0
+
+namespace Valkey.Glide;
+
+/// <summary>
+/// Determines how stream trimming works.
+/// </summary>
+public enum StreamTrimMode
```
currantw: Where/how is this Enum used? I wasn't able to find any usages... 🤔

sources/Valkey.Glide/StreamClaimOptions.cs
```diff
@@ -0,0 +1,39 @@
+// Copyright Valkey GLIDE Project Contributors - SPDX Identifier: Apache-2.0
+
+namespace Valkey.Glide;
+
+/// <summary>
+/// Options for XCLAIM command.
+/// </summary>
+public class StreamClaimOptions
```
currantw: Where is this class used? Same question for `StreamGroupOptions`, `StreamPendingOptions`, `StreamReadGroupOptions`, and `StreamTrimResult`.

tests/Valkey.Glide.IntegrationTests/StreamCommandTests.cs
```diff
@@ -0,0 +1,799 @@
...
+        await client.StreamAddAsync(key, "field", "value1");
+        await client.StreamAddAsync(key, "field", "value2");
+        await client.StreamAddAsync(key, "field", "value3");
+
+        // Read in ascending order first to verify entries exist
+        StreamEntry[] ascEntries = await client.StreamRangeAsync(key);
+        Assert.Equal(3, ascEntries.Length);
+
+        // Read in descending order (most recent first) - XREVRANGE uses + to -
+        StreamEntry[] entries = await client.StreamRangeAsync(key, start: "+", end: "-", order: Order.Descending);
```
currantw: Hmmm. Interesting! In the current implementation `BaseClient.StreamCommand`, we always default `start` to `-` and `end` to `+` if they are not provided. But based on this behaviour, don't we need to check the order, and swap these default values if the order is `Descending`?

tests/Valkey.Glide.IntegrationTests/StreamCommandTests.cs
```diff
@@ -0,0 +1,799 @@
...
+        // Verify group was created by checking group info
+        StreamGroupInfo[] groups = await client.StreamGroupInfoAsync(key);
+        Assert.Single(groups);
+        Assert.Equal("mygroup", groups[0].Name);
+
+        // EntriesRead should be 10 if server supports it (Valkey 7.0+)
+        if (groups[0].EntriesRead is long entriesRead)
+        {
+            Assert.Equal(10L, entriesRead);
+        }
```
currantw: Why if this `if` check needed?

tests/Valkey.Glide.IntegrationTests/StreamCommandTests.cs
```diff
@@ -0,0 +1,799 @@
...
+        // Get full stream info
+        Dictionary<string, object> info = await client.StreamInfoFullAsync(key);
+
+        // Verify basic fields
+        Assert.True(info.ContainsKey("length"));
+        Assert.True(info.ContainsKey("entries"));
+        Assert.True(info.ContainsKey("groups"));
+
+        // Verify length
+        long length = info["length"] is GlideString gs ? long.Parse(gs.ToString()) : (long)info["length"];
```
currantw: Why is there differences in what type we are returning here? Is this coming directly from Valkey itself?

currantw (commented):
Oooof! That's a big PR. Can't say I reviewed _every_ line, but did a pretty comprehensive look through. Let me know if anything is unclear.

sources/Valkey.Glide/Commands/IStreamCommands.cs
```diff
@@ -0,0 +1,463 @@
...
+    /// <summary>
+    /// Trims the stream to a specified size.
+    /// </summary>
+    /// <seealso href="https://valkey.io/commands/xtrim"/>
+    /// <param name="key">The key of the stream.</param>
+    /// <param name="maxLength">The maximum length of the stream.</param>
+    /// <param name="useApproximateMaxLength">If true, uses approximate trimming (~) for better performance.</param>
+    /// <param name="flags">The flags to use for this operation. Currently flags are ignored.</param>
+    /// <returns>The number of entries deleted.</returns>
+    Task<long> StreamTrimAsync(ValkeyKey key, int maxLength, bool useApproximateMaxLength, CommandFlags flags);
```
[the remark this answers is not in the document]
alexr-bq: Not supported by Valkey

sources/Valkey.Glide/Commands/IStreamCommands.cs
alexr-bq: Not supported by valkey.

sources/Valkey.Glide/Commands/IStreamCommands.cs
```diff
@@ -0,0 +1,463 @@
...
+
+    /// <summary>
+    /// Returns detailed stream information including entries, groups, and consumers.
+    /// </summary>
+    /// <seealso href="https://valkey.io/commands/xinfo-stream"/>
+    /// <param name="key">The key of the stream.</param>
+    /// <param name="count">Maximum number of entries to return. If null, returns all entries.</param>
+    /// <param name="flags">The flags to use for this operation. Currently flags are ignored.</param>
+    /// <returns>Dictionary containing detailed stream information with keys like "length", "entries", "groups", etc.</returns>
+    Task<Dictionary<string, object>> StreamInfoFullAsync(ValkeyKey key, int? count = null, CommandFlags flags = CommandFlags.None);
```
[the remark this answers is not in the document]
alexr-bq: Added it for matching command support with the other clients. I am not sure that there is information here that isn't available through a combination of other xinfo commands. Let me know if you think we should remove it.

sources/Valkey.Glide/Internals/Request.StreamCommands.cs
alexr-bq: Will update with proper enforcement.

sources/Valkey.Glide/Commands/IStreamCommands.cs
alexr-bq: They do, I'm a little confused what you mean here

currantw: Sorry, that was _very_ unclear! 🤦 

The following method include this documentation for `messageId`: "Use null or "*" for auto-generation, or "&lt;ms&gt;-*" to auto-generate sequence number for a specific timestamp."
- `StreamAddAsync(ValkeyKey key, NameValueEntry[] streamPairs, ValkeyValue? messageId, int? maxLength, bool useApproximateMaxLength, CommandFlags flags)`
- `StreamAddAsync(ValkeyKey key, ValkeyValue streamField, ValkeyValue streamValue, ValkeyValue? messageId = null, long? maxLength = null, bool useApproximateMaxLength = false, long? limit = null, bool noMakeStream = false, CommandFlags flags = CommandFlags.None)`
- `StreamAddAsync(ValkeyKey key, NameValueEntry[] streamPairs, ValkeyValue? messageId = null, long? maxLength = null, bool useApproximateMaxLength = false, long? limit = null, bool noMakeStream = false, CommandFlags flags = CommandFlags.None)`

While the description for the first one (`StreamAddAsync(ValkeyKey key, ValkeyValue streamField, ValkeyValue streamValue, ValkeyValue? messageId, int? maxLength, bool useApproximateMaxLength, CommandFlags flags)`) does not include the "or "&lt;ms&gt;-*" to auto-generate sequence number for a specific timestamp" portion'

Is that intentional or just a bit of inconsistent documentation?

sources/Valkey.Glide/Commands/IStreamCommands.cs
```diff
@@ -0,0 +1,463 @@
...
+    /// </summary>
+    /// <seealso href="https://valkey.io/commands/xread"/>
+    /// <param name="key">The key of the stream.</param>
+    /// <param name="position">The position from which to start reading. Use "0-0" to read from the beginning.</param>
+    /// <param name="count">The maximum number of entries to return.</param>
+    /// <param name="block">The maximum time to block waiting for entries (in milliseconds). Use null for non-blocking.</param>
+    /// <param name="flags">The flags to use for this operation. Currently flags are ignored.</param>
+    /// <returns>An array of stream entries, or an empty array if no entries are available.</returns>
+    /// <exception cref="Errors.RequestException">Thrown if the command fails to execute on the server.</exception>
+    Task<StreamEntry[]> StreamReadAsync(ValkeyKey key, ValkeyValue position, int? count = null, int? block = null, CommandFlags flags = CommandFlags.None);
```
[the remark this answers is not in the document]
currantw: Sounds good. Thanks.

sources/Valkey.Glide/Commands/IStreamCommands.cs
currantw: Arggg. Okay sounds good.

sources/Valkey.Glide/Commands/IStreamCommands.cs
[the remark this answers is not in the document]
currantw: Hmmmm. I am not sure how we are handling these cases?

In general, I _think_ my preference would be to keep the interface as small as possible. So long as this functionality is still available another way and we are consistent with `StackExchange.Redis`, that seems fine to me. 

There might not be a _good_ reason that it was added to other clients. Perhaps another client simply added it for compatibility with some existing library in _that_ language, and then it is being copied blindly in all our client in other languages? As we've seen from `StackExchange.Redis`, sometimes the existing interface that we are trying to mirror is not perfectly designed -- and we would want to avoid replicating that in other languages. 

So unless there is a clear reason to add this in C# (it is a core command, it is specifically requested, or a user is likely to expect it to exist), my gut feeling would be to avoid it?

@jbrinkman / @jduo do you two have any thoughts on this?

sources/Valkey.Glide/Internals/Request.StreamCommands.cs
currantw: Great, thank you!

sources/Valkey.Glide/Internals/Request.StreamCommands.cs
```diff
@@ -0,0 +1,868 @@
...
+            args.Add(id.ToGlideString());
+        }
+        return new(RequestType.XDel, [.. args], false, response => response);
+    }
+
+    public static Cmd<long, long> StreamTrimAsync(ValkeyKey key, long? maxLength, ValkeyValue minId, bool useApproximateTrimming, long? limit)
+    {
+        List<GlideString> args = [key.ToGlideString()];
+
+        if (maxLength.HasValue)
```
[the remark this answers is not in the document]
currantw: Lovely. Thank you. Always better to enforce this way if possible.

sources/Valkey.Glide/Commands/IStreamCommands.cs
```diff
@@ -0,0 +1,480 @@
...
+public interface IStreamCommands
+{
+    /// <summary>
+    /// Appends a new entry to a stream with a single field-value pair.
+    /// </summary>
+    /// <seealso href="https://valkey.io/commands/xadd"/>
+    /// <param name="key">The key of the stream.</param>
+    /// <param name="streamField">The field name.</param>
+    /// <param name="streamValue">The field value.</param>
+    /// <param name="messageId">The message ID. Use null or "*" for auto-generation.</param>
```
currantw: Nit. Should we reference the constant here instead of the hard-coded string? I assume that user would generally want to use the constant? Same for all other similar constants in documentation (including `~` below in this method).

currantw (commented):
Just one comment

currantw (approved):
Thanks for labouring through all my comments.
Look good to me ✅

yipin-chen (approved):

Issue #154, opened 2025-11-25
[C#][Flaky Test] PubSubFFIMemoryLeakTests.ProcessVariousMessageSizes_NoMemoryLeak_ConsistentBehavior


xShinnRyuu:
### Test Name

Valkey.Glide.UnitTests.PubSubFFIMemoryLeakTests.ProcessVariousMessageSizes_NoMemoryLeak_ConsistentBehavior

### Test Location

valkey-glide-csharp/valkey-glide-csharp/tests/Valkey.Glide.UnitTests/PubSubFFIMemoryLeakTests.cs:line 121

### Failure Permlink

[Valkey.Glide.UnitTests.PubSubFFIMemoryLeakTests.ProcessVariousMessageSizes_NoMemoryLeak_ConsistentBehavior
](https://github.com/valkey-io/valkey-glide-csharp/actions/runs/19622387773/job/56184863757#step:7:979)

### System Information

aarch64-unknown-linux-gnu

### Language and Version

net8.0

### Engine Version

valkey 8.0

### Logs

```
Failed Valkey.Glide.UnitTests.PubSubFFIMemoryLeakTests.ProcessVariousMessageSizes_NoMemoryLeak_ConsistentBehavior [34 ms]
  Error Message:
   Excessive memory growth for 10-byte messages: 142,496 bytes, expected max: 30,000 bytes
  Stack Trace:
     at Valkey.Glide.UnitTests.PubSubFFIMemoryLeakTests.ProcessVariousMessageSizes_NoMemoryLeak_ConsistentBehavior() in /home/runner/work/valkey-glide-csharp/valkey-glide-csharp/tests/Valkey.Glide.UnitTests/PubSubFFIMemoryLeakTests.cs:line 121
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)

```

### Glide Version

main

PR #165, opened 2025-12-17
Fix(tests): Redesign memory leak test for robustness (Issue #154)

hank95179:
### Motivation
The previous implementation of `ProcessVariousMessageSizes_NoMemoryLeak_ConsistentBehavior` relied on a fixed formula to predict expected memory growth (`messageSize * iterations * 3`). This approach proved to be flawed and caused flaky test failures, particularly with small message sizes.

When messages are small, the managed object overhead (wrapping, headers, internal structures) often vastly exceeds the size of the payload itself. We observed instances where the test expected ~30,000 bytes of growth but the actual memory usage exceeded 1,000,000 bytes due to this overhead, leading to false positives.

### Changes
This PR completely redesigns the memory leak detection strategy to be robust against object overhead and GC variability:

1.  **Baseline vs. Delta Approach:** Instead of predicting absolute memory usage, the test now:
    *   **Step 1:** Runs the processing loop once and performs a full GC to establish a **stable baseline**. This ensures all one-time static allocations and long-lived objects are accounted for.
    *   **Step 2:** Runs the exact same loop a second time.
    *   **Step 3:** Measures the **memory growth** between the baseline and the second run.

2.  **Fixed Threshold:** We now use a constant `leakThresholdBytes` (50KB) to detect unreclaimed memory. If the code is not leaking, the memory usage after the second run (cleaned by GC) should be nearly identical to the baseline, regardless of the message size or object overhead.

3.  **Refactoring:** Converted the test from a single `[Fact]` loop to a `[Theory]` with `[InlineData]`. This improves reporting by isolating failures to specific message sizes.

### Impact
*   Eliminates false positives caused by underestimating .NET object overhead for small messages.
*   Focuses the test on detecting *actual* memory leaks (unreclaimed memory between runs) rather than validating GC behavior or object size.


yipin-chen (approved):
LGTM

tests/Valkey.Glide.UnitTests/PubSubFFIMemoryLeakTests.cs
```diff
@@ -76,61 +76,72 @@ public void ProcessLargeVolumeMessages_NoMemoryLeak_MemoryUsageRemainsBounded()
...
+        GC.Collect();
+        return GC.GetTotalMemory(true);
+    }
 
-            // Memory growth should be reasonable for the message size
-            // Allow 3x overhead to account for GC variance and object headers
-            long expectedMaxGrowth = messageSize * iterationsPerSize * 3;
-            Assert.True(sizeTestGrowth < expectedMaxGrowth,
-                $"Excessive memory growth for {messageSize}-byte messages: {sizeTestGrowth:N0} bytes, expected max: {expectedMaxGrowth:N0} bytes");
+    private static void RunMessageProcessingLoop(string message, string channel, string? pattern, int iterations)
```
currantw: It appears that this method is unused? I assume that you intended to use it to remove the loops in the method above?

PR #166, opened 2025-12-17
Fix: Improve performance test logic for PubSub

hank95179:
### Motivation
The existing performance test for PubSub (`ChannelBasedProcessing_LongRunning_MaintainsStablePerformance`) suffered from an imprecise rate-limiting mechanism. The previous approach of sending messages in fixed batches (`targetRate / 10`) combined with a `Thread.Sleep(100)` interval could lead to an inconsistent message send rate, especially under varying system loads or during initial test phases. This made it difficult to accurately measure and assert stable performance over a long duration.

The primary issue was that the `Thread.Sleep(100)` might pause for longer or shorter than intended, and sending messages only at these fixed intervals meant the test wasn't continuously adjusting to maintain the desired `targetRate`.

### Changes
This PR refines the rate-limiting logic within the `ChannelBasedProcessing_LongRunning_MaintainsStablePerformance` test to ensure a more consistent and accurate message sending rate:

1.  **Warm-up Phase Refinement:** The warm-up phase is now explicitly defined to run for a duration (1 second), ensuring that the system stabilizes before performance measurements begin. Counters are reset after warm-up for clean measurement.

2.  **Robust Rate-Limiting:** The message sending loop has been replaced with a more robust, time-based mechanism:
    *   It now calculates the `expectedMessagesSent` based on the `elapsedSeconds` and the `targetRate`.
    *   It continuously sends messages in a `while` loop until `sentMessageCount` catches up to the `expectedMessagesSent`. This ensures that messages are sent closer to the actual target rate, compensating for any delays in previous cycles.
    *   A `Thread.Sleep(100)` is still used within the main loop, but its role has changed. It now primarily acts to prevent a tight loop from consuming 100% CPU when the system is operating faster than the `targetRate`, rather than being the sole mechanism for rate control.

3.  **Message Content Adjustment:** The message content now includes `sentMessageCount` to ensure unique message identities for better traceability during debugging.

### Impact
*   **More Accurate Performance Measurement:** The test now achieves a more precise and stable message sending rate, leading to more reliable performance metrics.
*   **Reduced Flakiness:** The improved rate-limiting reduces test flakiness caused by inconsistent message delivery, providing a more trustworthy assessment of PubSub's long-running performance.
*   **Clearer Intent:** The code better reflects the intention of maintaining a consistent throughput over time.

### DIFF
--- a/.github/json_matrices/version-matrix.json
+++ b/.github/json_matrices/version-matrix.json
@@ -1,4 +1,4 @@
 {
-    "versions": ["8.0", "6.0"],
+    "versions": ["8.0"],
     "always-run-versions": ["8.0"]
 }
--- a/.github/workflows/api-compatibility.yml
+++ b/.github/workflows/api-compatibility.yml
@@ -30,7 +30,7 @@ jobs:
                   submodules: true
 
             - name: Setup .NET
-              uses: actions/setup-dotnet@v4
+              uses: actions/setup-dotnet@v5
               with:
                   dotnet-version: |
                       8.x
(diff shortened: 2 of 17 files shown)


currantw (approved):
Thanks @hank95179 for this fix to a tricky problem!
One small question, but looks good to me. ✅

yipin-chen (approved):

Issue #193, opened 2026-01-22
Add Dynamic Pub/Sub


currantw:
## Summary

Add support for dynamic pub/sub subscriptions.

### Overview

The Valkey GLIDE C# client currently only supports **static pub/sub**, where all subscriptions must be configured when creating the client:

```csharp
var pubsubConfig = new StandalonePubSubSubscriptionConfig()
    .WithChannel("channel1")
    .WithPattern("news.*");

var config = new StandaloneClientConfigurationBuilder()
    .WithPubSubSubscriptions(pubsubConfig)
    .Build();

await using var client = await GlideClient.CreateClient(config);
```
Since the initial C# implementation of pub/sub, the Rust core has been updated to support **dynamic pub/sub**.

### Dynamic Pub/Sub Support

The following was added to the Rust core to support dynamic pub/sub:

- New protobuf request types:
    - `Subscribe` / `Unsubscribe`  for exact channels.
    - `PSubscribe` / `PUnsubscribe` for patterns.
    - `SSubscribe` / `SUnsubscribe` for sharded channels (cluster only).
    - `GetSubscriptions` for state query (desired and actual subscriptions).
- A _pub/sub synchronizer_ that automatically handles subscription updates, reconnections, and topology changes. It tracks desired vs. actual subscriptions (i.e. subscription changes that have not been propagates to the server(s)).

❗**Note**: The Rust core also supports blocking operation for synchronous pub/sub (e.g. protobuf request type `SubscribeBlocking`). Since the C# client as asynchronous, these have been omitted above.

### Current C# Interface

| Class | Type | Command | Signature |
|--------|--------|--------|--------|
| `IPubSubCommands` | State | [PUBSUB CHANNELS](https://valkey.io/commands/pubsub-channels/) | `Task<string[]> PubSubChannelsAsync(CommandFlags flags)` |
| | State | [PUBSUB CHANNELS](https://valkey.io/commands/pubsub-channels/) | `Task<string[]> PubSubChannelsAsync(string pattern, CommandFlags flags)` |
|  | State | [PUBSUB NUMSUB](https://valkey.io/commands/pubsub-numsub/) | `Task<Dictionary<string, long>> PubSubNumSubAsync(string[] channels, CommandFlags flags)` |
|  | State | [PUBSUB NUMPAT](https://valkey.io/commands/pubsub-numpat/) | `Task<long> PubSubNumPatAsync(CommandFlags flags)` |
[...]

PR #197, opened 2026-01-22
feat: dynamic pub/sub 1 - update existing interface

currantw:
## Summary

Update existing pub/sub interface. Preliminary work for dynamic pub/sub (#193).

## Changes

1) Rename existing method `IPubSubClusterCommands.PublishAsync(string channel, string message, bool sharded, CommandFlags flags)` to `SPublishAsync` and remove `sharded` boolean parameter.
  
> **Rationale**: `PublishAsync(string channel, string message, CommandFlags flags)` already provides an interface for publishing non-sharded message with cluster clients. `SPublishAsync` also more closely matches the corresponding Valkey [SPUBLISH](https://valkey.io/commands/spublish/) command and is likely to be more intuitive for users who are familiar with Valkey/Redis.

2) Combine existing interface methods `IPubSubClusterCommands.PublishAsync(string channel, string message, CommandFlags flags)` and `IPubSubStandaloneCommands.PublishAsync(string channel, string message, CommandFlags flags)` into a single `IPubSubCommands.PublishAsync(string channel, string message, CommandFlags flags)` method.

> **Rationale**: These two signatures are _identical_. They should be moved to the parent to reduce duplication.

## Test Strategy

✅ Pipeline passes.

There are existing tests for these methods. No need to add new tests.

## Related Issues

- #193

### DIFF
--- a/sources/Valkey.Glide/Commands/IPubSubStandaloneCommands.cs
+++ b/sources/Valkey.Glide/Commands/IPubSubStandaloneCommands.cs
@@ -9,5 +9,5 @@ namespace Valkey.Glide.Commands;
 /// </summary>
 public interface IPubSubStandaloneCommands : IPubSubCommands
 {
-    // TODO #183: Add support for dynamic PubSub.
+    // TODO #193: Add support for dynamic PubSub.
 }


sources/Valkey.Glide/Commands/IPubSubStandaloneCommands.cs
```diff
@@ -9,21 +9,5 @@ namespace Valkey.Glide.Commands;
...
-    /// <remarks>
-    /// <example>
-    /// <code>
-    /// long subscriberCount = await client.PublishAsync("news", "Breaking news!");
-    /// Console.WriteLine($"Message delivered to {subscriberCount} subscribers");
-    /// </code>
-    /// </example>
-    /// </remarks>
-    Task<long> PublishAsync(string channel, string message, CommandFlags flags = CommandFlags.None);
+    // TODO #183: Add support for dynamic PubSub.
```
currantw: Leaving this (empty) interface class because I expect to add some methods to it as part of later pub/sub work.

xShinnRyuu (approved):

PR #225, opened 2026-03-17
feat(core): Add Valkey Search Support

### LINKED ISSUE #229: (feat): Add Valkey Search Support
We need to add Valkey Search support so that users can integrate GLIDE into their agentic workflows. Currently, C# does not support any of the `FT.*` commands.

I can work on this

edlng:
## Summary

Currently, we need to wait until https://github.com/valkey-io/valkey-glide/pull/5571 gets merged for the required GLIDE core changes.

This adds Valkey Search command support to the C# GLIDE client: 12 FT.* commands, full options models, response parsing, and test coverage.

## Issue link

Closes #229 

## Features and behavior changes

Adds the `IVectorSearchCommands` interface and `BaseClient.VectorSearchCommands.cs` with the following commands:

- Index management: `FT.CREATE`, `FT.DROPINDEX`, `FT._LIST`, `FT.INFO`
- Query: `FT.SEARCH`, `FT.AGGREGATE`
- Alias management: `FT.ALIASADD`, `FT.ALIASDEL`, `FT.ALIASUPDATE`, `FT._ALIASLIST`

`FT.CREATE` supports HASH and JSON index types with TEXT, TAG, NUMERIC, and VECTOR (FLAT/HNSW) field types. Index-level options cover prefixes, score, language, stop words, skip initial scan, min stem size, no offsets, and suffix trie control.

`FT.SEARCH` supports NOCONTENT, VERBATIM, SORTBY with WITHSORTKEYS, RETURN fields with aliases, LIMIT, PARAMS, TIMEOUT, DIALECT, INORDER, SLOP, and cluster-mode options (ALLSHARDS/SOMESHARDS, CONSISTENT/INCONSISTENT). WITHSORTKEYS is silently stripped when NOCONTENT is set to match server behavior.

`FT.AGGREGATE` supports LOAD (all or specific fields), pipeline clauses (GROUPBY with reducers, SORTBY, FILTER, APPLY, LIMIT), VERBATIM, TIMEOUT, PARAMS, and DIALECT.

`FT.INFO` supports LOCAL, PRIMARY, and CLUSTER scope with shard scope and consistency options.

> Note: `AGGREGATE` and `ALIAS` commands are still in progress on the Valkey Search side and won't work against the server yet. A fix will follow once the next server release lands.

## Implementation

- `IVectorSearchCommands` — public async interface with XML docs and valkey.io links
- `BaseClient.VectorSearchCommands.cs` — thin partial delegating to `Request` builders
- `Request.VectorSearchCommands.cs` — argument construction and response parsers (`ParseFtSearchResponse`, `ParseFtAggregateResponse`, `ParseFtInfoResponse`) with recursive `ConvertFtValue` for nested GlideString conversion
- `Commands/Options/FtCreateOptions.cs` — field types (`TextField`, `TagField`, `NumericField`, `VectorFieldFlat`, `VectorFieldHnsw`), `IField` interface, `FtCreateOptions` with mutual exclusivity validation
- `Commands/Options/FtSearchOptions.cs` — search options with shard scope/consistency enums
- `Commands/Options/FtAggregateOptions.cs` — aggregate pipeline clause types (`FtAggregateGroupBy`, `FtAggregateSortBy`, `FtAggregateFilter`, `FtAggregateApply`, `FtAggregateLimit`)
- `Commands/Options/FtInfoOptions.cs` — info scope/shard/consistency options
- `Commands/Options/FtSearchResult.cs` — `FtSearchResult` and `FtSearchDocument` response types
- `Commands/Constants/Constants.cs` — 46 new keyword constants for FT.* commands
- `Internals/Cmd.cs` — 12 new `RequestType` → command name mappings

All enum-to-string serialization uses defined constants via switch expressions rather than `ToString()`.

## Limitations

- JSON module commands (JSON.SET) are called via `CustomCommand` in integration tests pending native JSON command support.
- `FT.SEARCH` response parsing assumes the glide-core normalized format (`[count, {key: fields}]`).

## Testing

12 unit tests (`VectorSearchCommandTests.cs`) cover argument serialization for all commands and option combinations, including mutual exclusivity validation errors.

~25 integration tests (`VssCommandTests.cs`) cover index creation (HNSW, FLAT, JSON, multiple prefixes, field options), search (basic, NOCONTENT, SORTBY, WITHSORTKEYS, DIALECT, KNN vector), aggregate (GROUPBY with reducers, APPLY, SORTBY, JSON bicycles), FT.INFO (LOCAL/PRIMARY/CLUSTER scope), alias CRUD, drop index, FT._LIST, and error cases.

Integration tests skip gracefully when the Search module isn't available on the server. All tests pass on net8.0. Lint build (`--configuration Lint`) passes with 0 warnings.

## Checklist

- [x] This pull request is related to one issue.
- [x] Destination branch is correct — `main` or release.


currantw (commented):
Preliminary review of the API.

alexr-bq (approved):

CHANGELOG.md
```diff
@@ -6,3 +6,4 @@
 
 - Add support for Windows CI and testing with WSL (#184)
 - Add StackExchange.Redis compatible pub/sub API (#202)
+- Add Valkey Search (VSS) command support: FT.CREATE, FT.DROPINDEX, FT._LIST, FT.SEARCH, FT.AGGREGATE, FT.INFO, FT.ALIASADD, FT.ALIASDEL, FT.ALIASUPDATE, FT._ALIASLIST (#225)
```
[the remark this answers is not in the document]
currantw: @Aryex Is there any plan to add this to docs?
@edlng can we raise an issue on `valkey-glide-docs` for this if it doesn't already exist?

In my opinion, we should get into the habit of always updating documentation as part of new features, and we have been trying to avoid putting all this on Alex – so, if you're able to Edward, would be great to do this docs update as part of your scope for adding Valkey Search support. If not, let's at least make sure we have an issue to track it.

sources/Valkey.Glide/Commands/IVectorSearchCommands.cs
```diff
@@ -0,0 +1,100 @@
...
+
+    /// <summary>
+    /// Runs an aggregation pipeline against an index.
+    /// </summary>
+    /// <seealso href="https://valkey.io/commands/ft.aggregate/">valkey.io</seealso>
+    /// <param name="indexName">The name of the index to aggregate.</param>
+    /// <param name="query">The filter query string.</param>
+    /// <param name="options">Optional aggregation parameters.</param>
+    /// <returns>An array of result row dictionaries.</returns>
+    Task<Dictionary<string, object?>[]> FtAggregateAsync(string indexName, string query, FtAggregateOptions? options = null);
```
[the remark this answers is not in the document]
currantw: Thanks I'll take a look! 🙏

sources/Valkey.Glide/Client/IBaseClient.VectorSearchCommands.cs
```diff
@@ -0,0 +1,100 @@
...
+    /// <seealso href="https://valkey.io/commands/ft.info/">valkey.io</seealso>
+    /// <param name="indexName">The name of the index to inspect.</param>
+    /// <param name="options">Optional info parameters.</param>
+    /// <returns>A dictionary of field names to their values describing the index.</returns>
+    Task<Dictionary<string, object?>> FtInfoAsync(string indexName, FtInfoOptions? options = null);
+
+    /// <summary>
+    /// Adds an alias to an existing index.
+    /// </summary>
+    /// <seealso href="https://valkey.io/commands/ft.aliasadd/">valkey.io</seealso>
```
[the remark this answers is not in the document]
currantw: We now have automated link checking, so I think these will cause the pipeline to fail. You can exclude them explicitly in `lychee.toml` – but please make sure that this work (actually enabling these links) is tracked with a GitHub issue and assigned to you? A TODO comment would also be helpful, i.e.

```text
#### lychee.toml
exclude = [
    "example\\.com",    # Example URLs
    "file://.*example",  # Example file URIs

     # Valkey Search URLs
     # TODO #123: Remove from exclude list after Valkey search 1.3 released
     "https://valkey.io/commands/ft.aliasadd/",
     ....
]

File: sources/Valkey.Glide/Commands/Options/FtAggregateOptions.cs [hunk 4]
```diff
@@ -0,0 +1,212 @@
...
+/// Sort order for FT.AGGREGATE SORTBY clause.
+/// </summary>
+/// <seealso href="https://valkey.io/commands/ft.aggregate/">valkey.io</seealso>
+public enum FtAggregateOrderBy
+{
+    /// <summary>Ascending order.</summary>
+    ASC,
+    /// <summary>Descending order.</summary>
+    DESC,
+}
```
currantw: There is an existing `SortOrder` enum. It includes a default value. I think we should probably use it here instead (I believe we use it for all other sort orders right now).

sources/Valkey.Glide/Commands/Options/FtAggregateOptions.cs
```diff
@@ -0,0 +1,212 @@
...
+
+/// <summary>
+/// Base interface for all FT.AGGREGATE pipeline clauses.
+/// </summary>
+/// <seealso href="https://valkey.io/commands/ft.aggregate/">valkey.io</seealso>
+public interface IFtAggregateClause
+{
+    /// <summary>Returns the command arguments for this clause.</summary>
+    GlideString[] ToArgs();
```
currantw: This is only every used internally, so I think we should mark it `internal` so it's not visible to users.

sources/Valkey.Glide/Commands/Options/FtAggregateOptions.cs
```diff
@@ -0,0 +1,212 @@
...
+public class FtAggregateLimit(int offset, int count) : IFtAggregateClause
+{
+    /// <summary>Number of results to skip.</summary>
+    public int Offset { get; } = offset;
+    /// <summary>Number of results to return.</summary>
+    public int Count { get; } = count;
+
+    /// <inheritdoc/>
+    public GlideString[] ToArgs() => [ValkeyLiterals.LIMIT, Offset.ToString(), Count.ToString()];
+}
```
currantw: This is similar to a portion of the existing `RangeOptions` class. I think it might make sense to match those implementation details?

- Use `long` for `Offset` and `Count` properties.
- Add private defaults for `Offset` and `Count`.
- Use a parameterless constructor with init-only properties

I think this would be nice for consistency. I would also allow users to avoid needing to specifying _both_ `Offset` and `Count` unless they want to (I assume it will be more common to only specify one or the other than both).

CHANGELOG.md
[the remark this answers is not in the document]
Aryex: We do have a section on the Search modules. However it is part of a generic "Valkey GLIDE modules" guide which only contains code examples. I'm not sure if we need more in depth instructions on the modules or not. Open to splitting it up into their own sections. 

https://glide.valkey.io/how-to/modules-api/

Edit: There's a PR on splitting them up 

https://github.com/valkey-io/valkey-glide-docs/pull/198

sources/Valkey.Glide/Commands/Options/FtAggregateOptions.cs
```diff
@@ -0,0 +1,212 @@
...
+/// <summary>
+/// Filters results using a predicate expression applied post-query.
+/// </summary>
+public class FtAggregateFilter(string expression) : IFtAggregateClause
+{
+    /// <summary>The filter expression.</summary>
+    public string Expression { get; } = expression;
+
+    /// <inheritdoc/>
+    public GlideString[] ToArgs() => [ValkeyLiterals.FILTER, Expression];
```
currantw: As mentioned previously, I think these `ToArgs` methods should all be `internal`?

sources/Valkey.Glide/Commands/Options/FtAggregateOptions.cs
```diff
@@ -0,0 +1,212 @@
...
+}
+
+/// <summary>
+/// A reducer function applied within a GROUPBY clause.
+/// </summary>
+public class FtAggregateReducer(string function)
+{
+    /// <summary>The reduction function name (e.g. "COUNT", "SUM", "TOLIST").</summary>
+    public string Function { get; } = function;
```
currantw: Is this a set number of known functions – i.e. can this be an enum instead?

sources/Valkey.Glide/Commands/Options/FtAggregateOptions.cs
currantw, line 1: I think we should make all our options classes sealed?

sources/Valkey.Glide/Commands/Options/FtAggregateOptions.cs
```diff
@@ -0,0 +1,212 @@
...
+            args.Add(ValkeyLiterals.PARAMS);
+            args.Add((Params.Length * 2).ToString());
+            foreach (var p in Params) { args.Add(p.Key); args.Add(p.Value); }
+        }
+        if (Clauses is not null)
+        {
+            foreach (var clause in Clauses) args.AddRange(clause.ToArgs());
+        }
+        if (Dialect.HasValue) { args.Add(ValkeyLiterals.DIALECT); args.Add(Dialect.Value.ToString()); }
+        return [.. args];
```
currantw: Do you have errors enabled in C# for your IDE? There should be some warning generated here for adding braces for `if` and `foreach`.

sources/Valkey.Glide/Commands/Options/FtAggregateOptions.cs
```diff
@@ -0,0 +1,212 @@
...
+    public string Key { get; } = key;
+    /// <summary>The parameter value.</summary>
+    public string Value { get; } = value;
+}
+
+/// <summary>
+/// Optional arguments for the FT.AGGREGATE command.
+/// </summary>
+/// <seealso href="https://valkey.io/commands/ft.aggregate/">valkey.io</seealso>
+public class FtAggregateOptions
```
currantw: This seems like a class that might benefit from fluent builders (i.e. `internal set`)?

sources/Valkey.Glide/Commands/Options/FtAggregateOptions.cs
```diff
@@ -0,0 +1,212 @@
...
+{
+    /// <summary>Loads all fields declared in the index. Mutually exclusive with <see cref="LoadFields"/>.</summary>
+    public bool LoadAll { get; set; }
+    /// <summary>Loads only the specified fields. Mutually exclusive with <see cref="LoadAll"/>.</summary>
+    public string[]? LoadFields { get; set; }
+    /// <summary>Overrides the module timeout in milliseconds.</summary>
+    public int? Timeout { get; set; }
```
currantw: Let's use `TimeSpan` here instead – that the general pattern through the C# client.

sources/Valkey.Glide/Commands/Options/FtAggregateOptions.cs
```diff
@@ -0,0 +1,212 @@
...
+    /// <summary>Key/value pairs referenced from within the query expression.</summary>
+    public FtAggregateParam[]? Params { get; set; }
+    /// <summary>Pipeline clauses (FILTER, LIMIT, GROUPBY, SORTBY, APPLY) applied in order.</summary>
+    public IFtAggregateClause[]? Clauses { get; set; }
```
currantw: For params,  clauses, and fields, we could provide builder methods like this for convenience perhaps?

```csharp

WithParam(FtAggregateParam param) // Adds a param
WithParams(IEnumerable<FtAggregateParam> params) // Sets params (overwrites any existing)

WithClause(IFtAggregateClause clause) // Adds a clause
WithClauses(IEnumerable<IFtAggregateClause> clauses) // Sets clauses (overwrites any existing) 

WithField(string field); // Adds a field to load
WithFields(IEnumerable<string> fields; // Sets fields (overwrites any existing) 
WithAllFields() // Sets all fields
```

Would also help hide that mutual exclusivity condition for field loading behind the interface, perhaps?

currantw: One more thing! We should use `IEnumerable` throughout for any "collection" inputs (see https://github.com/valkey-io/valkey-glide-csharp/pull/234) to provide more flexibility for users).

sources/Valkey.Glide/Commands/Options/FtAggregateOptions.cs
```diff
@@ -0,0 +1,212 @@
...
+
+    /// <inheritdoc/>
+    public GlideString[] ToArgs()
+    {
+        List<GlideString> args = [ValkeyLiterals.GROUPBY, Properties.Length.ToString()];
+        foreach (var p in Properties) args.Add(p);
+        foreach (var r in Reducers) args.AddRange(r.ToArgs());
+        return [.. args];
+    }
+}
```
currantw, line 95: More general comment – but can you look through all the classes were adding in this PR and see if we can improve the interfaces at all for usability and/or better align with existing ones (they are not terribly consistent right now, but we at least avoid making it worse haha):
- As mentioned in another comment, let's use `IEnumerable`
- I think we should probably prefer immutable types? Either simple constructors, implicit parameterless constructors (where reasonable defaults exist), or fluent builder depending on complexity.
- `sealed` classes or `readonly` structs.

sources/Valkey.Glide/Commands/Options/FtAggregateOptions.cs
```diff
@@ -0,0 +1,212 @@
...
+
+/// <summary>
+/// A single sort property with direction for <see cref="FtAggregateSortBy"/>.
+/// </summary>
+public class FtAggregateSortProperty(string property, FtAggregateOrderBy order)
+{
+    /// <summary>The property name.</summary>
+    public string Property { get; } = property;
+    /// <summary>The sort direction.</summary>
+    public FtAggregateOrderBy Order { get; } = order;
```
currantw: We can use `SortOrder` here, and make this optional with a `SortOrder.Default` default.

sources/Valkey.Glide/Commands/Options/FtAggregateOptions.cs
```diff
@@ -0,0 +1,212 @@
...
+
+/// <summary>
+/// Sorts the pipeline by a list of properties.
+/// </summary>
+public class FtAggregateSortBy(params FtAggregateSortProperty[] properties) : IFtAggregateClause
+{
+    /// <summary>Fields and their sort directions.</summary>
+    public FtAggregateSortProperty[] Properties { get; } = properties;
+    /// <summary>Optimizes sorting by only sorting the n-largest elements.</summary>
+    public int? Max { get; set; }
```
currantw: Can this just be an optional constructor parameter and default to `null`?

sources/Valkey.Glide/Commands/Options/FtCreateOptions.cs
```diff
@@ -0,0 +1,318 @@
...
+}
+
+/// <summary>
+/// Represents the algorithm used for vector similarity search.
+/// </summary>
+/// <seealso href="https://valkey.io/commands/ft.create/">valkey.io</seealso>
+public enum VectorAlgorithm
+{
+    /// <summary>Brute force (flat) algorithm.</summary>
+    FLAT,
```
currantw: I think we generally use `PascalCase` for these parameters.

```suggestion
    Flat,
```

sources/Valkey.Glide/Commands/Options/FtCreateOptions.cs
```diff
@@ -0,0 +1,318 @@
...
+/// <seealso href="https://valkey.io/commands/ft.create/">valkey.io</seealso>
+public enum DistanceMetric
+{
+    /// <summary>Euclidean distance.</summary>
+    L2,
+    /// <summary>Inner product.</summary>
+    IP,
+    /// <summary>Cosine distance.</summary>
+    COSINE,
+}
```
currantw: Do we need to be tied to the underlying Valkey server values? Perhaps something a bit more readable would be better here? Also, use PascalCase for consistency.


```suggestion
/// <summary>
/// Represents the distance metric used to measure similarity between vectors.
/// </summary>
/// <seealso href="https://valkey.io/commands/ft.create/">valkey.io</seealso>
public enum DistanceMetric
{
    /// <summary>Euclidean distance.</summary>
    Euclidean,
    /// <summary>Inner product.</summary>
    InnerProduct,
    /// <summary>Cosine distance.</summary>
    Cosine,
}
```

sources/Valkey.Glide/Commands/Options/FtCreateOptions.cs
```diff
@@ -0,0 +1,318 @@
...
+
+/// <summary>
+/// Represents the vector data type for vector fields.
+/// </summary>
+/// <seealso href="https://valkey.io/commands/ft.create/">valkey.io</seealso>
+public enum VectorDataType
+{
+    /// <summary>32-bit floating point (default).</summary>
+    FLOAT32,
+}
```
currantw: Do we need to expose this for the time being? Since there is currently only one option available, seems like something we don't need in our interface for now – we can just handle this internally, and expose it in the future only when there are other options?

sources/Valkey.Glide/Commands/Options/FtCreateOptions.cs
```diff
@@ -0,0 +1,318 @@
...
+/// Represents the type of the index dataset.
+/// </summary>
+/// <seealso href="https://valkey.io/commands/ft.create/">valkey.io</seealso>
+public enum IndexDataType
+{
+    /// <summary>Data stored in hashes; field identifiers are field names within the hashes.</summary>
+    HASH,
+    /// <summary>Data stored as JSON documents; field identifiers are JSON Path expressions.</summary>
+    JSON,
+}
```
currantw: Same comment about PascalCase. Can we make this change throughout please?


```suggestion
/// <summary>
/// Represents the type of the index dataset.
/// </summary>
/// <seealso href="https://valkey.io/commands/ft.create/">valkey.io</seealso>
public enum IndexDataType
{
    /// <summary>Data stored in hashes; field identifiers are field names within the hashes.</summary>
    Hash,
    /// <summary>Data stored as JSON documents; field identifiers are JSON Path expressions.</summary>
    Json,
}
```

sources/Valkey.Glide/Commands/Options/FtCreateOptions.cs
```diff
@@ -0,0 +1,318 @@
...
+    /// <summary>Whether the field is sortable.</summary>
+    public bool Sortable { get; set; }
+    /// <summary>Disables stemming when indexing the field.</summary>
+    public bool NoStem { get; set; }
+    /// <summary>The importance of this field when calculating result accuracy.</summary>
+    public double? Weight { get; set; }
+    /// <summary>Keeps a suffix trie for the field. Mutually exclusive with <see cref="NoSuffixTrie"/>.</summary>
+    public bool WithSuffixTrie { get; set; }
+    /// <summary>Disables the suffix trie for the field. Mutually exclusive with <see cref="WithSuffixTrie"/>.</summary>
+    public bool NoSuffixTrie { get; set; }
```
currantw: - Can we make this a sealed class with fluent builder methods?
- Can we tweak the interface to reduce/eliminate the mutual exclusivity condition?

currantw: More generally, I don't think we need to strictly match the actual Valkey command syntax. We can (and _should_) diverge from it where it is useful – where it make it more readable or easier for the user.

For example, could we have a `private bool? WithSuffixTrie = null` field, and then a single `WithSuffixTrie(bool? include)` method. Would that eliminate the mutual exclusivity exception completely? (I haven't dug into this particular command in too much detail, so let me know if I'm wrong!)

sources/Valkey.Glide/Commands/Options/FtCreateOptions.cs
```diff
@@ -0,0 +1,318 @@
...
+
+        List<GlideString> args = [Name];
+        if (Alias is not null) { args.Add(ValkeyLiterals.AS); args.Add(Alias); }
+        args.Add(nameof(FieldType.Text).ToUpper());
+        if (NoStem) args.Add(ValkeyLiterals.NOSTEM);
+        if (Weight.HasValue) { args.Add(ValkeyLiterals.WEIGHT); args.Add(Weight.Value.ToString("G")); }
+        if (WithSuffixTrie) args.Add(ValkeyLiterals.WITHSUFFIXTRIE);
+        else if (NoSuffixTrie) args.Add(ValkeyLiterals.NOSUFFIXTRIE);
+        if (Sortable) args.Add(ValkeyLiterals.SORTABLE);
+        return [.. args];
```
currantw: As mentioned in another comment, can you please enable C# warning in your IDE? (I use `muhammad-sammy` extension for Kiro). This should flag a bunch of warning that can be quickly fixed here. Helpful for code quality, and we may eventually bump up the severity.

sources/Valkey.Glide/Commands/Options/FtCreateOptions.cs
```diff
@@ -0,0 +1,318 @@
...
+    /// <summary>The vector data type. Defaults to FLOAT32.</summary>
+    public VectorDataType Type { get; set; } = VectorDataType.FLOAT32;
+    /// <summary>Initial vector capacity in the index affecting memory allocation size.</summary>
+    public int? InitialCap { get; set; }
+    /// <summary>Max number of outgoing edges per node per layer (M parameter). Default 16, max 512.</summary>
+    public int? NumberOfEdges { get; set; }
+    /// <summary>Vectors examined during index construction (EF_CONSTRUCTION). Default 200, max 4096.</summary>
+    public int? VectorsExaminedOnConstruction { get; set; }
+    /// <summary>Vectors examined during query operations (EF_RUNTIME). Default 10, max 4096.</summary>
+    public int? VectorsExaminedOnRuntime { get; set; }
```
currantw: Hmmm. If these are only enforced on the Valkey level, not sure if we should mention specific values here? In my mind, we are introducing a bit of coupling that isn't needed?

sources/Valkey.Glide/Commands/Options/FtCreateOptions.cs
```diff
@@ -0,0 +1,318 @@
...
+    /// <summary>Stores term offsets. Mutually exclusive with <see cref="NoOffsets"/>.</summary>
+    public bool WithOffsets { get; set; }
+    /// <summary>Does not store term offsets. Mutually exclusive with <see cref="WithOffsets"/>.</summary>
+    public bool NoOffsets { get; set; }
+    /// <summary>Disables stop-word filtering. Mutually exclusive with <see cref="StopWords"/>.</summary>
+    public bool NoStopWords { get; set; }
+    /// <summary>Custom list of stop words. Mutually exclusive with <see cref="NoStopWords"/>.</summary>
+    public string[]? StopWords { get; set; }
+    /// <summary>Custom set of punctuation characters for tokenization.</summary>
+    public string? Punctuation { get; set; }
```
currantw: As mentioned elsewhere, think this could be nice with a builder pattern?

sources/Valkey.Glide/Commands/Options/FtInfoOptions.cs
```diff
@@ -0,0 +1,86 @@
...
+    public GlideString[] ToArgs()
+    {
+        List<GlideString> args = [];
+        if (Scope.HasValue)
+            args.Add(Scope.Value switch
+            {
+                FtInfoScope.LOCAL => ValkeyLiterals.LOCAL,
+                FtInfoScope.PRIMARY => ValkeyLiterals.PRIMARY,
+                FtInfoScope.CLUSTER => ValkeyLiterals.CLUSTER,
+                _ => Scope.Value.ToString(),
```
currantw: Three comments here:

1. Let's move all these enums to `abstract_Enums` directory.
2. It can be convenient to match the pattern in enums like `GeoAddCondition` to add a static internal class that provides a `ToArgs` method for the enum
3. We should throw `ArgumentOutOfRangeException` for unknown/unexpected values.

Can you make this change for all the enums you've declared in this PR?

sources/Valkey.Glide/Commands/Options/FtSearchOptions.cs
```diff
@@ -0,0 +1,179 @@
...
+/// Specifies a field to return in FT.SEARCH results.
+/// </summary>
+/// <seealso href="https://valkey.io/commands/ft.search/">valkey.io</seealso>
+public class FtSearchReturnField(string fieldIdentifier)
+{
+    /// <summary>The field name to return.</summary>
+    public string FieldIdentifier { get; } = fieldIdentifier;
+    /// <summary>Optional alias to override the field name in the result.</summary>
+    public string? Alias { get; set; }
+}
```
currantw: Similar to other comments. Might match elsewhere better and make it a bit easier for users?


```suggestion
/// <summary>
/// Specifies a field to return in FT.SEARCH results.
/// </summary>
/// <seealso href="https://valkey.io/commands/ft.search/">valkey.io</seealso>
public class FtSearchReturnField(string fieldIdentifier, string? alias = null)
{
    /// <summary>The field name to return.</summary>
    public string FieldIdentifier { get; init; } = fieldIdentifier;
    /// <summary>Optional alias to override the field name in the result.</summary>
    public string? Alias { get; init; } = alias
}
```

sources/Valkey.Glide/Commands/Options/FtSearchOptions.cs
```diff
@@ -0,0 +1,179 @@
...
+/// <summary>
+/// Provides pagination for FT.SEARCH results.
+/// </summary>
+/// <seealso href="https://valkey.io/commands/ft.search/">valkey.io</seealso>
+public class FtSearchLimit(int offset, int count)
+{
+    /// <summary>Number of results to skip.</summary>
+    public int Offset { get; } = offset;
+    /// <summary>Number of results to return.</summary>
+    public int Count { get; } = count;
```
currantw: Nit. Can we match the layout style that predominates elsewhere?


```suggestion
    /// <summary>
    /// Number of results to skip.
    /// </summary>
    public int Offset { get; } = offset;
    
    /// <summary>
    /// Number of results to return.
    /// </summary>
    public int Count { get; } = count;
```

currantw: This exact class (count + offset options) is repeated in a couple places! Is this something we can abstract to a shared class? It's also embedded in `RangeOptions`, but don't think we can extract that now without a breaking change 😢

CHANGELOG.md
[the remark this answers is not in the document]
edlng: ~~I'll leave an issue~~ I'll leave a note for now to make it more detailed and across all languages. No idea if I will have some time to work on it so I'll leave it open for the time being. Not a bad idea to make it more detailed for each command though

sources/Valkey.Glide/Commands/Options/FtSearchResult.cs
```diff
@@ -0,0 +1,45 @@
...
+/// <summary>
+/// Holds the parsed response from an FT.SEARCH command.
+/// Documents are returned as an ordered list that preserves the server's
+/// iteration order, which matters when SORTBY is specified.
+/// </summary>
+/// <seealso href="https://valkey.io/commands/ft.search/">valkey.io</seealso>
+public class FtSearchResult(long totalResults, List<FtSearchDocument> documents)
+{
+    /// <summary>The total number of documents matching the query.</summary>
+    public long TotalResults { get; } = totalResults;
```
currantw: Why is this needed? Does it not just match the length of `documents` below?

sources/Valkey.Glide/Commands/Options/FtSearchResult.cs
```diff
@@ -0,0 +1,45 @@
...
+    public string SortKey { get; } = sortKey;
+}
+
+/// <summary>
+/// Holds the parsed response from an FT.SEARCH command.
+/// Documents are returned as an ordered list that preserves the server's
+/// iteration order, which matters when SORTBY is specified.
+/// </summary>
+/// <seealso href="https://valkey.io/commands/ft.search/">valkey.io</seealso>
+public class FtSearchResult(long totalResults, List<FtSearchDocument> documents)
```
currantw: Since this is a results type, can the constructor be `internal`?

CHANGELOG.md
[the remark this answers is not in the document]
edlng: https://github.com/valkey-io/valkey-glide-docs/issues/197

sources/Valkey.Glide/Client/IBaseClient.VectorSearchCommands.cs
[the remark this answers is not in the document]
edlng: #353 . However, I can't assign myself because of no perms? Feel free to assign me the issue

sources/Valkey.Glide/Commands/IVectorSearchCommands.cs
```diff
@@ -0,0 +1,208 @@
...
+    /// <remarks>
+    /// <example>
+    /// <code>
+    /// await client.FtCreateAsync("my-index",
+    ///     [new TextField("title"), new NumericField("published_at"), new TagField("category")],
+    ///     new FtCreateOptions { DataType = IndexDataType.HASH, Prefixes = ["blog:post:"] });
+    /// </code>
+    /// </example>
+    /// </remarks>
+    Task FtCreateAsync(string indexName, IEnumerable<IField> schema, FtCreateOptions? options = null);
```
currantw: I the current interface, we _tend_ to split this into two separate methods with and without options:

- `FtCreateAsync(string indexName, IEnumerable<IField> schema)`
- `FtCreateAsync(string indexName, IEnumerable<IField> schema, FtCreateOptions options)`

It saves us from having to explicitly support `null` in the future 🤷 

Similar for other methods.

currantw: Can this index be a bit array? I think we typically use `ValkeyKey` here for this reason (though I'm not sure the specific implementation details for search). If that is the case, can we also make sure we add at least one integration test for this?

sources/Valkey.Glide/Commands/IVectorSearchCommands.cs
```diff
@@ -0,0 +1,208 @@
...
+    /// </example>
+    /// </remarks>
+    Task FtAliasAddAsync(string alias, string indexName);
+
+    /// <summary>
+    /// Removes an alias from an index.
+    /// </summary>
+    /// <seealso href="https://valkey.io/commands/ft.aliasdel/">valkey.io</seealso>
+    /// <param name="alias">The alias name to remove.</param>
+    /// <returns>A task that completes when the alias is removed.</returns>
```
currantw: Nit. The general docs pattern is not to include `returns` statements for blank `Task` (should be self-evident for any C# developer).

sources/Valkey.Glide/Client/IBaseClient.VectorSearchCommands.cs
```diff
@@ -0,0 +1,208 @@
...
+    /// <param name="alias">The alias name to remove.</param>
+    /// <returns>A task that completes when the alias is removed.</returns>
+    /// <remarks>
+    /// <example>
+    /// <code>
+    /// await client.FtAliasDelAsync("my-alias");
+    /// </code>
+    /// </example>
+    /// </remarks>
+    Task FtAliasDelAsync(string alias);
```
currantw: Similar to other comments, does Valkey support byte arrays as aliases – or just strings? Should we be using `ValkeyKey` here instead?

sources/Valkey.Glide/Internals/Request.VectorSearchCommands.cs
```diff
@@ -0,0 +1,170 @@
...
+            args.AddRange(field.ToArgs());
+        }
+        return Simple<string>(RequestType.FtCreate, [.. args]);
+    }
+
+    public static Cmd<string, string> FtDropIndex(string indexName)
+        => Simple<string>(RequestType.FtDropIndex, [(GlideString)indexName]);
+
+    public static Cmd<object[], ISet<string>> FtList()
+        => new(RequestType.FtList, [], false, arr => arr.Cast<GlideString>().Select(gs => gs.ToString()).ToHashSet());
```
currantw: Can we extract some sort of helper here that avoids creating a list _and_ a hash set (i.e. can we convert the array directly to a set)? If the list is very long, this could be slower. There might be some other `Request` methods that could also benefit from this helper.

sources/Valkey.Glide/Internals/Request.VectorSearchCommands.cs
```diff
@@ -0,0 +1,170 @@
...
+
+    public static Cmd<object[], FtSearchResult> FtSearch(string indexName, string query, FtSearchOptions? options)
```
currantw: Similar to other comments, I think `query` might need to be a `ValkeyValue` instead of a raw string?

sources/Valkey.Glide/Commands/Options/FtSearchOptions.cs
```diff
@@ -0,0 +1,179 @@
...
+
+    /// <summary>
+    /// Returns the command arguments for these options.
+    /// </summary>
+    public GlideString[] ToArgs()
+    {
+        if (WithSortKeys && SortBy is null)
+            throw new ArgumentException("WithSortKeys requires SortBy to be set.");
+        if (SortByOrder.HasValue && SortBy is null)
+            throw new ArgumentException("SortByOrder requires SortBy to be set.");
```
currantw: Can we alter the interface to eliminate these errors?

sources/Valkey.Glide/Commands/Options/FtSearchResult.cs
```diff
@@ -0,0 +1,45 @@
...
+public class FtSearchDocument(string key, Dictionary<string, object?> fields, string sortKey = "")
+{
+    /// <summary>The document key (e.g. the Valkey key name).</summary>
+    public string Key { get; } = key;
+
+    /// <summary>
+    /// The document's field data as key-value pairs.
+    /// Empty when NOCONTENT is used.
+    /// </summary>
+    public Dictionary<string, object?> Fields { get; } = fields;
```
currantw: For other commands, we often split a single Valkey command into multiple independent methods when the command can return multiple different return types (in this case, one of more command options are instead reflected in the different methods rather than the options):

For example:

- ` Task<ValkeyValue[]> SortedSetRangeAsync(ValkeyKey key, RangeOptions options)`
- `Task<SortedSetEntry[]> SortedSetRangeWithScoresAsync(ValkeyKey key)`

Is this something that would work with search? For example, could we add a `FtSearchNoContentAsync` method that returns a simple `ValkeyValue[]`?

currantw: It seems like `object?` is used here primarily to support JSON return types, which might be arrays or dictionaries. Is that right? We don't seem to support JSON commands yet – do we know how these will actually be returned (i.e. do we know that there are the types that will actually be used for JSON commands when we add them?)

currantw: What I'm getting at: should JSON responses be returned as strings (which is how Valkey returns them, I believe), rather than parsing them directly to more complex types?

sources/Valkey.Glide/Commands/Options/FtSearchResult.cs
```diff
@@ -0,0 +1,45 @@
...
+
+    /// <summary>
+    /// The sort key value returned when WITHSORTKEYS is used.
+    /// Empty string when WITHSORTKEYS is not requested.
+    /// </summary>
+    public string SortKey { get; } = sortKey;
```
currantw: I think this may need to be `string?` (and `null` when no sort keys requested), because the sort key itself could be equal to the empty string, right?

sources/Valkey.Glide/Client/IBaseClient.VectorSearchCommands.cs
currantw, line 1: Because it isn't part of the StackExchange.Redis interface, these methods should actually be declared in `IBaseClient.VectorSearchCommands`.

sources/Valkey.Glide/Internals/Request.VectorSearchCommands.cs
```diff
@@ -0,0 +1,170 @@
...
+            return new FtSearchResult(0, []);
+
+        long count = Convert.ToInt64(data[0]);
+        List<FtSearchDocument> docs = [];
+
+        if (data[1] is Dictionary<GlideString, object> map)
+        {
+            foreach (var kvp in map)
+            {
+                string key = kvp.Key.ToString();
```
currantw: I think this needs to parse to a `ValkeyKey`. Otherwise this will fail for a byte array key. We need to avoid using `ToString` throughout, I think?

sources/Valkey.Glide/Internals/Request.VectorSearchCommands.cs
```diff
@@ -0,0 +1,170 @@
...
+    private static FtAggregateRow[] ParseFtAggregateResponse(object[] data)
+    {
+        // The Rust core normalizes the response: strips leading count, converts rows to maps.
+        var results = new List<FtAggregateRow>();
+        foreach (var row in data)
+        {
+            if (row is Dictionary<GlideString, object> map)
+            {
+                results.Add(new FtAggregateRow(map.ToDictionary(
+                    kvp => kvp.Key.ToString(),
```
currantw: Same comment. Needs to use `ValkeyKey`/`ValkeyValue` to parse byte arrays properly

sources/Valkey.Glide/Internals/Request.VectorSearchCommands.cs
```diff
@@ -0,0 +1,170 @@
...
+
+    private static Dictionary<string, object> ParseFtInfoResponse(object data)
+    {
+        // May arrive as a Map or as a flat key/value array.
+        if (data is Dictionary<GlideString, object> map)
+        {
+            return map.ToDictionary(
+                kvp => kvp.Key.ToString(),
+                kvp => ConvertFtValue(kvp.Value)!);
+        }
```
currantw: Why is there different possible return types. That seems strange...?

sources/Valkey.Glide/BaseClient.VectorSearchCommands.cs
```diff
@@ -0,0 +1,50 @@
...
+using Valkey.Glide.Commands.Options;
+using Valkey.Glide.Internals;
+
+namespace Valkey.Glide;
+
+public abstract partial class BaseClient : IVectorSearchCommands
+{
+    /// <inheritdoc/>
+    public async Task FtCreateAsync(string indexName, IEnumerable<IField> schema, FtCreateOptions? options = null)
+        => await Command(Request.FtCreate(indexName, schema, options));
```
currantw: In cases like this, we can remove `async`/`await`. Slightly more efficient.


```suggestion
    public Task FtCreateAsync(string indexName, IEnumerable<IField> schema, FtCreateOptions? options = null)
        => Command(Request.FtCreate(indexName, schema, options));
```

tests/Valkey.Glide.UnitTests/VectorSearchCommandTests.cs
```diff
@@ -0,0 +1,327 @@
...
+using Valkey.Glide.Commands.Options;
+
+namespace Valkey.Glide.UnitTests;
+
+public class VectorSearchCommandTests
+{
+    [Fact]
+    public void ValidateFtCreateArgs()
+    {
+        Assert.Multiple(
```
currantw: Can we just have separate assertions, rather than lumping everything into one giant `Assert.Multiple`? Not sure why some otter tests do this, but just seems to make a mess with no clear benefit 🤷

CHANGELOG.md
```diff
@@ -7,3 +7,4 @@
 - Add support for Windows CI and testing with WSL (#184)
 - Add StackExchange.Redis compatible pub/sub API (#202)
 - Add transparent compression support with Zstd and LZ4 backends (#213)
+- Add Valkey Search command support: FT.CREATE, FT.DROPINDEX, FT._LIST, FT.SEARCH, FT.AGGREGATE, FT.INFO, FT.ALIASADD, FT.ALIASDEL, FT.ALIASUPDATE, FT._ALIASLIST (#225)
```
currantw: Nit. Don't think we need to include all the commands. Information easily available on the PR./ 🤷 


```suggestion
- Add Valkey Search command support (#225)
```

CHANGELOG.md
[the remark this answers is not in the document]
currantw: The issue linked above is closed with a comment that we have "Added for all languages except C#." So is there anything tracking adding this documentation for C#...?

sources/Valkey.Glide/Client/IBaseClient.VectorSearchCommands.cs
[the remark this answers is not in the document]
currantw: Thanks. Assigned to you.

CHANGELOG.md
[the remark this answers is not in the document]
edlng: Honestly think it shouldn't have been closed

sources/Valkey.Glide/Commands/Options/FtSearchResult.cs
edlng: No, `TotalResults` is the total number of documents that match the query across the entire index. `documents` only contains the results after applying limit and offset.

sources/Valkey.Glide/Client/IBaseClient.VectorSearchCommands.cs
edlng: ~~Think it's just strings~~ Actually, there's a chance that it supports it. However, I can't test it out until `valkey-search` 1.3 comes out with actual support for this command. For now, we can just leave it as a string and when 1.3 releases, I can change it and work on support for the 1.3 commands

sources/Valkey.Glide/Commands/Options/FtSearchResult.cs
edlng: Yeah we could, but IIRC, when we did the first iteration of the C# client, I thought we were kind of trying to get away from something like this because it caused a lot more methods than necessary (and we were trying to align with SER). I get that it's for consistency with the other clients as well, but IMO I feel like it's possible that doing this could degrade user experience sometimes? That being said, if it's something you guys would still like to push for now, I'd still be happy to split it.

currantw: I'm happy for you to leave it. It can (and did) create complication with SER compatibility that we needed to fix right before release when there were multiple defaults (can explain in more detail if you'd like). But given that these are GLIDE-only, I'm happy to leave as-is.

edlng: JSON responses (at least for `FT.SEARCH`) are returned as strings so I've tightened up the type here for it. `FT.AGGREGATE` doesn't do the type coercion from core so I think we can leave it for now

sources/Valkey.Glide/Commands/Options/FtSearchResult.cs
edlng: Good point!

sources/Valkey.Glide/Internals/Request.VectorSearchCommands.cs
edlng: My bad. It was leftover code when I was trying to debug the return type from core

CHANGELOG.md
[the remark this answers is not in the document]
currantw: Okay thanks for re-opening.

sources/Valkey.Glide/Internals/Request.VectorSearchCommands.cs
currantw: Thanks!

sources/Valkey.Glide/BaseClient.VectorSearchCommands.cs
currantw: 🙇

sources/Valkey.Glide/abstract_Enums/VS/DistanceMetric.cs
```diff
@@ -0,0 +1,29 @@
...
+}
+
+internal static class DistanceMetricExtensions
+{
+    internal static GlideString ToLiteral(this DistanceMetric metric)
+        => metric switch
+        {
+            DistanceMetric.Euclidean => "L2",
+            DistanceMetric.InnerProduct => "IP",
+            DistanceMetric.Cosine => "COSINE",
```
currantw, line 26: These should map to ValkeyLiterals constants if possible. Same with other enums.

sources/Valkey.Glide/Commands/Options/FtInfoOptions.cs
```diff
@@ -0,0 +1,41 @@
...
+/// </summary>
+/// <seealso href="https://valkey.io/commands/ft.info/">valkey.io</seealso>
+public sealed class FtInfoOptions
+{
+    /// <summary>Controls the scope of information returned.</summary>
+    public FtInfoScope? Scope { get; set; }
+    /// <summary>Controls which shards participate in the query.</summary>
+    public FtInfoShardScope? ShardScope { get; set; }
+    /// <summary>Controls consistency requirements.</summary>
+    public FtInfoConsistencyMode? Consistency { get; set; }
```
currantw, line 16: It seems to me like "consistent" is the default. As such, could we make this just a `bool Inconsistent` parameter, and remove `FtInfoConsistencyMode` entirely?

sources/Valkey.Glide/Commands/Options/FtInfoOptions.cs
```diff
@@ -0,0 +1,41 @@
...
+/// <summary>
+/// Optional arguments for the FT.INFO command.
+/// </summary>
+/// <seealso href="https://valkey.io/commands/ft.info/">valkey.io</seealso>
+public sealed class FtInfoOptions
+{
+    /// <summary>Controls the scope of information returned.</summary>
+    public FtInfoScope? Scope { get; set; }
+    /// <summary>Controls which shards participate in the query.</summary>
+    public FtInfoShardScope? ShardScope { get; set; }
```
currantw, line 14: Same as my comment below. Since ALLSHARDS is the default, can we just replace with a `bool SomeShards` that defaults to false, and then remove `FtInfoShardScope` enum?

sources/Valkey.Glide/abstract_Enums/VS/FtSearchConsistencyMode.cs
```diff
@@ -0,0 +1,26 @@
...
+internal static class FtSearchConsistencyModeExtensions
+{
+    internal static GlideString ToLiteral(this FtSearchConsistencyMode mode)
+        => mode switch
+        {
+            FtSearchConsistencyMode.Consistent => ValkeyLiterals.CONSISTENT,
+            FtSearchConsistencyMode.Inconsistent => ValkeyLiterals.INCONSISTENT,
+            _ => throw new ArgumentOutOfRangeException(nameof(mode)),
+        };
+}
```
currantw, line 26: Similar to other later comments, I think we can just replace this with a boolean (`bool Inconsistent = false`) and remove this enum completely.

sources/Valkey.Glide/abstract_Enums/VS/FtSearchShardScope.cs
currantw, line 1: And same here. We should be able to have a `bool SomeShards = false` parameter and use that to remove this enum completely.

sources/Valkey.Glide/abstract_Enums/VS/FtSearchSortOrder.cs
currantw, line 1: Can we use `SortOrder` instead, and remove this enum?

sources/Valkey.Glide/abstract_Enums/VS/IndexDataType.cs
```diff
@@ -0,0 +1,26 @@
...
+internal static class IndexDataTypeExtensions
+{
+    internal static GlideString ToLiteral(this IndexDataType dataType)
+        => dataType switch
+        {
+            IndexDataType.Hash => "HASH",
+            IndexDataType.Json => "JSON",
+            _ => throw new ArgumentOutOfRangeException(nameof(dataType)),
+        };
+}
```
currantw, line 26: > ON HASH | ON JSON (optional): Only keys that match the specified type are included into this index. If omitted, HASH is assumed.

Based on this, can we also replace this enum with an `bool onJson = false` property instead? Or do you think it's likely more will be added in the future?

sources/Valkey.Glide/abstract_Enums/VS/VectorAlgorithm.cs
```diff
@@ -0,0 +1,26 @@
...
+}
+
+internal static class VectorAlgorithmExtensions
+{
+    internal static GlideString ToLiteral(this VectorAlgorithm algorithm)
+        => algorithm switch
+        {
+            VectorAlgorithm.Flat => "FLAT",
+            VectorAlgorithm.Hnsw => "HNSW",
+            _ => throw new ArgumentOutOfRangeException(nameof(algorithm)),
```
currantw, line 24: Seems reasonable to me to maintain this enum (even with only two values), as I could see more vector similarity measures added in the future?

sources/Valkey.Glide/Internals/Request.cs
```diff
@@ -93,6 +93,19 @@ private static ISet<ValkeyValue> ToValkeyValueSet(HashSet<object> set)
...
+    /// </summary>
+    private static HashSet<string> ToStringSet(object[] objects)
+    {
+        HashSet<string> result = new(objects.Length);
+        foreach (var obj in objects)
+        {
+            _ = result.Add(((GlideString)obj).ToString());
+        }
+        return result;
+    }
```
currantw: I think these methods probably need to return `ISet<ValkeyValue>` instead to support `byte[]`

alexr-bq:
Is there a reason for the difference in design between this and the Java client? (this was the one that I looked at, have not gone in depth on each glide client)

The java client seems to have these modules as separate interfaces that wrap the main client.

Sorry for late comment I only clued into this as I was implementing the JSON module support.

sources/Valkey.Glide/Commands/Options/FtAggregateResult.cs
```diff
@@ -0,0 +1,24 @@
...
+/// Represents a single result row returned by <c>IBaseClient.FtAggregateAsync</c>.
+/// Each row is a flat set of field/value pairs produced by the aggregate pipeline.
+/// <para>
+/// Values are typed as <see cref="ValkeyValue"/> to preserve binary data. The glide-core
+/// layer does not coerce FT.AGGREGATE values, so the actual runtime type depends on the
+/// server and RESP protocol version.
+/// </para>
+/// </summary>
+/// <seealso href="https://valkey.io/commands/ft.aggregate/">valkey.io</seealso>
+public sealed class FtAggregateRow(IReadOnlyDictionary<string, ValkeyValue> fields)
```
currantw: This now seems like it is just a wrapper around a dictionary? Can we just get rid of this class entirely and simply use `IReadOnlyDictionary<string, ValkeyValue>`? (Or perhaps `IDictionary`)?

sources/Valkey.Glide/Commands/Options/FtCreateOptions.cs
```diff
@@ -0,0 +1,593 @@
...
+    public string? Alias { get; set; }
+
+    /// <summary>Whether the field is sortable.</summary>
+    public bool Sortable { get; set; }
+
+    /// <summary>Disables stemming when indexing the field.</summary>
+    public bool NoStem { get; set; }
+
+    /// <summary>The importance of this field when calculating result accuracy.</summary>
+    public double? Weight { get; set; }
```
currantw, line 41: Since we have `With*` methods now, we can make these all `internal set`?

sources/Valkey.Glide/Commands/Options/FtCreateOptions.cs
```diff
@@ -0,0 +1,593 @@
...
+
+    /// <summary>
+    /// Controls suffix trie behaviour for the field.
+    /// <list type="bullet">
+    ///   <item><description><see langword="true"/> — emit <c>WITHSUFFIXTRIE</c></description></item>
+    ///   <item><description><see langword="false"/> — emit <c>NOSUFFIXTRIE</c></description></item>
+    ///   <item><description><see langword="null"/> (default) — omit the option entirely</description></item>
+    /// </list>
+    /// </summary>
+    public bool? SuffixTrie { get; set; }
```
currantw, line 51: > WITHSUFFIXTRIE | NOSUFFIXTRIE (optional): Enables/Disables the use of a suffix trie to implement suffix-based wildcard queries. If NOSUFFIXTRIE is specified, query strings which specify suffix-based wildcard matching will be rejected with an error. The default is WITHSUFFIXTRIE.

Looks like this can be simplified to a simple `bool` as well!

sources/Valkey.Glide/Commands/Options/FtCreateOptions.cs
```diff
@@ -0,0 +1,593 @@
...
+    public TextField WithAlias(string alias) { Alias = alias; return this; }
+
+    /// <summary>Marks the field as sortable.</summary>
+    public TextField AsSortable() { Sortable = true; return this; }
+
+    /// <summary>Disables stemming for the field.</summary>
+    public TextField WithNoStem() { NoStem = true; return this; }
+
+    /// <summary>Sets the field weight used when calculating result accuracy.</summary>
+    public TextField WithWeight(double weight) { Weight = weight; return this; }
```
currantw, line 65: > WEIGHT <weight> (optional): The current implementation only allows the value to be 1.0. This parameter is accepted to make valkey-search more interoperable with RediSearch. (default: 1.0)

Given this, I would suggest that we should just remove this completely, since there is no interoperability concerns here.

sources/Valkey.Glide/Commands/Options/FtCreateOptions.cs
```diff
@@ -0,0 +1,593 @@
...
+
+    /// <summary>
+    /// Executes <paramref name="configure"/> only when <paramref name="condition"/> is
+    /// <see langword="true"/>. Useful for conditionally including options without breaking the chain.
+    /// </summary>
+    public FtCreateOptionsBuilder When(bool condition, Func<FtCreateOptionsBuilder, FtCreateOptionsBuilder> configure)
+        => condition ? configure(this) : this;
+
+    /// <summary>Constructs the <see cref="FtCreateOptions"/> from the accumulated state.</summary>
+    public FtCreateOptions Build() => new()
```
currantw, line 454: Rather than a totally separate builder class, I think the general pattern is just to have `With*` methods that are actually setters. So it's not _really_ a builder pattern per se, but does provide a bit more hiding and `internal set`. Make any sense?

sources/Valkey.Glide/Commands/Options/FtLimit.cs
```diff
@@ -0,0 +1,29 @@
...
+    public long Offset { get; init; } = offset;
+
+    /// <summary>
+    /// Number of results to return. <c>-1</c> returns all results.
+    /// </summary>
+    public long Count { get; init; } = count;
+
+    internal GlideString[] ToArgs() => [ValkeyLiterals.LIMIT, Offset.ToGlideString(), Count.ToGlideString()];
+
+    GlideString[] IFtAggregateClause.ToArgs() => ToArgs();
```
currantw, line 28: Don't think we need to `ToArgs` methods?

sources/Valkey.Glide/Commands/Options/FtLimit.cs
```diff
@@ -0,0 +1,29 @@
...
+/// <summary>
+/// Provides offset/count pagination for FT commands (LIMIT clause).
+/// </summary>
+/// <remarks>
+/// <paramref name="offset"/> defaults to <c>0</c> (start from the first result).
+/// <paramref name="count"/> defaults to <c>-1</c> (return all results).
+/// </remarks>
+/// <seealso href="https://valkey.io/commands/ft.search/">valkey.io</seealso>
+/// <seealso href="https://valkey.io/commands/ft.aggregate/">valkey.io</seealso>
+public sealed class FtLimit(long offset = 0, long count = -1) : IFtAggregateClause
```
currantw, line 14: Can we just object initialization/default constructor so that users don't need to specify _both_ values if they only want one?

```csharp
public sealed class FtLimit
{
    /// <summary>
    /// Number of results to skip.
    /// </summary>
    public long Offset { get; init; } = 0;

    /// <summary>
    /// Number of results to return. <c>-1</c> returns all results.
    /// </summary>
    public long Count { get; init; } = -1;
}
```

Then only need to specify what we want:

```csharp
var noLimit = new FtLimit();
var countOnly = new FtLimit { Count = 10 };
var both = new FtLimit { Offset = 20, Count = 10 };
```

sources/Valkey.Glide/Commands/Options/FtSearchOptions.cs
```diff
@@ -0,0 +1,229 @@
+// Copyright Valkey GLIDE Project Contributors - SPDX Identifier: Apache-2.0
+
+namespace Valkey.Glide.Commands.Options;
+
+/// <summary>
+/// Specifies a field to return in FT.SEARCH results.
+/// </summary>
+/// <seealso href="https://valkey.io/commands/ft.search/">valkey.io</seealso>
+public sealed class FtSearchReturnField(string fieldIdentifier, string? alias = null)
```
currantw: Does this need to support `ValkeyValue` instead of `string`?

sources/Valkey.Glide/Commands/Options/FtSearchOptions.cs
```diff
@@ -0,0 +1,229 @@
...
+    /// Optional alias to override the field name in the result.
+    /// </summary>
+    public string? Alias { get; init; } = alias;
+}
+
+/// <summary>
+/// A key/value pair passed as a query parameter for FT.SEARCH.
+/// </summary>
+/// <seealso href="https://valkey.io/commands/ft.search/">valkey.io</seealso>
+public sealed class FtSearchParam(string key, string value)
```
currantw: Does these need to use `ValkeyKey` and/or `ValkeyValue` instead of `string` – can either be byte[]?

sources/Valkey.Glide/Commands/Options/FtSearchOptions.cs
```diff
@@ -0,0 +1,229 @@
...
+    public string Value { get; init; } = value;
+}
+
+/// <summary>
+/// Specifies the SORTBY clause for FT.SEARCH, bundling the field name with
+/// optional sort direction and WITHSORTKEYS flag so that invalid combinations
+/// (e.g., <c>WithSortKeys</c> without a field) are unrepresentable.
+/// </summary>
+/// <seealso href="https://valkey.io/commands/ft.search/">valkey.io</seealso>
+public sealed class FtSearchSortBy(string fieldName, FtSearchSortOrder? order = null, bool withSortKeys = false)
```
currantw: Similar to previous comments:
- Do we need to support `ValkeyValue` instead of `string`?
- Can we use object initialization, default constructor, and default property values to make this easier to use for users (for example, if they only want to specify field name but not `WithSortKeys`.

currantw (approved):
Approving! Will be subsequent follow-up refactoring to move to separate interface.

PR #242, opened 2026-03-23
feat: Use valkey-bundle docker images for modules related test (9.0+)

### LINKED ISSUE #245: (feat): Add Modules Testing in CI
Currently, we don't have a way to test modules related commands due to valkey-server having no loaded modules. We should add a step that also tests this capability. 

One approach we can do it is to build the modules from source and have valkey-server load those modules. However, this can be very time consuming and can hold a runner for extended periods of time which is not good in busy parts of the day.

The second approach (my recommendation) is to start another valkey-server using the `valkey-bundle` Docker image, which already includes all the modules for us so that we can get to testing right away. We'll have modules-related tests run on this valkey-server, while the others run on the normal server built from source.

**Potential Question:** Why not use the Docker image for all the tests?
**Answer:** `valkey-bundle` only supports Valkey 8+ so we wouldn't be able to test 6.2. Not only that, but the CI tests seem to timeout quite often and consistently when using the Docker container, which could be because of the loaded modules tackling on a bunch of different features and slowing things down, or some other underlying issue with the image itself.

edlng:
### Summary

Add Docker-based valkey-bundle servers on dedicated ports for module tests (Search, JSON, Bloom, LDAP). Regular tests still run against the source-built server.

On the 9.0 / ubuntu matrix, CI now builds from source (like every other combo) AND pulls the `valkey/valkey-bundle` Docker image. The source-built server handles all existing tests. The Docker containers run on separate ports (standalone: 6389, cluster: 8001-8006) with modules loaded, and export `modules-standalone-endpoint` / `modules-cluster-endpoints` env vars for module-specific tests to pick up.

We tried routing everything through Docker first, but the bundle image is pinned to `9.1-rc1` (no stable 9.0 tag yet) and that caused consistent standalone timeouts. Separate ports and separate servers to handle module-related tests should be sufficient and better.

### Issue link

Closes #245

### What changed

- New composite action `.github/workflows/start-valkey-docker/action.yml`. Pulls `valkey/valkey-bundle:9.1-rc1`, starts standalone (port 6389) and cluster (ports 8001-8006) containers with `--network host`, forms the cluster, writes endpoint env vars. Uses the source-built `valkey-cli` already on the host, so no binary extraction needed.
- `install-shared-dependencies` now always builds from source first. For 9.0 on ubuntu it also runs the Docker action afterward.

### Limitations

- `9.1-rc1` pin means module tests hit a pre-release server . TODO in the action to swap once stable tags exist.
- Only 9.0 on ubuntu gets module servers, other combos skip module tests due to limitations with Docker.
- Nothing reads the `modules-*` env vars yet. #225.

### Testing

Ran the 9.0 / ubuntu matrix three times. Regular tests pass on the source-built server. Docker containers start and respond to PING. TLS tests work via `cluster_manager.py` + source-built `valkey-server`. Confirmed ports 6389 and 8001-8006 don't collide with anything in the test suite or `cluster_manager.py`'s random port range.

### Checklist

### DIFF
--- a/.github/workflows/install-shared-dependencies/action.yml
+++ b/.github/workflows/install-shared-dependencies/action.yml
@@ -78,9 +78,8 @@ runs:
           with:
               target: ${{ inputs.target }}
 
-        # Always build from source first. This provides valkey-server and
-        # valkey-cli on the host for TLS tests (cluster_manager.py), benchmarks,
-        # and — on the Docker path — the CLI needed to form the cluster.
+        # Always build from source first. This installs 'valkey-cli', which is
+        # needed by 'start-valkey-docker' to start servers in Docker.
         - name: Install server from source
           if: "${{ inputs.server-version }}"
           uses: ./.github/workflows/install-server
--- a/.github/workflows/install-shared-dependencies/action.yml
+++ b/.github/workflows/install-shared-dependencies/action.yml
@@ -78,13 +78,27 @@ runs:
           with:
               target: ${{ inputs.target }}
 
-        - name: Install server
+        # Always build from source first. This provides valkey-server and
+        # valkey-cli on the host for TLS tests (cluster_manager.py), benchmarks,
+        # and — on the Docker path — the CLI needed to form the cluster.
+        - name: Install server from source
           if: "${{ inputs.server-version }}"
           uses: ./.github/workflows/install-server
           with:
               os: ${{ inputs.os }}
               target: ${{ inputs.target }}
               server-version: ${{ inputs.server-version }}
+
+        # For valkey 9.0 on ubuntu, also start Docker containers with the
+        # valkey-bundle image (includes modules: Search, JSON, Bloom, LDAP) on
+        # dedicated ports. Regular tests use the source-built server; module
+        # tests connect to these Docker containers via modules-standalone-endpoint
+        # and modules-cluster-endpoints env vars.
+        - name: Start module servers via Docker (valkey 9.0, ubuntu)
+          if: "${{ inputs.server-version == '9.0' && inputs.os == 'ubuntu' }}"
+          uses: ./.github/workflows/start-valkey-docker
+          with:
+              engine-version: ${{ inputs.server-version }}
 
         - name: Install zig
           if: ${{ contains(inputs.target, 'linux-gnu') }}
--- a/.github/workflows/start-valkey-docker/action.yml
+++ b/.github/workflows/start-valkey-docker/action.yml
@@ -1,0 +1,135 @@
+name: Start Valkey Servers via Docker
+description: >
+    Start standalone and cluster Valkey servers with modules (Search, JSON,
+    Bloom, LDAP) using Docker containers on dedicated ports. These run
+    alongside the source-built servers (on default ports) used by regular
+    tests. Module-specific tests read the env vars written here to connect.
+    Assumes valkey-cli is already on the host (installed by install-server).
+
+inputs:
+    engine-version:
+        description: >
+            Valkey engine version (e.g. 9.0, 8.1). The Docker image is
+            pinned to 9.1-rc1 regardless - see the TODO in the image step.
+        required: true
+
+runs:
+    using: composite
+    steps:
+        - name: Determine Docker image
+          id: docker-image
+          shell: bash
+          env:
+              ENGINE_VERSION: ${{ inputs.engine-version }}
+          run: |
+              # TEMPORARY: Pin to 9.1-rc1 until stable per-version valkey-bundle tags exist.
+              # TODO: switch to valkey/valkey-bundle:${ENGINE_VERSION} once stable tags are published.
+              IMAGE="valkey/valkey-bundle:9.1-rc1"
+              echo "image=$IMAGE" >> $GITHUB_OUTPUT
+              echo "Using Docker image: $IMAGE"
+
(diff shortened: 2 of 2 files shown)


jduo (approved):
lgtm. Similar question as https://github.com/valkey-io/valkey-glide/pull/5661, should cluster creation happen via cluster_manager.py?

jduo (approved):

.github/workflows/install-shared-dependencies/action.yml
```diff
@@ -78,14 +78,28 @@ runs:
           with:
               target: ${{ inputs.target }}
 
-        - name: Install server
+        # Always build from source first. This provides valkey-server and
+        # valkey-cli on the host for TLS tests (cluster_manager.py), benchmarks,
+        # and — on the Docker path — the CLI needed to form the cluster.
+        - name: Install server from source
```
currantw: A few notes:

- `cluster_manager.py` is used for all existing integration tests
- Benchmarking has been removed (see #253)

So perhaps something like this would be a bit easier?


```suggestion
        # Always build from source first. This installs 'valkey-cli', which is
        # needed by 'start-valkey-docker' to start servers in Docker.
        - name: Install server from source
```

.github/workflows/start-valkey-docker/action.yml
```diff
@@ -0,0 +1,135 @@
...
+    using: composite
+    steps:
+        - name: Determine Docker image
+          id: docker-image
+          shell: bash
+          env:
+              ENGINE_VERSION: ${{ inputs.engine-version }}
+          run: |
+              # TEMPORARY: Pin to 9.1-rc1 until stable per-version valkey-bundle tags exist.
+              # TODO: switch to valkey/valkey-bundle:${ENGINE_VERSION} once stable tags are published.
```
currantw: Can you raise an issue for this and tag it here? e.g.


```suggestion
              # TODO #456: switch to valkey/valkey-bundle:${ENGINE_VERSION} once stable tags are published.
```

Have a future cleanup issue to remove all "naked" TODOs (i.e. ones that aren't linked to open issues).

.github/workflows/start-valkey-docker/action.yml
currantw, line 1: There seems to be few elements of this action that will only work on Linux. Probably worth a comment to make this clear to future devs?

.github/workflows/start-valkey-docker/action.yml
```diff
@@ -0,0 +1,135 @@
...
+              IMAGE: ${{ steps.docker-image.outputs.image }}
+              ENGINE_VERSION: ${{ inputs.engine-version }}
+          run: |
+              echo "=== Starting Valkey cluster with modules on ports 8001-8006 ==="
+
+              MAJOR=$(echo "$ENGINE_VERSION" | cut -d. -f1)
+              EXTRA_ARGS=""
+              if [ "$MAJOR" -ge 9 ]; then
+                EXTRA_ARGS="--cluster-databases 16"
+              fi
```
currantw: Kiro seems to think this would be better with array approach to avoid any ambiguity with expansion of unquoted variables 🤷 

```bash
EXTRA_ARGS=()
if [ "$MAJOR" -ge 9 ]; then
  EXTRA_ARGS+=(--cluster-databases 16)
fi

# later:
docker run ... "${EXTRA_ARGS[@]}"
```

.github/workflows/start-valkey-docker/action.yml
```diff
@@ -0,0 +1,135 @@
...
+                  --cluster-node-timeout 5000 \
+                  --appendonly no \
+                  --protected-mode no \
+                  --enable-debug-command yes \
+                  --save "" \
+                  $EXTRA_ARGS
+              done
+
+              # Wait for all nodes
+              for port in 8001 8002 8003 8004 8005 8006; do
```
currantw: Might be best to define the ports as environement variables to that they are easy to change in the future if we want?

.github/workflows/start-valkey-docker/action.yml
```diff
@@ -0,0 +1,135 @@
...
+              echo "Cluster with modules ready on ports 8001-8006"
+
+        - name: Verify module servers
+          shell: bash
+          run: |
+              echo "=== Module server verification ==="
+              echo "Standalone with modules (6389): $(valkey-cli -p 6389 PING)"
+              for port in 8001 8002 8003 8004 8005 8006; do
+                echo "Cluster with modules node $port: $(valkey-cli -p $port PING)"
+              done
```
currantw: Is this needed? Seems like we already have logic above to verify that the servers are working?

currantw (commented):
Reviewed to `4bbe038`.

.github/workflows/start-valkey-docker/action.yml
edlng: Agreed. Redundancy removed.

.github/workflows/start-valkey-docker/action.yml
edlng: All hail the mighty Kiro 🙇

.github/workflows/start-valkey-docker/action.yml
currantw: Thank you!

currantw (approved):
Thanks for addressing review comments!
Looks good to me. ✅

Issue #298, opened 2026-04-13
Investigate Memory Leak


Aryex:
Potential memory leak issue was identified during a stability test. This needs to be addressed before the next release. 

<img width="1800" height="352" alt="Image" src="https://github.com/user-attachments/assets/063a45dc-4497-478f-ae38-a7491a75e7bf" />

Aryex:
Countinued memory leak is still observed with the latest run, but it is a lot less than before. We will continue to investigate this issue. 

<img width="1867" height="369" alt="Image" src="https://github.com/user-attachments/assets/90afb39e-46ea-44cf-8ddb-72cf82f9e113" />

Aryex:
The most recent leak was found to be an issue with the test program itself. After the patch to the test harness memory has been stable for 17h and continuing.

<img width="1839" height="370" alt="Image" src="https://github.com/user-attachments/assets/8f734a51-0941-4e1d-8412-85416f52bb55" />

PR #300, opened 2026-04-13
fix: memory leaks for commands and pubsub

jduo:
### Summary

Fix native memory leak in free_response, fix managed memory leak from undisposed PubSub subscriptions in ConnectionMultiplexer, and clean up dead/duplicate code in the FFI layer.

### Issue Link

This pull request is linked to issue: #298 

No feature or API changes. These are internal fixes to memory management.

### Implementation

* Fix critical native memory leak in free_response. Box::leak(Box::from_raw(ptr)).free_memory() reclaimed the heap-allocated ResponseValue via Box::from_raw, but immediately leaked it again with Box::leak. While free_memory() correctly freed nested contents (strings, arrays, maps), the top-level ResponseValue Box (~24 bytes) was never
deallocated. This leaked on every single command response. Fixed by letting the Box drop naturally after free_memory() completes. free_memory(&self) only reads Copy-type fields (typ, val, size) to
reconstruct and drop inner allocations, so it's safe to drop the Box after the call returns.

* Fix managed memory leak in ConnectionMultiplexer.Dispose(). Dispose() was not calling RemoveAllSubscriptions(), leaving Subscription objects alive in the _subscriptions dictionary. Each Subscription holds delegate references (_handlers) and ChannelMessageQueue objects, which keep handler targets rooted. Added RemoveAllSubscriptions() before _db.Dispose() so subscriptions are cleaned up while the underlying client is still alive. The call is placed outside _lock since RemoveAllSubscriptions() has its own internal lock on _subscriptions.

* Remove duplicate MarshalPubSubMessage from BaseClient and unused FreeDropScriptError method declaration.

### Limitations

N/A

### Testing

Long-running test to confirm the memory leaks are solved.

### Related Issues

#298 

### Checklist

- [ x] This Pull Request is related to one issue.
- [ x] Commit message has a detailed description of what changed and why.
- [ x] Tests are added or updated and all checks pass.
- [ x] `CHANGELOG.md`, `README.md`, `DEVELOPER.md`, and other documentation files are updated.
- [ x] Destination branch is correct - `main` or release
- [ x] Create merge commit if merging release branch into `main`, squash otherwise.

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -703,7 +703,10 @@ pub unsafe extern "C-unwind" fn batch(
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn free_response(ptr: *mut ResponseValue) {
     unsafe {
-        Box::leak(Box::from_raw(ptr)).free_memory();
+        // Take ownership of ptr in the Box.
+        let boxed = Box::from_raw(ptr);
+        boxed.free_memory(); // This releases the contents of ResponseValue, but not the memory used by ResponseValue itself. This must be done explicitly.
+        // boxed is dropped here, releasing the memory used directly by ResponseValue itself.
     }
 }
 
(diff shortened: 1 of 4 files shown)


alexr-bq (approved):

currantw (approved):
Thanks @jduo!

jduo:
Verified that we see flat memory usage after 20 minutes (vs linear memory growth previously).

Issue #270, opened 2026-04-05
refactor: Command API cleanup — StackExchange.Redis compatibility and cross-client consistency


currantw:
### Summary

Parent issue for a comprehensive cleanup of the Valkey GLIDE C# client's command API surface. The goal is to ensure that commands are both compatible with [StackExchange.Redis](https://github.com/StackExchange/StackExchange.Redis)  and consistent with the other [Valkey GLIDE](https://github.com/valkey-io/valkey-glide) clients (Java, Python, Go, Node).

### Description

The Valkey GLIDE C# client has two API obligations:

1. **StackExchange.Redis (SER) compatibility** — The `Abstract/` layer (`IDatabaseAsync`, `IDatabase`, `IConnectionMultiplexer`, etc.) should closely match the StackExchange.Redis public API surface. This allows users of StackExchange.Redis to migrate to Valkey GLIDE with minimal code changes.

2. **Cross-client consistency** — The `Commands/` layer (`IStringCommands`, `IHashCommands`, etc.) and `BaseClient` partials should be consistent with the other Valkey GLIDE clients (Java, Python, Go, Node) in terms of command coverage, parameter names/order, return types, option types, and behavior.

These two surfaces should be clearly separated (see #263) so that GLIDE-native methods don't leak into the SER-compatible surface and vice versa.

This parent issue tracks the overall effort. Child issues will cover specific command groups and their associated option types, enums, and return types.

#### Scope

The cleanup is scoped to commands and their directly related types. For each command group, audit for:

- Missing commands (present in other GLIDE clients but not in C#).
- Missing overloads or variants (e.g., with options, with count, with values).
- Signature mismatches (parameter names, order, types, return types) vs. other GLIDE clients.
- Signature mismatches vs. StackExchange.Redis in the `Abstract/` layer.
- Missing or inconsistent command option types, enums, and return types.

Command groups:

- [x] String – #317 @currantw 
  - [x] #268 (SET)
  - [x] #262 (MSET, MSETNX)
- [x] Hash
  - [x] #259
  - [x] #280
- [x] List – #301 @alexr-bq 
- [x] Set – #304 @currantw 
- [x] Sorted Set – #302
- [x] Stream
  - [x] #260
  - [x] #303
  - [x] #324
- [x] Bitmap – #281
- [x] HyperLogLog – _No pre-release changes needed_. 
- [x] Geospatial – #288
- [x] Generic – #292
[...]

PR #301, opened 2026-04-13
List command refactor

alexr-bq:
### Summary

Refactors List commands to seperate out non-SER commands and ensure full SER interface compatiblity.

### Issue Link

This pull request is linked to issue: [270](https://github.com/valkey-io/valkey-glide-csharp/issues/270)

This PR improves StackExchange.Redis (SER) API compatibility for list commands:

1. **Move blocking list commands to GLIDE-only interface** - `BLMOVE`, `BLMPOP`, `BLPOP`, `BRPOP` are not supported by SER (they would block the multiplexer). These methods have been moved from `IListBaseCommands` to `IBaseClient.ListCommands.cs`.

2. **Add `ListRightPopLeftPushAsync`** - SER-compatible wrapper for the deprecated `RPOPLPUSH` command. Internally delegates to `ListMoveAsync(source, dest, ListSide.Right, ListSide.Left)`.

3. **Add `CommandFlags`-only overloads for push methods** - Added overloads for `ListLeftPushAsync` and `ListRightPushAsync` that accept `CommandFlags` without requiring `When` parameter. This fixes compilation errors for code like `ListRightPushAsync(key, values, flags)`.

4. **Remove blocking command wrappers from SER-compat interface** - Removed `ListBlockingLeftPopAsync`, `ListBlockingRightPopAsync`, `ListBlockingMoveAsync`, and `ListBlockingPopAsync` with `CommandFlags` from `IDatabaseAsync` since SER doesn't have these methods.

### Implementation

Key changes:

- `IBaseClient.ListCommands.cs` (new) - Contains GLIDE-only blocking list commands
- `IListBaseCommands.cs` - Shared interface for SER-compatible list commands only
- `IDatabaseAsync.ListCommands.cs` - SER-compat wrappers with `CommandFlags` parameter
- `Database.ListCommands.cs` - Implementation of SER-compat wrappers

The blocking commands remain fully functional via `IBaseClient` for GLIDE users who need them. They are simply not exposed through the SER-compatible `IDatabaseAsync` interface.

### Limitations

- Blocking list commands (`BLPOP`, `BRPOP`, `BLMOVE`, `BLMPOP`) are not available through the `IDatabaseAsync` interface. Users who need blocking operations should use the `IBaseClient` interface directly.
- `RPOPLPUSH` is deprecated since Redis 6.2.0. The `ListRightPopLeftPushAsync` method is provided only for SER compatibility; new code should use `ListMoveAsync`.

### Testing

- Added integration tests for `ListRightPopLeftPushAsync` covering:
  - Basic element move between lists
  - Empty source list returns null
  - Same source and destination rotates the list
- Added integration tests for `CommandFlags`-only overloads of `ListLeftPushAsync` and `ListRightPushAsync`
- Build and lint pass

### Checklist


sources/Valkey.Glide/Commands/IListBaseCommands.cs
currantw: Should this be split out into `LeftLeftPushAsync(ValkeyKey key, ValkeyValue value)` (LPUSH) and `LeftListPushIfExists(ValkeyKey key, ValkeyValue value)` (LPUSHX)? I think that this would be more compatible with the other GLIDE clients, and would also eliminate the (in my opinion) sub-par SER interface property that `When.NotExists` throws at runtime.

currantw, line 238: Since the signature for SER is

```csharp
Task<long> ListRemoveAsync(ValkeyKey key, ValkeyValue value, long count = 0, CommandFlags flags = CommandFlags.None)
```

I think we might need to move this to `IBaseClient.ListCommands`, or a call to `ListRemoveAsync(key, value [count[)` will be ambiguous?

currantw, line 261: Would it make sense to have defaults of `start = 0` and `stop = -1` like `ListRangeAsync`?

currantw, line 285: Similar to previous comment, is there any ambiguity with SER defaults that would require us to split these up?

currantw: ANy thought to naming `ListIndexAsync` for consistency with other GLIDE clients/command name?

currantw, line 349: If we are going to have `ListSide` in our interface, why aren't we using it more to simplify things in the GLIDE API? Seems like we could use this to simplify a bunch of commands:

- Combine `ListInsertBeforeAsync` and `ListInsertAfterAsync` into `ListInsertAsync(ValkeyKey key, ValkeyValue pivot, ValkeyValue value, ListSide side)`.
- Combine `ListRightPopAsync` and `ListLeftPopAsync` variations into `ListPopAsync`.
- Combine `ListRightPushAsync` and `ListLeftPushAsync` variations into `ListPushAsync`.
- etc.

We seem to already be doing this to some degree for the blocking commands, so might be good to match? 🤔

currantw, line 392: Other GLIDE clients seem to extract a `ListPositionOptions` type. Should we do the same?

sources/Valkey.Glide/Commands/IListBaseCommands.cs
```diff
@@ -452,131 +452,4 @@ public interface IListBaseCommands
     /// </example>
     /// </remarks>
     Task ListSetByIndexAsync(ValkeyKey key, long index, ValkeyValue value);
```
currantw: Do we want to extract `ListIndexAsync` for consistency with other GLIDE clients?

sources/Valkey.Glide/Commands/IListBaseCommands.cs
currantw: Do we want to extract `ListIndexAsync` for GLIDE-compatibility?

sources/Valkey.Glide/Client/IBaseClient.ListCommands.cs
```diff
@@ -0,0 +1,132 @@
...
+    /// If no element could be popped and the timeout expired, <see langword="null"/> will be returned.
+    /// </returns>
+    /// <remarks>
+    /// <example>
+    /// <code>
+    /// ValkeyValue[]? result = await client.ListBlockingLeftPopAsync(["list1", "list2"], TimeSpan.FromSeconds(5));
+    /// </code>
+    /// </example>
+    /// </remarks>
+    Task<ValkeyValue[]?> ListBlockingLeftPopAsync(IEnumerable<ValkeyKey> keys, TimeSpan timeout);
```
currantw, line 36: - Do we want to add a single key for convenience, as is typical for many other methods? Same comment for other methods below where applicable.
- Since we already have `ListBlockingPopAsync` below, do we even need to separate right and left methods?

sources/Valkey.Glide/Client/IBaseClient.ListCommands.cs
```diff
@@ -0,0 +1,132 @@
...
+    /// If no element could be popped and the timeout expired, <see cref="ListPopResult.Null"/> will be returned.
+    /// </returns>
+    /// <remarks>
+    /// <example>
+    /// <code>
+    /// ListPopResult result = await client.ListBlockingPopAsync(["list1", "list2"], ListSide.Left, TimeSpan.FromSeconds(5));
+    /// </code>
+    /// </example>
+    /// </remarks>
+    Task<ListPopResult> ListBlockingPopAsync(IEnumerable<ValkeyKey> keys, ListSide side, TimeSpan timeout);
```
currantw, line 149: - Do we want to depend on `ListPopResult`, since it is SER-only? Or would just a tuple or `KeyValuePair` be better? (`ListPopResult` is pretty minimal and allow `Null`, so could be reasonable?)
- I think this method would only every return one ValkeyValue. Is the `ListPopResult` misleading/unnecessarily general in that it can contain multiple ValkeyValue's?

sources/Valkey.Glide/Commands/IListBaseCommands.cs
alexr-bq: Added

alexr-bq: Removing default parameter to remove ambiguity.

alexr-bq: Removing default flag to resolve ambiguity.

alexr-bq: Current approach seems to match other glide clients better actually? So I'd say stick with it.

alexr-bq: I think options are minimal enough that this is fine and we could add that later if desired.

sources/Valkey.Glide/Client/IBaseClient.ListCommands.cs
alexr-bq: Added single key. 

This way seems consistent with other glide clients.

sources/Valkey.Glide/Client/IBaseClient.ListCommands.cs
alexr-bq: As you say seems reasonable to keep

The type is meant to support the other overload where multiple values can be returned. I think this is better than created a seperate return type

affonsov (approved):
I don't have any additional comment

sources/Valkey.Glide/Commands/IListBaseCommands.cs
currantw: Okay sounds good! Seems similar to what I just went through with sorted sets: SER uses an "order" (ascending or descending) for all the "pop" methods, while GLIDE uses separate `PopMin` and `PopMax` methods throughout.

(Going to save myself from digressing too much about how "left" and "right" are poor ways to refer to the "front" and "back" of a list because they depend on the culturally-derived concept that a list is naturally ordered from left to right, and GLIDE should really have gone with something like C++'s deque and used "push front/back" and "pop front/back" but anyway...)

currantw: Sure. Worth a cleanup issue to add?

sources/Valkey.Glide/Client/IBaseClient.ListCommands.cs
currantw: Sounds good. Fine with me. Ended up going with `SortedSetEntry` using similar logic for sorted sets. Did NOT do this with `HashEntry` for hash commands and re-thinking this decision. Might go back and updated with a small PR if time permits

currantw (approved):
Looks good. Thanks! ✅

PR #306, opened 2026-04-14
Scripting/Functions command refactor

alexr-bq:
### Summary

Moves scripting and function commands not supported by SER into GLIDE-specific interface.

### Issue Link

This pull request is linked to issue: https://github.com/valkey-io/valkey-glide-csharp/issues/270

This PR refactors the scripting and function command interfaces to properly separate GLIDE-specific methods from StackExchange.Redis-compatible methods:

- **FUNCTION commands moved to GLIDE-specific interfaces**: `FCallAsync`, `FCallReadOnlyAsync`, `FunctionLoadAsync`, `FunctionFlushAsync`, `FunctionListAsync`, `FunctionStatsAsync`, `FunctionDeleteAsync`, `FunctionKillAsync`, `FunctionDumpAsync`, `FunctionRestoreAsync` are now in `IBaseClient`, `IGlideClient`, and `IGlideClusterClient` partials since these commands are not available in StackExchange.Redis.

- **SCRIPT commands moved to GLIDE-specific interfaces**: `ScriptInvokeAsync`, `ScriptExistsAsync`, `ScriptFlushAsync`, `ScriptShowAsync`, `ScriptKillAsync` are now in `IBaseClient` and `IGlideClusterClient` partials since these explicit script management methods are not part of the StackExchange.Redis API.

- **SER-compatible methods retained in shared interfaces**: Only `ScriptEvaluateAsync` overloads remain in `IScriptingAndFunctionBaseCommands` as these match the StackExchange.Redis API.

- **Default-route cluster overloads added**: `FunctionDumpAsync`, `FunctionListAsync`, `FunctionKillAsync`, `FunctionRestoreAsync` now have overloads without Route parameter using sensible defaults (RandomRoute for dump, AllPrimaries for others).

### New Files
- `IBaseClient.ScriptingCommands.cs` - GLIDE-specific script and function commands for base client
- `IGlideClient.ScriptingCommands.cs` - Standalone-specific function commands (list, stats, delete, kill, dump, restore)
- `IGlideClusterClient.ScriptingCommands.cs` - Cluster-specific script and function commands with routing

### Modified Files
- `IScriptingAndFunctionBaseCommands.cs` - Now contains only SER-compatible `ScriptEvaluateAsync` methods
- `IScriptingAndFunctionStandaloneCommands.cs` - Now empty (inherits from base only)
- `IScriptingAndFunctionClusterCommands.cs` - Now empty (inherits from base only)
- `IGlideClient.cs` - Made partial to support new scripting commands file
- `IGlideClusterClient.cs` - Made partial to support new scripting commands file

### Key Design Decisions
- Interfaces in `Commands/` folder should only contain methods that exist in both GLIDE and StackExchange.Redis
- GLIDE-specific methods belong in `IBaseClient`, `IGlideClient`, or `IGlideClusterClient` partials
- This separation allows the SER-compat layer (`IDatabaseAsync`, `Database`) to only expose methods that have SER equivalents

## Limitations

- No functional changes to the underlying implementations - this is purely an interface reorganization
- The `IScriptingAndFunctionStandaloneCommands` and `IScriptingAndFunctionClusterCommands` interfaces are now effectively empty but retained for potential future SER-compatible additions

### Checklist


sources/Valkey.Glide/Commands/IScriptingAndFunctionBaseCommands.cs
```diff
@@ -327,4 +69,8 @@ Task<ValkeyResult> ScriptEvaluateAsync(
     /// a NOSCRIPT error will be thrown.
     /// </remarks>
     Task<ValkeyResult> ScriptEvaluateAsync(LoadedLuaScript script, object? parameters = null);
+
+    // Script management methods (ScriptInvokeAsync, ScriptExistsAsync, ScriptFlushAsync, ScriptShowAsync, ScriptKillAsync)
```
affonsov, line 73: I think we can move this comment as a class comment. Maybe not said that the commands are moved, but something in the line that x,y,z commands are in this class instead of this one

sources/Valkey.Glide/Commands/IScriptingAndFunctionClusterCommands.cs
```diff
@@ -6,402 +6,12 @@ namespace Valkey.Glide.Commands;
...
-    ///     Route.AllPrimaries);
-    /// </code>
-    /// </example>
-    /// </remarks>
-    Task FunctionRestoreAsync(
-        byte[] payload,
-        FunctionRestorePolicy policy,
-        Route route,
-        CancellationToken cancellationToken = default);
+    // Script management methods with routing (ScriptInvokeAsync, ScriptExistsAsync, ScriptFlushAsync, ScriptKillAsync)
```
affonsov: same here

sources/Valkey.Glide/Commands/IScriptingAndFunctionStandaloneCommands.cs
```diff
@@ -11,124 +11,6 @@ namespace Valkey.Glide.Commands;
...
-    /// <code>
-    /// await client.FunctionRestoreAsync(backup, FunctionRestorePolicy.Replace);
-    /// </code>
-    /// </example>
-    /// </remarks>
-    Task FunctionRestoreAsync(
-        byte[] payload,
-        FunctionRestorePolicy policy,
-        CancellationToken cancellationToken = default);
+    // This interface currently has no additional script methods beyond the base interface.
```
affonsov: same here

affonsov (approved):
Small nits, but LGTM

sources/Valkey.Glide/Commands/IScriptingAndFunctionBaseCommands.cs
currantw, line 27: Should we add StackExchange integration tests for these?

sources/Valkey.Glide/Commands/IScriptingAndFunctionStandaloneCommands.cs
currantw, line 1: Do we need this class at all? I understand having an empty `Base` one for completeness – but do we also want empty `Standalone` _and_ `Cluster` interfaces for all command groups?

currantw, line 1: Why are scripting and function command groups together anyway? Would it make more sense to break these into separate interfaces or are they intertwined somehow?

sources/Valkey.Glide/Client/IBaseClient.ScriptingAndFunctionCommands.cs
currantw, line 1: As per my other comment, I think it makes sense to _either_:
- Split out scripting and function commands into separate classes (sounds best to me?)
- Rename this and other classes "ScriptingAndFunction*Commands" for consistency

sources/Valkey.Glide/Client/IBaseClient.ScriptingAndFunctionCommands.cs
```diff
@@ -0,0 +1,273 @@
...
+    /// <remarks>
+    /// <example>
+    /// <code>
+    /// bool[] exists = await client.ScriptExistsAsync([script1.Hash, script2.Hash]);
+    /// </code>
+    /// </example>
+    /// </remarks>
+    Task<bool[]> ScriptExistsAsync(
+        IEnumerable<string> sha1Hashes,
+        CancellationToken cancellationToken = default);
```
currantw, line 91: Should we provide a single script method as well?


```suggestion
    Task<bool> ScriptExistsAsync(
        string sha1Hash,
        CancellationToken cancellationToken = default);
```

sources/Valkey.Glide/Client/IGlideClient.ScriptingCommands.cs
```diff
@@ -0,0 +1,135 @@
...
+    /// <param name="cancellationToken">The cancellation token.</param>
+    /// <exception cref="Errors.ValkeyServerException">Thrown if the library does not exist.</exception>
+    /// <remarks>
+    /// <example>
+    /// <code>
+    /// await client.FunctionDeleteAsync("mylib");
+    /// </code>
+    /// </example>
+    /// </remarks>
+    Task FunctionDeleteAsync(
```
currantw: Is there a reasonable routing default such that this could be moved up to the base class?

sources/Valkey.Glide/Client/IGlideClusterClient.ScriptingCommands.cs
```diff
@@ -0,0 +1,497 @@
...
+    /// <remarks>
+    /// Uses <see cref="Route.RandomRoute"/> as the default route.
+    /// <example>
+    /// <code>
+    /// byte[] backup = await client.FunctionDumpAsync();
+    /// </code>
+    /// </example>
+    /// </remarks>
+    Task<byte[]> FunctionDumpAsync(
+        CancellationToken cancellationToken = default);
```
currantw: Looks like this exists for both standalone and cluster. Can we move it to base?

sources/Valkey.Glide/Client/IBaseClient.ScriptingCommands.cs
```diff
@@ -0,0 +1,273 @@
...
+    /// bool[] exists = await client.ScriptExistsAsync([script1.Hash, script2.Hash]);
+    /// </code>
+    /// </example>
+    /// </remarks>
+    Task<bool[]> ScriptExistsAsync(
+        IEnumerable<string> sha1Hashes,
+        CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Flushes all scripts from the server cache using default flush mode (SYNC).
```
currantw: Is it true that sync is the default?

From [FUNCTION FLUSH](https://valkey.io/commands/function-flush/):

> Unless called with the optional mode argument, the `lazyfree-lazy-user-flush` configuration directive sets the effective behavior. Valid modes are:
> 
> - `ASYNC`: Asynchronously flush the libraries.
> - `!SYNC`: Synchronously flush the libraries.

sources/Valkey.Glide/Client/IBaseClient.ScriptingAndFunctionCommands.cs
```diff
@@ -0,0 +1,273 @@
...
+    /// <param name="cancellationToken">The cancellation token.</param>
+    /// <remarks>
+    /// <example>
+    /// <code>
+    /// await client.FunctionFlushAsync();
+    /// </code>
+    /// </example>
+    /// </remarks>
+    Task FunctionFlushAsync(
+        CancellationToken cancellationToken = default);
```
currantw, line 277: Same comment about default behaviour that I also (perhaps erroneously) made for scripts

sources/Valkey.Glide/Client/IGlideClient.ScriptingCommands.cs
```diff
@@ -0,0 +1,135 @@
...
+    /// <exception cref="Errors.ValkeyServerException">Thrown if no function is running or if the function has written data.</exception>
+    /// <remarks>
+    /// <example>
+    /// <code>
+    /// await client.FunctionKillAsync();
+    /// </code>
+    /// </example>
+    /// </remarks>
+    Task FunctionKillAsync(
+        CancellationToken cancellationToken = default);
```
currantw: The SCRIPT KILL command is part of the base client, but not the FUNCTION KILL one. Is there a reason for this?

sources/Valkey.Glide/Client/IGlideClient.ScriptingCommands.cs
```diff
@@ -0,0 +1,135 @@
...
+    /// <remarks>
+    /// <example>
+    /// <code>
+    /// LibraryInfo[] libraries = await client.FunctionListAsync();
+    /// </code>
+    /// </example>
+    /// </remarks>
+    Task<LibraryInfo[]> FunctionListAsync(
+        FunctionListQuery? query = null,
+        CancellationToken cancellationToken = default);
```
currantw: Some thoughts on `FunctionListQuery`:

- I think it should be renamed to `FunctionListOptions` for consistency with other GLIDE clients.
- Other `Options` don't tend to have fluent builder methods, and so can be `readonly struct` with `default` values instead. I think that is a bit simpler type _and_ interface. You could always have static constructors like `ForLibrary(string name)` if we thought that was useful, but if both properties are options then it is pretty easy to build. 🤷 

While I am looking at this, we should probably also make the constructors for `LibraryInfo` and `FunctionInfo` internal, and get rid of all the extra `?? throw new ArgumentNullException(nameof(name))` checks? I don't think we want users to somehow depend on being able to create result types, and certainly not on the fact that those result types include what should be unnecessary null checks 😬 

Same comment for  these, unfortunately:
- `RunningScriptInfo`
- `FunctionStatsResult`

sources/Valkey.Glide/Client/IGlideClusterClient.ScriptingCommands.cs
```diff
@@ -0,0 +1,497 @@
...
+    ///     replace: true,
+    ///     Route.AllPrimaries);
+    /// </code>
+    /// </example>
+    /// </remarks>
+    Task<ClusterValue<string>> FunctionLoadAsync(
+        string libraryCode,
+        bool replace,
+        Route route,
+        CancellationToken cancellationToken = default);
```
currantw, line 257: Should there also be a version with default replace?

sources/Valkey.Glide/Client/IGlideClient.ScriptingCommands.cs
```diff
@@ -0,0 +1,135 @@
...
+    /// <example>
+    /// <code>
+    /// await client.FunctionRestoreAsync(backup, FunctionRestorePolicy.Replace);
+    /// </code>
+    /// </example>
+    /// </remarks>
+    Task FunctionRestoreAsync(
+        byte[] payload,
+        FunctionRestorePolicy policy,
+        CancellationToken cancellationToken = default);
```
currantw: Can these be moved to `IBaseClient`?

sources/Valkey.Glide/Client/IGlideClient.ScriptingCommands.cs
```diff
@@ -0,0 +1,135 @@
...
+    /// <returns>Function statistics including engine stats and running script information.</returns>
+    /// <remarks>
+    /// <example>
+    /// <code>
+    /// FunctionStatsResult stats = await client.FunctionStatsAsync();
+    /// </code>
+    /// </example>
+    /// </remarks>
+    Task<FunctionStatsResult> FunctionStatsAsync(
+        CancellationToken cancellationToken = default);
```
currantw: Is there a reasonable default for clusters so we can move to base?

currantw (commented):
Some comments. More generally, can you check whether some more of these methods can be moved to `IBaseClient` with reasonable route defaults for cluster?

sources/Valkey.Glide/Commands/IScriptingAndFunctionBaseCommands.cs
alexr-bq: Adding

sources/Valkey.Glide/Commands/IScriptingAndFunctionStandaloneCommands.cs
alexr-bq: Consistent with other clients to group together.

sources/Valkey.Glide/Client/IGlideClient.ScriptingCommands.cs
alexr-bq: This is something that could be added later with a non-breaking change so deferring for now.

sources/Valkey.Glide/Client/IBaseClient.ScriptingAndFunctionCommands.cs
alexr-bq: Sorry I'm not clear on what this is referring to.

sources/Valkey.Glide/Client/IGlideClient.ScriptingCommands.cs
currantw: Good call. Resolving.

sources/Valkey.Glide/Client/IBaseClient.ScriptingAndFunctionCommands.cs
currantw: Ahhh. Can we clarify whether the default behaviour here is actual SYNC, or is it based on server configuration (i.e. is "using default flush mode (SYNC)." part accurate>?)

currantw (approved):
Looks good ✅

PR #310, opened 2026-04-14
perf: Reduce per-request GC allocation overhead

jduo:
### Summary

Reduce per-operation GC allocation overhead in the hot path for command execution. Eliminates unnecessary Task, delegate, and async state machine
allocations on every GET/SET.

### Issue Link

This pull request is linked to issue: n/a

n/a

### Implementation

1. Replace Task.Run with ThreadPool.UnsafeQueueUserWorkItem in FFI callbacks (BaseClient.cs)

SuccessCallback and FailureCallback used Task.Run to offload work from the Rust Tokio thread. Task.Run allocates a Task object (~96 bytes) + closure (~32 bytes) per call. ThreadPool.UnsafeQueueUserWorkItem achieves the same offloading without creating a Task object.

2. Cache static converter delegates (Request.cs, Request.String.cs)

OKToBool, Ok, and StringGet factory methods allocated a new Func<> delegate on every call despite the lambdas being statically identical. Cached as static readonly fields: IsOkConverter, ToOkConverter, ToValkeyValueConverter. Saves ~32 bytes per GET/SET operation.

3. Elide async/await in passthrough methods (Database.StringCommands.cs)

The StackExchange.Redis compatibility layer had async methods that only validated flags then delegated to the inner method. Each async method allocates a state machine even when it just passes through. Changed to return the Task directly, eliminating one state machine allocation (~100 bytes) per call through this layer.

### Benchmark Results

Tested with [resp-bench](https://github.com/Aryex/resp-bench) — 4 connections, 80/20 GET/SET, 256-byte values, 2-minute steady state per run.

Throughput (req/s):

| Pipeline Depth | Before | After | Change |
|---|---|---|---|
| 1 | 21,823 | 22,216 | +1.8% |
| 10 | 94,708 | 94,656 | ~0% |
| 100 | 172,832 | 176,883 | +2.3% |

Memory (3-minute steady state, pipeline=1):

| Metric | Before | After |
|---|---|---|
| RSS | 80 MB (stable) | 80 MB (stable) |
| GC Heap | 1,278 KB | 1,240 KB |

### Limitations
n/a

### Testing
No additional tests for internal changes.

### Related Issues
n/a

### Checklist


sources/Valkey.Glide/Abstract/Database.GenericCommands.cs
```diff
@@ -231,16 +231,16 @@ public async Task<ValkeyKey> KeyRandomAsync(CommandFlags flags = CommandFlags.No
     }
 
     /// <inheritdoc cref="IDatabaseAsync.SortAsync(ValkeyKey, long, long, Order, SortType, ValkeyValue, IEnumerable{ValkeyValue}?, CommandFlags)"/>
-    public async Task<ValkeyValue[]> SortAsync(ValkeyKey key, long skip = 0, long take = -1, Order order = Order.Ascending, SortType sortType = SortType.Numeric, ValkeyValue by = default, IEnumerable<ValkeyValue>? get = null, CommandFlags flags = CommandFlags.None)
+    public Task<ValkeyValue[]> SortAsync(ValkeyKey key, long skip = 0, long take = -1, Order order = Order.Ascending, SortType sortType = SortType.Numeric, ValkeyValue by = default, IEnumerable<ValkeyValue>? get = null, CommandFlags flags = CommandFlags.None)
     {
         GuardClauses.ThrowIfCommandFlags(flags);
-        return await ((IGenericBaseCommands)this).SortAsync(key, skip, take, order, sortType, by, get);
+        return ((IGenericBaseCommands)this).SortAsync(key, skip, take, order, sortType, by, get);
```
currantw, line 237: @alexr-bq I've seen this before. Usually means that we made some sort of interface mistake with SER-compatibility and defaults. Can you take a look and fix if needed?

sources/Valkey.Glide/Internals/Request.cs
```diff
@@ -28,14 +28,17 @@ private static Cmd<T, T> Simple<T>(RequestType request, GlideString[] args, bool
     private static Cmd<T, bool> Boolean<T>(RequestType request, GlideString[] args)
         => new(request, args, false, response => Convert.ToInt64(response) == 1);
 
+    private static readonly Func<string, bool> IsOkConverter = response => response == "OK";
+    private static readonly Func<string, ValkeyValue> ToOkConverter = _ => ValkeyValue.Ok;
```
currantw, line 32: Don't think we actually need to use this anymore:
https://github.com/valkey-io/valkey-glide-csharp/pull/261

 I'll fix it in one of my branches.

currantw (approved):
Thanks James!

alexr-bq (approved):

sources/Valkey.Glide/Abstract/Database.GenericCommands.cs
alexr-bq: Fixed here https://github.com/valkey-io/valkey-glide-csharp/pull/301

PR #316, opened 2026-04-15
Scan Command Refactor

alexr-bq:
### Summary

Makes Scan interfaces consistent, separating SER from glide interfaces

### Issue Link

This pull request is linked to issue: [287](https://github.com/valkey-io/valkey-glide-csharp/issues/287)

- **Unified `IAsyncEnumerable` pattern for all SCAN commands**: Changed `ScanAsync` from cursor-based `Task<(string cursor, ValkeyKey[] keys)>` to `IAsyncEnumerable<ValkeyKey>` for consistency with `SetScanAsync` and `SortedSetScanAsync`.

- **Interface separation (GLIDE vs SER)**: Moved SCAN methods to appropriate interfaces based on their API style:
  - `ScanAsync` → `IGlideClient` / `IGlideClusterClient` (GLIDE-only, not in SER)
  - `SetScanAsync` → `IBaseClient.SetCommands.cs` (GLIDE) / `IDatabaseAsync.SetCommands.cs` (SER)
  - `SortedSetScanAsync` → `IBaseClient.SortedSetCommands.cs` (GLIDE) / `IDatabaseAsync.SortedSetCommands.cs` (SER)

- **Batch API alignment**: Fixed `IBatchSetCommands.SetScan` to match `SortedSetScan` pattern (parameter order: `pattern` before `cursor`, `int pageSize` instead of `long count`, default `pageSize = 250`).

- **Internal type visibility**: Made `ClusterScanCursor` internal since it's an implementation detail.

### Implementation

Key changes:
- Removed `ScanAsync` from shared `IGenericCommands` / `IGenericClusterCommands` interfaces
- Added `ScanAsync` to client-specific `IGlideClient` and `IGlideClusterClient` interfaces
- Created new partial interface files `IBaseClient.SetCommands.cs` and `IBaseClient.SortedSetCommands.cs` for GLIDE-specific scan methods
- SER-compat versions in `IDatabaseAsync.*.cs` include `CommandFlags flags = CommandFlags.None` parameter
- Updated `GlideClient` and `GlideClusterClient` to implement `IAsyncEnumerable`-based scan

### Limitations

- `CommandFlags` parameter in SER-compat scan methods throws `NotImplementedException` for non-`None` values (consistent with other SER-compat methods)
- Breaking changes to `ScanAsync` signature and `ClusterScanCursor` visibility (acceptable for pre-1.0 library)

### Testing

- Added new SER-compat integration tests for `SetScanAsync` and `SortedSetScanAsync` covering:
  - Empty set handling
  - All members retrieval
  - Pattern filtering
  - Page size behavior
  - Page offset (client-side skip)
  - Score preservation (sorted sets)
  - CommandFlags validation

### Checklist


sources/Valkey.Glide/Client/IGlideClient.cs
```diff
@@ -55,4 +55,23 @@ public interface IGlideClient :
...
+    /// <example>
+    /// <code>
+    /// await foreach (var key in client.ScanAsync("user:*"))
+    /// {
+    ///     Console.WriteLine(key);
+    /// }
+    /// </code>
+    /// </example>
+    /// <seealso href="https://valkey.io/commands/scan/">SCAN command</seealso>
+    IAsyncEnumerable<ValkeyKey> ScanAsync(ValkeyValue pattern = default, int pageSize = 250, long cursor = 0, int pageOffset = 0);
```
currantw: This doesn't really make sense to me?

SCAN is a cursor based iterator. This means that at every call of the command, the server returns an updated cursor that the user needs to use as the cursor argument in the next call.
An iteration starts when the cursor is set to 0, and terminates when the cursor returned by the server is 0. The following is an example of SCAN iteration.

Based on my reading, it doesn't seem like you COULD use any cursor other than zero here? They don't seem to otherwise correspond to anything?
pageOffset also isn't a concept for the Valkey server. Is this something that we need to support in our API? Or would it be pretty easy to a user to implement this (i.e. just skip the first N values) themselves?

We also do not seem to support the TYPE parameter that Valkey uses. Not a big deal, so long as we can add in the future.

Based on all this, can we just use ScanOptions here? Same for all the other scan commands?

sources/Valkey.Glide/Client/IGlideClusterClient.cs
```diff
@@ -139,4 +139,22 @@ public interface IGlideClusterClient :
...
+    /// <code>
+    /// await foreach (var key in client.ScanAsync("user:*"))
+    /// {
+    ///     Console.WriteLine(key);
+    /// }
+    /// </code>
+    /// </example>
+    /// <seealso href="https://valkey.io/commands/scan/">SCAN command</seealso>
+    /// <seealso href="https://glide.valkey.io/how-to/scan-cluster/">Valkey GLIDE – Scan a Cluster</seealso>
+    IAsyncEnumerable<ValkeyKey> ScanAsync(ValkeyValue pattern = default, int pageSize = 250);
```
currantw: We don't support `Type` here it seems. I guess we could add that in the future?

sources/Valkey.Glide/Client/GlideClient.cs
```diff
@@ -140,8 +140,41 @@ public async Task<string> LolwutAsync()
     }
 
     /// <inheritdoc/>
-    public async Task<(string cursor, ValkeyKey[] keys)> ScanAsync(string cursor, ScanOptions? options = null)
-        => await Command(Request.ScanAsync(cursor, options));
+    public async IAsyncEnumerable<ValkeyKey> ScanAsync(ValkeyValue pattern = default, int pageSize = 250, long cursor = 0, int pageOffset = 0)
+    {
+        string currentCursor = cursor.ToString();
```
currantw: Doesn't seem like any case where we wouldn't start from cursor zero?

sources/Valkey.Glide/Client/GlideClient.cs
```diff
@@ -140,8 +140,41 @@ public async Task<string> LolwutAsync()
...
+
+            IEnumerable<ValkeyKey> keysToYield = currentOffset > 0 ? keys.Skip(currentOffset) : keys;
+
+            foreach (ValkeyKey key in keysToYield)
+            {
+                yield return key;
+            }
+
+            currentCursor = nextCursor;
+            currentOffset = 0; // Only skip on first iteration
```
currantw: As mentioned elswhere, seems like something that could either be incorporated into `ScanOptions` (if we want to handle it), or that a user could pretty easily implement themselves?

sources/Valkey.Glide/Client/IGlideClient.cs
alexr-bq: Updated all scan interfaces to use ScanOptions, I had made the interfaces consistent in one way but I agree upon reflection that the other way is better. 

Agreed on pageOffset not being needed for glide interface.

sources/Valkey.Glide/Client/IGlideClient.cs
```diff
@@ -54,4 +55,28 @@ public partial interface IGlideClient :
...
+    /// // Scan with pattern and type filter
+    /// var options = new ScanOptions { MatchPattern = "user:*", Type = ValkeyType.String };
+    /// await foreach (var key in client.ScanAsync(options))
+    /// {
+    ///     Console.WriteLine(key);
+    /// }
+    /// </code>
+    /// </example>
+    /// <seealso href="https://valkey.io/commands/scan/">SCAN command</seealso>
+    IAsyncEnumerable<ValkeyKey> ScanAsync(ScanOptions? options = null);
```
currantw, line 81: Seems identical to the cluster method? Can it be moved up? (No need now if it won't break things)

currantw (approved):
One small comment, otherwise LGTM

yipin-chen (approved):

affonsov (approved):

PR #330, opened 2026-04-17
feat(core): add ttl based client-side cache configuration and metrics API

### LINKED ISSUE #342: feat(core): add ttl based client-side cache configuration and metrics API
### Describe the feature

Add client-side caching support to the C# GLIDE client. This includes configuration options for TTL-based local caching, eviction policies (LRU/LFU), and cache metrics APIs (hit/miss rates, entry counts, evictions, expirations, total lookups).

### Use Case

Client-side caching reduces round-trips to the server for frequently accessed keys, improving latency and throughput for read-heavy workloads. This is especially valuable for hot keys accessed via GET, HGETALL, and SMEMBERS.

### Proposed Solution

- Add a `ClientSideCacheConfig` class with a fluent builder for configuring TTL and eviction policy.
- Wire configuration through FFI to glide-core's caching layer.
- Expose cache metric methods on `BaseClient` (`GetCacheHitRateAsync`, `GetCacheMissRateAsync`, `GetCacheEntryCountAsync`, `GetCacheEvictionsAsync`, `GetCacheExpirationsAsync`, `GetCacheTotalLookupsAsync`).
- Add `WithClientSideCache()` builder methods to both standalone and cluster client configuration builders.

### Acknowledgements

- [ ] I may be able to implement this feature request – assigned to @affonsov 
- [ ] This feature might cause a breaking change – new functionality

### Related

:white_circle: None

### Additional Information

:white_circle: None

affonsov:
### Summary

Add client-side caching support to the C# GLIDE client. This includes configuration, FFI, cache metric commands, and integration tests.

### Issue Link

Closes #342

- New `ClientSideCacheConfig` class for configuring TTL-based local caching of read commands (GET, HGETALL, SMEMBERS)
- New `EvictionPolicy` enum (LRU, LFU)
- `WithClientSideCache()` builder method on both `StandaloneClientConfigurationBuilder` and `ClusterClientConfigurationBuilder`
- Cache metric methods on `BaseClient`:
  - `GetCacheHitRateAsync()` / `GetCacheMissRateAsync()` — rates (0.0–1.0)
  - `GetCacheEntryCountAsync()` — current entry count
  - `GetCacheEvictionsAsync()` / `GetCacheExpirationsAsync()` / `GetCacheTotalLookupsAsync()` — counters

### Implementation

- **C# config**: `ClientSideCacheConfig.cs` with fluent builder, auto-generated unique cache IDs, and `ToFfi()` marshalling
- **C# FFI**: Added `ClientSideCacheConfig` struct and `CacheMetricsType` enum to `FFI.structs.cs`, `GetCacheMetricsFfi` to `FFI.methods.cs`
- **C# client**: `BaseClient.ClientSideCacheCommands.cs` implements metrics via message container + FFI callback pattern
- **Rust FFI** (`rust/src/ffi.rs`): Added `ClientSideCacheConfig` and `EvictionPolicy` structs, wired into `ConnectionConfig` and `create_connection_request()`
- **Rust lib** (`rust/src/lib.rs`): Added `get_cache_metrics` FFI function that calls glide-core's `cache_hit_rate()`, `cache_miss_rate()`, etc.
- **ConnectionConfiguration**: Added `ClientSideCacheConfig?` field to internal `ConnectionConfig` record and `WithClientSideCache()` to the builder

### Limitations

- Only TTL-based caching is supported (no server-driven invalidation via `CLIENT TRACKING`)
- Lazy eviction only — expired entries are removed on access, not proactively
- Only GET, HGETALL, and SMEMBERS responses are cached (determined by glide-core)
- No StackExchange.Redis equivalent — this is a GLIDE-specific feature

### Testing

- new integration tests:
  - Basic cache hit/miss with metrics verification
  - Metrics disabled — rate calls throw, entry count still works
  - No cache configured — all metric calls throw
  - Multiple keys with hit rate validation
  - TTL expiration with expiration counter verification
  - LRU eviction policy behavior
  - LFU eviction policy behavior
  - Config validation (positive values, unique IDs, defaults, fluent chaining)

### Related Issues

⚪ None

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,10 @@
 # Changelog
 
-### 1.0.0
-
-- Add client-side caching support with TTL-based expiration, LRU/LFU eviction policies, and cache metrics (#330)
+## 1.0.0
 
 ### Changes
+
+- Add client-side caching support with TTL-based expiration, LRU/LFU eviction policies, and cache metrics (#330)
 
 ## 0.10.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 1.0.0
+## 1.1.0
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+### 1.0.0
+
+- Add client-side caching support with TTL-based expiration, LRU/LFU eviction policies, and cache metrics (#330)
+
+### Changes
+
 ## 0.10.0
 
 ### Changes
--- a/rust/deny.toml
+++ b/rust/deny.toml
@@ -53,8 +53,8 @@ allow = [
     "Apache-2.0 WITH LLVM-exception",
     "BSD-2-Clause",
     "BSD-3-Clause",
-    "Unicode-3.0",
     "ISC",
+    "Unicode-3.0",
     "Zlib",
 ]
 # The confidence threshold for detecting a license from license text.
--- a/rust/deny.toml
+++ b/rust/deny.toml
@@ -53,8 +53,8 @@ allow = [
     "Apache-2.0 WITH LLVM-exception",
     "BSD-2-Clause",
     "BSD-3-Clause",
-    "ISC",
     "Unicode-3.0",
+    "ISC",
     "Zlib",
 ]
 # The confidence threshold for detecting a license from license text.
--- a/sources/Valkey.Glide/Client/BaseClient.ClientSideCacheCommands.cs
+++ b/sources/Valkey.Glide/Client/BaseClient.ClientSideCacheCommands.cs
@@ -26,25 +26,25 @@ public abstract partial class BaseClient
 
     /// <inheritdoc/>
     public Task<double> GetCacheHitRateAsync()
-        => GetCacheMetricAsync(CacheMetricsType.HitRate, v => Convert.ToDouble(v));
+        => GetCacheMetricAsync(CacheMetricsType.HitRate, Convert.ToDouble);
 
     /// <inheritdoc/>
     public Task<double> GetCacheMissRateAsync()
-        => GetCacheMetricAsync(CacheMetricsType.MissRate, v => Convert.ToDouble(v));
+        => GetCacheMetricAsync(CacheMetricsType.MissRate, Convert.ToDouble);
 
     /// <inheritdoc/>
     public Task<long> GetCacheEntryCountAsync()
-        => GetCacheMetricAsync(CacheMetricsType.EntryCount, v => Convert.ToInt64(v));
+        => GetCacheMetricAsync(CacheMetricsType.EntryCount, Convert.ToInt64);
 
     /// <inheritdoc/>
     public Task<long> GetCacheEvictionsAsync()
-        => GetCacheMetricAsync(CacheMetricsType.Evictions, v => Convert.ToInt64(v));
+        => GetCacheMetricAsync(CacheMetricsType.Evictions, Convert.ToInt64);
 
     /// <inheritdoc/>
     public Task<long> GetCacheExpirationsAsync()
-        => GetCacheMetricAsync(CacheMetricsType.Expirations, v => Convert.ToInt64(v));
+        => GetCacheMetricAsync(CacheMetricsType.Expirations, Convert.ToInt64);
 
     /// <inheritdoc/>
     public Task<long> GetCacheTotalLookupsAsync()
-        => GetCacheMetricAsync(CacheMetricsType.TotalLookups, v => Convert.ToInt64(v));
+        => GetCacheMetricAsync(CacheMetricsType.TotalLookups, Convert.ToInt64);
 }
--- a/sources/Valkey.Glide/Client/BaseClient.ClientSideCacheCommands.cs
+++ b/sources/Valkey.Glide/Client/BaseClient.ClientSideCacheCommands.cs
@@ -1,0 +1,113 @@
+// Copyright Valkey GLIDE Project Contributors - SPDX Identifier: Apache-2.0
+
+using Valkey.Glide.Internals;
+
+using static Valkey.Glide.Internals.FFI;
+using static Valkey.Glide.Internals.ResponseHandler;
+
+namespace Valkey.Glide;
+
+public abstract partial class BaseClient
+{
+    /// <inheritdoc/>
+    public async Task<double> GetCacheHitRateAsync()
+    {
+        Message message = MessageContainer.GetMessageForCall();
+        GetCacheMetricsFfi(ClientPointer, (ulong)message.Index, (uint)CacheMetricsType.HitRate);
+        IntPtr response = await message;
+        try
+        {
+            object? value = HandleResponse(response);
+            return value is double d ? d : Convert.ToDouble(value);
+        }
+        finally
+        {
+            FreeResponse(response);
+        }
+    }
+
+    /// <inheritdoc/>
+    public async Task<double> GetCacheMissRateAsync()
+    {
+        Message message = MessageContainer.GetMessageForCall();
+        GetCacheMetricsFfi(ClientPointer, (ulong)message.Index, (uint)CacheMetricsType.MissRate);
+        IntPtr response = await message;
+        try
+        {
+            object? value = HandleResponse(response);
+            return value is double d ? d : Convert.ToDouble(value);
+        }
+        finally
+        {
+            FreeResponse(response);
+        }
+    }
+
+    /// <inheritdoc/>
+    public async Task<long> GetCacheEntryCountAsync()
+    {
+        Message message = MessageContainer.GetMessageForCall();
+        GetCacheMetricsFfi(ClientPointer, (ulong)message.Index, (uint)CacheMetricsType.EntryCount);
+        IntPtr response = await message;
+        try
+        {
+            object? value = HandleResponse(response);
+            return value is long l ? l : Convert.ToInt64(value);
+        }
+        finally
+        {
+            FreeResponse(response);
+        }
+    }
+
+    /// <inheritdoc/>
+    public async Task<long> GetCacheEvictionsAsync()
+    {
+        Message message = MessageContainer.GetMessageForCall();
+        GetCacheMetricsFfi(ClientPointer, (ulong)message.Index, (uint)CacheMetricsType.Evictions);
+        IntPtr response = await message;
+        try
+        {
+            object? value = HandleResponse(response);
+            return value is long l ? l : Convert.ToInt64(value);
+        }
+        finally
+        {
+            FreeResponse(response);
+        }
+    }
+
+    /// <inheritdoc/>
+    public async Task<long> GetCacheExpirationsAsync()
+    {
+        Message message = MessageContainer.GetMessageForCall();
+        GetCacheMetricsFfi(ClientPointer, (ulong)message.Index, (uint)CacheMetricsType.Expirations);
+        IntPtr response = await message;
+        try
+        {
+            object? value = HandleResponse(response);
+            return value is long l ? l : Convert.ToInt64(value);
+        }
+        finally
+        {
+            FreeResponse(response);
+        }
+    }
+
+    /// <inheritdoc/>
+    public async Task<long> GetCacheTotalLookupsAsync()
+    {
+        Message message = MessageContainer.GetMessageForCall();
+        GetCacheMetricsFfi(ClientPointer, (ulong)message.Index, (uint)CacheMetricsType.TotalLookups);
+        IntPtr response = await message;
+        try
+        {
+            object? value = HandleResponse(response);
+            return value is long l ? l : Convert.ToInt64(value);
+        }
+        finally
+        {
+            FreeResponse(response);
+        }
+    }
+}
--- a/sources/Valkey.Glide/Client/BaseClient.ClientSideCacheCommands.cs
+++ b/sources/Valkey.Glide/Client/BaseClient.ClientSideCacheCommands.cs
@@ -9,16 +9,14 @@ namespace Valkey.Glide;
 
 public abstract partial class BaseClient
 {
-    /// <inheritdoc/>
-    public async Task<double> GetCacheHitRateAsync()
+    private async Task<T> GetCacheMetricAsync<T>(CacheMetricsType type, Func<object?, T> convert)
     {
         Message message = MessageContainer.GetMessageForCall();
-        GetCacheMetricsFfi(ClientPointer, (ulong)message.Index, (uint)CacheMetricsType.HitRate);
+        GetCacheMetricsFfi(ClientPointer, (ulong)message.Index, (uint)type);
         IntPtr response = await message;
         try
         {
-            object? value = HandleResponse(response);
-            return value is double d ? d : Convert.ToDouble(value);
+            return convert(HandleResponse(response));
         }
         finally
         {
@@ -27,87 +25,26 @@ public abstract partial class BaseClient
     }
 
     /// <inheritdoc/>
-    public async Task<double> GetCacheMissRateAsync()
-    {
-        Message message = MessageContainer.GetMessageForCall();
-        GetCacheMetricsFfi(ClientPointer, (ulong)message.Index, (uint)CacheMetricsType.MissRate);
-        IntPtr response = await message;
-        try
-        {
-            object? value = HandleResponse(response);
-            return value is double d ? d : Convert.ToDouble(value);
-        }
-        finally
-        {
-            FreeResponse(response);
-        }
-    }
+    public Task<double> GetCacheHitRateAsync()
+        => GetCacheMetricAsync(CacheMetricsType.HitRate, v => Convert.ToDouble(v));
 
     /// <inheritdoc/>
-    public async Task<long> GetCacheEntryCountAsync()
-    {
-        Message message = MessageContainer.GetMessageForCall();
-        GetCacheMetricsFfi(ClientPointer, (ulong)message.Index, (uint)CacheMetricsType.EntryCount);
-        IntPtr response = await message;
-        try
-        {
-            object? value = HandleResponse(response);
-            return value is long l ? l : Convert.ToInt64(value);
-        }
-        finally
-        {
-            FreeResponse(response);
-        }
-    }
+    public Task<double> GetCacheMissRateAsync()
+        => GetCacheMetricAsync(CacheMetricsType.MissRate, v => Convert.ToDouble(v));
 
     /// <inheritdoc/>
-    public async Task<long> GetCacheEvictionsAsync()
-    {
-        Message message = MessageContainer.GetMessageForCall();
-        GetCacheMetricsFfi(ClientPointer, (ulong)message.Index, (uint)CacheMetricsType.Evictions);
-        IntPtr response = await message;
-        try
-        {
-            object? value = HandleResponse(response);
-            return value is long l ? l : Convert.ToInt64(value);
-        }
-        finally
-        {
-            FreeResponse(response);
-        }
-    }
+    public Task<long> GetCacheEntryCountAsync()
+        => GetCacheMetricAsync(CacheMetricsType.EntryCount, v => Convert.ToInt64(v));
 
     /// <inheritdoc/>
-    public async Task<long> GetCacheExpirationsAsync()
-    {
-        Message message = MessageContainer.GetMessageForCall();
-        GetCacheMetricsFfi(ClientPointer, (ulong)message.Index, (uint)CacheMetricsType.Expirations);
-        IntPtr response = await message;
-        try
-        {
-            object? value = HandleResponse(response);
-            return value is long l ? l : Convert.ToInt64(value);
-        }
-        finally
-        {
-            FreeResponse(response);
-        }
-    }
+    public Task<long> GetCacheEvictionsAsync()
+        => GetCacheMetricAsync(CacheMetricsType.Evictions, v => Convert.ToInt64(v));
 
     /// <inheritdoc/>
-    public async Task<long> GetCacheTotalLookupsAsync()
-    {
-        Message message = MessageContainer.GetMessageForCall();
-        GetCacheMetricsFfi(ClientPointer, (ulong)message.Index, (uint)CacheMetricsType.TotalLookups);
-        IntPtr response = await message;
-        try
-        {
-            object? value = HandleResponse(response);
-            return value is long l ? l : Convert.ToInt64(value);
-        }
-        finally
-        {
-            FreeResponse(response);
-        }
-    }
+    public Task<long> GetCacheExpirationsAsync()
+        => GetCacheMetricAsync(CacheMetricsType.Expirations, v => Convert.ToInt64(v));
+
+    /// <inheritdoc/>
+    public Task<long> GetCacheTotalLookupsAsync()
+        => GetCacheMetricAsync(CacheMetricsType.TotalLookups, v => Convert.ToInt64(v));
 }
--- a/sources/Valkey.Glide/Client/IBaseClient.ClientSideCacheCommands.cs
+++ b/sources/Valkey.Glide/Client/IBaseClient.ClientSideCacheCommands.cs
@@ -1,0 +1,110 @@
+// Copyright Valkey GLIDE Project Contributors - SPDX Identifier: Apache-2.0
+
+namespace Valkey.Glide;
+
+/// ATTENTION: Methods should only be added to this interface if they are implemented
+/// by Valkey GLIDE clients but NOT by StackExchange.Redis databases. Methods implemented
+/// by both should be added to the corresponding Commands interface instead.
+
+/// <summary>
+/// Client-side cache commands for Valkey GLIDE clients.
+/// These methods provide access to cache metrics when client-side caching is enabled.
+/// </summary>
+/// <seealso cref="ClientSideCacheConfig"/>
+public partial interface IBaseClient
+{
+    /// <summary>
+    /// Returns the cache hit rate as a value between 0.0 and 1.0.
+    /// A higher hit rate indicates better cache performance.
+    /// </summary>
+    /// <returns>The cache hit rate (0.0 to 1.0).</returns>
+    /// <exception cref="Errors.RequestException">
+    /// Thrown when client-side caching is not enabled or metrics collection is disabled.
+    /// </exception>
+    /// <example>
+    /// <code>
+    /// double hitRate = await client.GetCacheHitRateAsync();
+    /// Console.WriteLine($"Cache hit rate: {hitRate:P2}");
+    /// </code>
+    /// </example>
+    Task<double> GetCacheHitRateAsync();
+
+    /// <summary>
+    /// Returns the cache miss rate as a value between 0.0 and 1.0.
+    /// A lower miss rate indicates better cache performance.
+    /// </summary>
+    /// <returns>The cache miss rate (0.0 to 1.0).</returns>
+    /// <exception cref="Errors.RequestException">
+    /// Thrown when client-side caching is not enabled or metrics collection is disabled.
+    /// </exception>
+    /// <example>
+    /// <code>
+    /// double missRate = await client.GetCacheMissRateAsync();
+    /// Console.WriteLine($"Cache miss rate: {missRate:P2}");
+    /// </code>
+    /// </example>
+    Task<double> GetCacheMissRateAsync();
+
+    /// <summary>
+    /// Returns the current number of entries in the client-side cache.
+    /// </summary>
+    /// <returns>The number of cache entries.</returns>
+    /// <exception cref="Errors.RequestException">
+    /// Thrown when client-side caching is not enabled.
+    /// </exception>
+    /// <example>
+    /// <code>
+    /// long entryCount = await client.GetCacheEntryCountAsync();
+    /// Console.WriteLine($"Cache contains {entryCount} entries");
+    /// </code>
+    /// </example>
+    Task<long> GetCacheEntryCountAsync();
+
+    /// <summary>
+    /// Returns the total number of cache evictions.
+    /// Evictions occur when the cache reaches its maximum size and entries are removed
+    /// based on the configured <see cref="EvictionPolicy"/>.
+    /// </summary>
+    /// <returns>The number of cache evictions.</returns>
+    /// <exception cref="Errors.RequestException">
+    /// Thrown when client-side caching is not enabled or metrics collection is disabled.
+    /// </exception>
+    /// <example>
+    /// <code>
+    /// long evictions = await client.GetCacheEvictionsAsync();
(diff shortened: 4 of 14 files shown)


sources/Valkey.Glide/Internals/FFI.structs.cs
```diff
@@ -1185,6 +1192,31 @@ internal struct CompressionConfig
         public bool Enabled;
     }
 
+    [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Ansi)]
+    internal struct ClientSideCacheConfig
```
currantw: - We can make this `readonly`.
- Might be easier to use a primary constructor here. See `OpenTelemetryConfig` as an example.

sources/Valkey.Glide/Internals/FFI.structs.cs
```diff
@@ -1185,6 +1192,31 @@ internal struct CompressionConfig
...
+    {
+        /// <summary>Unique identifier for the cache instance.</summary>
+        [MarshalAs(UnmanagedType.LPStr)]
+        public string CacheId;
+
+        /// <summary>Maximum size of the cache in kilobytes.</summary>
+        public ulong MaxCacheKb;
+
+        /// <summary>Time-To-Live for cached entries in milliseconds (0 = no expiration).</summary>
+        public ulong EntryTtlMs;
```
currantw: This needs to exactly match the memory layout of the corresponding Rust struct. So we need a `HasEntryTtlMs` boolean _and_ an `EntryTtlMs` long.

```csharp
/// <summary>
/// Time-To-Live for cached entries in
/// </summary>
[MarshalAs(UnmanagedType.U1)]
public bool HasEntryTtlMs;
public ulong EntryTtlMs;
```

Nit. I think it is probably fine to have just one comment for both these parameters. It's an internal class that is only used for marshalling, if you want to make it a bit more concise and maybe a bit easier to read. 🤷

sources/Valkey.Glide/Client/IBaseClient.ClientSideCacheCommands.cs
```diff
@@ -0,0 +1,110 @@
+// Copyright Valkey GLIDE Project Contributors - SPDX Identifier: Apache-2.0
+
+namespace Valkey.Glide;
+
+/// ATTENTION: Methods should only be added to this interface if they are implemented
+/// by Valkey GLIDE clients but NOT by StackExchange.Redis databases. Methods implemented
+/// by both should be added to the corresponding Commands interface instead.
```
currantw: I don't think "Methods implemented by both should be added to the corresponding Commands interface instead." is relevant in this case. 🤷 


```suggestion
/// ATTENTION: Methods should only be added to this interface if they are implemented
/// by Valkey GLIDE clients but NOT by StackExchange.Redis databases.
```

sources/Valkey.Glide/Client/IBaseClient.ClientSideCacheCommands.cs
```diff
@@ -0,0 +1,110 @@
...
+    Task<double> GetCacheHitRateAsync();
+
+    /// <summary>
+    /// Returns the cache miss rate as a value between 0.0 and 1.0.
+    /// A lower miss rate indicates better cache performance.
+    /// </summary>
+    /// <returns>The cache miss rate (0.0 to 1.0).</returns>
+    /// <exception cref="Errors.RequestException">
+    /// Thrown when client-side caching is not enabled or metrics collection is disabled.
+    /// </exception>
```
currantw, line 38: I think `ConfigurationError` ("error that is thrown when a request cannot be completed in current configuration settings.") might be more appropriate here? Same for other methods below.

sources/Valkey.Glide/Client/IBaseClient.ClientSideCacheCommands.cs
```diff
@@ -0,0 +1,110 @@
...
+    /// <example>
+    /// <code>
+    /// long totalLookups = await client.GetCacheTotalLookupsAsync();
+    /// Console.WriteLine($"Cache has performed {totalLookups} lookups");
+    /// </code>
+    /// </example>
+    Task<long> GetCacheTotalLookupsAsync();
```
currantw, line 108: For consistency with the name of the previous two methods (and since all three represent "totals"), would it make sense to either (a) update this method so it doesn't include "Total" (i.e. `GetCacheLookupsAsync`), or (b) update the methods above so they _do_ include total (i.e. `GetCacheTotalEvictionsAsync` and `GetCacheTotalExpirationsAsync`).

sources/Valkey.Glide/ClientSideCacheConfig.cs
```diff
@@ -0,0 +1,161 @@
...
+///     This means cached values may become stale if updated on the server before the TTL expires.
+///   </item>
+///   <item>
+///     Currently, Glide's client-side cache supports lazy eviction only. Expired entries
+///     are removed only when accessed after their TTL has expired. There is no proactive
+///     background cleanup of expired entries.
+///   </item>
+///   <item>
+///     Currently, only read commands that retrieve entire values are cached (GET, HGETALL, SMEMBERS).
+///   </item>
```
currantw, line 29: Nits:
- While we certainly aren't perfect, I think Valkey GLIDE is more common in existing documentation.
- Might be easier for for users if we link directly to the corresponding methods, rather than the names of the Valkey commands.
- Are there any command to support non-lazy eviction or caching for other commands? If not, I would just get rid of currently for brevity 🤷 


```suggestion
///   <item>
///     Valkey GLIDE supports TTL-based caching only. Invalidation-based client-side
///     caching (where the server notifies clients of key changes) is not supported.
///     This means cached values may become stale if updated on the server before the TTL expires.
///   </item>
///   <item>
///     Valkey GLIDE client-side cache supports lazy eviction only. Expired entries
///     are removed only when accessed after their TTL has expired. There is no proactive
///     background cleanup of expired entries.
///   </item>
///   <item>
///     Currently, only read commands that retrieve entire values are cached
///     (<see cref="IBaseClient.GetAsync(ValkeyKey)">GetAsync</see>,
///     <see cref="IBaseClient.HashGetAsync(ValkeyKey)">HashGetAsync</see>,
///     <see cref="IBaseClient.SetMembersAsync(ValkeyKey)">SetMembersAsync</see>).
///   </item>
```

sources/Valkey.Glide/ClientSideCacheConfig.cs
```diff
@@ -0,0 +1,161 @@
...
+    /// When this limit is reached, entries are evicted based on the <see cref="EvictionPolicy"/>.
+    /// </summary>
+    public ulong MaxCacheKb { get; }
+
+    /// <summary>
+    /// The Time-To-Live for cached entries in milliseconds.
+    /// After this duration, entries automatically expire and are removed from the cache.
+    /// A value of <c>0</c> means no expiration is applied.
+    /// </summary>
+    public ulong EntryTtlMs { get; private set; }
```
currantw: I think we should use `TimeSpan` here instead for the public API, consistent with https://github.com/valkey-io/valkey-glide-csharp/issues/270#issuecomment-4199693530. We can wait until as late as possible to convert it into ms.

sources/Valkey.Glide/ClientSideCacheConfig.cs
```diff
@@ -0,0 +1,161 @@
...
+
+    /// <summary>
+    /// The policy for evicting entries when the cache reaches its maximum size.
+    /// If <see langword="null"/>, the default policy of <see cref="Valkey.Glide.EvictionPolicy.LRU"/> will be used.
+    /// </summary>
+    public EvictionPolicy? EvictionPolicy { get; private set; }
```
currantw, line 67: Would it be simpler for user to make this field required with a default value of LRU? Or, if the default is coming from GLIDE core, should we even mention it here?


```suggestion
    /// <summary>
    /// The policy for evicting entries when the cache reaches its maximum size.
    /// </summary>
    public EvictionPolicy EvictionPolicy { get; private set; } = EvictionPolicy.LRU
```

sources/Valkey.Glide/ClientSideCacheConfig.cs
```diff
@@ -0,0 +1,161 @@
...
+    /// </summary>
+    internal Internals.FFI.ClientSideCacheConfig ToFfi() => new()
+    {
+        CacheId = CacheId,
+        MaxCacheKb = MaxCacheKb,
+        EntryTtlMs = EntryTtlMs,
+        HasEvictionPolicy = EvictionPolicy.HasValue,
+        EvictionPolicy = EvictionPolicy ?? default,
+        EnableMetrics = EnableMetrics,
+    };
```
currantw: As mentioned in another comment, using a primary constructor `Internals.FFI.ClientSideCacheConfig` could also simplify this quite a bit 🤷 


```suggestion
    internal Internals.FFI.ClientSideCacheConfig ToFfi() => new(CacheId, MaxCacheKb, EntryTtlMs, EvictionPolicy, EnableMetrics);
```

sources/Valkey.Glide/ClientSideCacheConfig.cs
```diff
@@ -0,0 +1,161 @@
...
+    /// <summary>
+    /// The Time-To-Live for cached entries in milliseconds.
+    /// After this duration, entries automatically expire and are removed from the cache.
+    /// A value of <c>0</c> means no expiration is applied.
+    /// </summary>
+    public ulong EntryTtlMs { get; private set; }
+
+    /// <summary>
+    /// The policy for evicting entries when the cache reaches its maximum size.
+    /// If <see langword="null"/>, the default policy of <see cref="Valkey.Glide.EvictionPolicy.LRU"/> will be used.
```
currantw: Looks like a lint error here?


```suggestion
    /// If <see langword="null"/>, the default policy of <see cref="EvictionPolicy.LRU"/> will be used.
```

sources/Valkey.Glide/ClientSideCacheConfig.cs
```diff
@@ -0,0 +1,161 @@
...
+///   <item>Clients using different databases cannot share the same cache.</item>
+///   <item>Clients using different ACL users cannot share the same cache.</item>
+/// </list>
+/// </remarks>
+/// <seealso cref="EvictionPolicy"/>
+public sealed class ClientSideCacheConfig
+{
+    private static readonly object CounterLock = new();
+    private static readonly string UuidPrefix = Guid.NewGuid().ToString("N")[..8];
+    private static ulong s_counter;
```
currantw: Probably worth a comment explaining how these work together to ensure unique cache IDs?

sources/Valkey.Glide/ClientSideCacheConfig.cs
```diff
@@ -0,0 +1,161 @@
...
+    public ClientSideCacheConfig WithEntryTtlMs(ulong ttlMs)
+    {
+        if (ttlMs == 0)
+        {
+            throw new ArgumentOutOfRangeException(nameof(ttlMs), "ttlMs must be positive.");
+        }
+
+        EntryTtlMs = ttlMs;
+        return this;
+    }
```
currantw: Is mentioned in another comment, I think we should use `TimeSpan` here instead of milliseconds. Would it also make sense to have `TimeSpan.Zero` indicate "no expiry"? That is the pattern that is already used through the codebase when `TimeSpan` is used for command timeouts.

tests/Valkey.Glide.IntegrationTests/ClientSideCacheTests.cs
```diff
@@ -0,0 +1,397 @@
...
+        value = await client.GetAsync("cache_test_key");
+        Assert.Equal("cache_test_value", value.ToString());
+
+        // Third GET — cache hit
+        value = await client.GetAsync("cache_test_key");
+        Assert.Equal("cache_test_value", value.ToString());
+
+        // Verify metrics: 1 miss + 2 hits = 3 total
+        double hitRate = await client.GetCacheHitRateAsync();
+        Assert.InRange(hitRate, 2.0 / 3.0 - 0.001, 2.0 / 3.0 + 0.001);
```
currantw: Can simplify this by specifying a precision with `Assert.Equal`:


```suggestion
        Assert.Equal(hitRate, 2.0 / 3.0, precision: 3);
```

tests/Valkey.Glide.IntegrationTests/ClientSideCacheTests.cs
```diff
@@ -0,0 +1,397 @@
...
+
+        // Rates should sum to 1.0
+        Assert.InRange(hitRate + missRate, 0.9999, 1.0001);
+    }
+
+    [Fact]
+    public async Task BasicCacheHitWithMetrics_Cluster()
+    {
+        var cache = new ClientSideCacheConfig(1024)
+            .WithEntryTtlMs(60000)
```
currantw: Is this TTL necessary? Seems like the default (no expiry) would be sufficient?

tests/Valkey.Glide.IntegrationTests/ClientSideCacheTests.cs
```diff
@@ -0,0 +1,397 @@
...
+
+        double missRate = await client.GetCacheMissRateAsync();
+        Assert.InRange(missRate, 1.0 / 3.0 - 0.001, 1.0 / 3.0 + 0.001);
+
+        // Rates should sum to 1.0
+        Assert.InRange(hitRate + missRate, 0.9999, 1.0001);
+    }
+
+    [Fact]
+    public async Task BasicCacheHitWithMetrics_Cluster()
```
currantw: Can we parametrize this method `TestUtils.Data.ClusterMode` and handle cluster and standalone once to reduce duplication? Same for other methods.

tests/Valkey.Glide.IntegrationTests/ClientSideCacheTests.cs
```diff
@@ -0,0 +1,397 @@
...
+
+        Assert.InRange(hitRate + missRate, 0.9999, 1.0001);
+    }
+
+    #endregion
+
+    #region Without Metrics
+
+    [Fact]
+    public async Task WithoutMetrics_MetricCallsFail_Standalone()
```
currantw: Think we should also test this for cluster. Can be parametrized, as per previous comment.

tests/Valkey.Glide.IntegrationTests/ClientSideCacheTests.cs
```diff
@@ -0,0 +1,397 @@
...
+        double oldHitRate = await client.GetCacheHitRateAsync();
+        retrieved = await client.GetAsync("key1");
+        Assert.Equal(largeValue, retrieved.ToString());
+        double newHitRate = await client.GetCacheHitRateAsync();
+        Assert.True(newHitRate > oldHitRate);
+    }
+
+    #endregion
+
+    #region Config Validation
```
currantw: These should be part of the unit tests, I think, rather than integration tests.

sources/Valkey.Glide/ClientSideCacheConfig.cs
```diff
@@ -0,0 +1,161 @@
...
+{
+    private static readonly object CounterLock = new();
+    private static readonly string UuidPrefix = Guid.NewGuid().ToString("N")[..8];
+    private static ulong s_counter;
+
+    /// <summary>
+    /// A unique identifier for the cache instance.
+    /// Multiple clients can share the same cache by using the same <see cref="ClientSideCacheConfig"/> instance.
+    /// </summary>
+    public string CacheId { get; }
```
currantw: I don't think there is any value in exposing this to the users? You can make it `internal` if you want to test uniqueness (add a comment that it is internal only for testing).

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+### 1.0.0
+
+- Add client-side caching support with TTL-based expiration, LRU/LFU eviction policies, and cache metrics (#330)
+
+### Changes
```
currantw: For consistency with below:


```suggestion
## 1.0.0

### Changes

- Add client-side caching support with TTL-based expiration, LRU/LFU eviction policies, and cache metrics (#330)
```

sources/Valkey.Glide/Client/BaseClient.ClientSideCacheCommands.cs
```diff
@@ -0,0 +1,113 @@
...
+        IntPtr response = await message;
+        try
+        {
+            object? value = HandleResponse(response);
+            return value is double d ? d : Convert.ToDouble(value);
+        }
+        finally
+        {
+            FreeResponse(response);
+        }
```
currantw, line 24: You can add a private helper method like this to reduce duplication?

```
private async Task<T> GetCacheMetricAsync<T>(CacheMetricsType type, Func<object?, T> convert)
{
    Message message = MessageContainer.GetMessageForCall();
    GetCacheMetricsFfi(ClientPointer, (ulong)message.Index, (uint)type);
    IntPtr response = await message;
    try
    {
        return convert(HandleResponse(response));
    }
    finally
    {
        FreeResponse(response);
    }
}
```

Then this could be simplified:


```suggestion
        return GetCacheMetricAsync(CacheMetricsType.HitRate, v => Convert.ToDouble(v));
```

tests/Valkey.Glide.IntegrationTests/ClientSideCacheTests.cs
```diff
@@ -0,0 +1,397 @@
...
+    public async Task BasicCacheHitWithMetrics_Standalone()
+    {
+        var cache = new ClientSideCacheConfig(1024)
+            .WithEntryTtlMs(60000)
+            .WithMetrics(true);
+
+        await using var client = await CreateStandaloneClientWithCache(cache);
+
+        // SET a key
+        await client.SetAsync("cache_test_key", "cache_test_value");
```
currantw: Because test servers are shared between many tests, we need to use unique keys for each one. Check out the GUID pattern in other integ. tests.

tests/Valkey.Glide.IntegrationTests/ClientSideCacheTests.cs
```diff
@@ -0,0 +1,397 @@
...
+
+        // Second GET — from cache
+        value = await client.GetAsync("key");
+        Assert.Equal("value", value.ToString());
+
+        // Metrics that require EnableMetrics should fail
+        _ = await Assert.ThrowsAsync<Errors.RequestException>(() => client.GetCacheHitRateAsync());
+        _ = await Assert.ThrowsAsync<Errors.RequestException>(() => client.GetCacheMissRateAsync());
+        _ = await Assert.ThrowsAsync<Errors.RequestException>(() => client.GetCacheEvictionsAsync());
+        _ = await Assert.ThrowsAsync<Errors.RequestException>(() => client.GetCacheExpirationsAsync());
```
currantw, line 118: Looks like we are missing `GetCacheTotalLookupsAsync`? We also seem to be missing any tests for this method in the configured/valid state?

currantw (commented):
Reviewed to `44f1149`.

sources/Valkey.Glide/Internals/FFI.structs.cs
affonsov: updated

sources/Valkey.Glide/Internals/FFI.structs.cs
affonsov: I updated this from the last core update, TTL is now mandatory so I removed the hasEntryTtlMS and update all the code accordingly

sources/Valkey.Glide/Client/IBaseClient.ClientSideCacheCommands.cs
affonsov: Done

sources/Valkey.Glide/Client/IBaseClient.ClientSideCacheCommands.cs
affonsov: This comes from the rust API and its all the same for all the other clients

sources/Valkey.Glide/ClientSideCacheConfig.cs
affonsov: updated

sources/Valkey.Glide/ClientSideCacheConfig.cs
affonsov: updated

sources/Valkey.Glide/ClientSideCacheConfig.cs
affonsov: updated, but this will not match how the other bindings are using this API. What are you thoughts?

sources/Valkey.Glide/ClientSideCacheConfig.cs
affonsov: updated

sources/Valkey.Glide/ClientSideCacheConfig.cs
affonsov: I simplified this, so now is easier to understand. The cacheId is just a guid

sources/Valkey.Glide/ClientSideCacheConfig.cs
affonsov: Updated

tests/Valkey.Glide.IntegrationTests/ClientSideCacheTests.cs
affonsov: the ttl changed to be mandatory from last core/binding update

tests/Valkey.Glide.IntegrationTests/ClientSideCacheTests.cs
affonsov: Done

sources/Valkey.Glide/ClientSideCacheConfig.cs
affonsov: Yes, with last core and other binding update this should not be exposed.

CHANGELOG.md
affonsov: Updated

sources/Valkey.Glide/Client/BaseClient.ClientSideCacheCommands.cs
affonsov: Updated

tests/Valkey.Glide.IntegrationTests/ClientSideCacheTests.cs
affonsov: added GetCacheTotalLookupsAsync

tests/Valkey.Glide.IntegrationTests/ClientSideCacheTests.cs
affonsov: Updated

tests/Valkey.Glide.IntegrationTests/ClientSideCacheTests.cs
affonsov: Updated

tests/Valkey.Glide.IntegrationTests/ClientSideCacheTests.cs
affonsov: Updated

tests/Valkey.Glide.IntegrationTests/ClientSideCacheTests.cs
affonsov: Updated

sources/Valkey.Glide/ClientSideCacheConfig.cs
affonsov: Updated

sources/Valkey.Glide/Client/IBaseClient.ClientSideCacheCommands.cs
affonsov: ConfigurationError sounds like for configuration such client connection configuration. This is going through the request api

rust/deny.toml
```diff
@@ -53,8 +53,8 @@ allow = [
     "Apache-2.0 WITH LLVM-exception",
     "BSD-2-Clause",
     "BSD-3-Clause",
-    "ISC",
     "Unicode-3.0",
+    "ISC",
     "Zlib",
 ]
```
currantw, line 59: Why this diff? List seems to otherwise be alphabetical?

sources/Valkey.Glide/Client/BaseClient.ClientSideCacheCommands.cs
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    /// <inheritdoc/>
+    public Task<double> GetCacheHitRateAsync()
+        => GetCacheMetricAsync(CacheMetricsType.HitRate, v => Convert.ToDouble(v));
```
currantw: Nit. Can these be simplified?


```suggestion
        => GetCacheMetricAsync(CacheMetricsType.HitRate, Convert.ToDouble);
```

sources/Valkey.Glide/Client/IBaseClient.ClientSideCacheCommands.cs
currantw: Sure 🤷

sources/Valkey.Glide/ClientSideCacheConfig.cs
```diff
@@ -0,0 +1,138 @@
...
+    ///     .WithEvictionPolicy(EvictionPolicy.LRU)
+    ///     .WithMetrics(true);
+    ///
+    /// var config = new ConnectionConfiguration.StandaloneClientConfigurationBuilder()
+    ///     .WithAddress("localhost", 6379)
+    ///     .WithClientSideCache(cache)
+    ///     .Build();
+    /// </code>
+    /// </example>
+    public ClientSideCacheConfig(ulong maxCacheKb, TimeSpan entryTtl)
```
currantw, line 93: Does it make sense to treat `EntryTtl` like the other two "optional" parameters? Have it default to `TimeSpan.Zero`, and then provide a `WithEntryTtl` method? Or do other clients always require this to be explicitly provided?

sources/Valkey.Glide/Client/IBaseClient.ClientSideCacheCommands.cs
currantw: Alright i guess 😬

tests/Valkey.Glide.IntegrationTests/ClientSideCacheTests.cs
```diff
@@ -0,0 +1,343 @@
...
+    {
+        if (clusterMode)
+        {
+            return await CreateClusterClientWithCache(cache);
+        }
+
+        return await CreateStandaloneClientWithCache(cache);
+    }
+
+    public static TheoryData<bool> ClusterModes => [false, true];
```
currantw: There should already be a constant for this that you can use in `TestUtils.Data`.

sources/Valkey.Glide/ClientSideCacheConfig.cs
```diff
@@ -0,0 +1,138 @@
...
+    {
+        if (maxCacheKb == 0)
+        {
+            throw new ArgumentOutOfRangeException(nameof(maxCacheKb), "maxCacheKb must be positive.");
+        }
+
+        if (entryTtl < TimeSpan.Zero)
+        {
+            throw new ArgumentOutOfRangeException(nameof(entryTtl), "entryTtl must not be negative.");
+        }
```
currantw: I believe that there is an existing `GuardClause` method that checks for negative time spans.

sources/Valkey.Glide/ClientSideCacheConfig.cs
```diff
@@ -0,0 +1,138 @@
...
+        return this;
+    }
+
+    /// <summary>
+    /// Converts to the FFI representation for marshalling to Rust core.
+    /// </summary>
+    internal Internals.FFI.ClientSideCacheConfig ToFfi() => new(
+        CacheId,
+        MaxCacheKb,
+        (ulong)EntryTtl.TotalMilliseconds,
```
currantw: To avoid casting a `double` to `ulong`, I think we want to use `timeSpan.Ticks / TimeSpan.TicksPerMillisecond`. Definitely seems like we need to extract a helper for this!

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 1.0.0
+
+### Changes
+
+- Add client-side caching support with TTL-based expiration, LRU/LFU eviction policies, and cache metrics (#330)
```
currantw: This is for 1.1.0, right?

alexr-bq (approved):

sources/Valkey.Glide/ClientSideCacheConfig.cs
currantw: Hmm. I think you're right. Let's make it optional, and let Rust core handle the default. Sorry about the rework 🤦

currantw (commented):
Thanks for changes! Looks good. Just a few minor comments.

rust/deny.toml
affonsov: I had a merge conflict here XD, fixed

sources/Valkey.Glide/Client/BaseClient.ClientSideCacheCommands.cs
affonsov: done

tests/Valkey.Glide.IntegrationTests/ClientSideCacheTests.cs
affonsov: refactored

sources/Valkey.Glide/ClientSideCacheConfig.cs
affonsov: Refactored

sources/Valkey.Glide/ClientSideCacheConfig.cs
affonsov: Created a helper in the Utils for this. Please take a look

CHANGELOG.md
affonsov: Yes, this file is very outdated. Do you have a issue to update that ?

sources/Valkey.Glide/ClientSideCacheConfig.cs
affonsov: No worries, refactored the code back

currantw: Thanks!

sources/Valkey.Glide/ClientSideCacheConfig.cs
currantw: Looks good. Thanks!

CHANGELOG.md
currantw: Nope. Thanks for updating.

sources/Valkey.Glide/ClientSideCacheConfig.cs
affonsov: This is following the same pattern as other clients, which entryTTL is required

currantw: Alright 🤷‍♂️

currantw (approved):
Looks good! ✅

PR #348, opened 2026-04-23
Update Compression Support in C#

alexr-bq:
### Summary

This PR brings in latest core changes for compression, and adds tests for supported commands and blocking of unsupported commands.

### Issue Link

This pull request is linked to issue: https://github.com/valkey-io/valkey-glide/issues/5299

This PR adds proper compression support for the C# FFI layer, enabling:

- **CustomCommand compression**: Commands sent via `CustomCommand()` now properly compress/decompress values when compression is enabled. Previously, CustomCommand calls bypassed compression processing.
- **Incompatible command detection**: Commands that are incompatible with compression (APPEND, GETRANGE, SETRANGE, STRLEN, INCR, INCRBY, INCRBYFLOAT, DECR, DECRBY, GETBIT, SETBIT, BITCOUNT, BITPOS, BITFIELD, BITFIELD_RO, BITOP) now return proper error messages explaining why they cannot be used with compression enabled.
- **Improved error messages**: Fixed an issue where handled errors were incorrectly reported as "Native function panicked" instead of the actual error message.

### Implementation

Key changes in `rust/src/lib.rs`:

1. **`resolve_custom_command_type` function**: Resolves the actual `RequestType` from the first argument of a CustomCommand. Uses the centralized `RequestType::from_command_name()` method from glide-core to avoid code duplication.

2. **Updated `command` function**:
   - Resolves the request type for CustomCommand before processing compression
   - Passes the resolved type to both `compress_cmd` and `process_response_for_decompression`
   - Fixed panic guard handling to set `panicked = false` before early returns

3. **Panic guard fix**: The `PanicGuard` is designed to catch actual Rust panics. Previously, when returning early due to handled errors (compression failure, command creation failure), the guard would incorrectly report "Native function panicked". Now we properly signal that these are controlled exits.

Key changes in `rust/src/ffi.rs`:

4. **Added `client_side_cache` field**: Updated `ConnectionRequest` creation to include the new `client_side_cache` field from the latest glide-core.

### Limitations

- Client-side caching configuration is stubbed out (`client_side_cache: None`) - full support to be added seperately

### Testing

All 68 compression integration tests pass:

- Basic compression/decompression with Zstd and Lz4
- Cluster and standalone client compression
- Minimum size threshold handling
- Custom compression levels
- Backward compatibility with uncompressed data
- Multiple operations data integrity
- Binary data handling
- Compression statistics
- Multi-key operations (MGET, MSET, MSETNX)
- CustomCommand compression (SETEX, PSETEX, SETNX via CustomCommand)
- Incompatible command error handling (APPEND, GETRANGE, SETRANGE, STRLEN, INCR, INCRBY, INCRBYFLOAT, DECR, DECRBY, GETBIT, SETBIT, BITCOUNT via both native methods and CustomCommand)
- 
### Checklist


rust/src/lib.rs
```diff
@@ -525,6 +525,7 @@ pub unsafe extern "C-unwind" fn command(
     let mut cmd = match unsafe { create_cmd(cmd_ptr) } {
         Ok(cmd) => cmd,
         Err(err) => {
+            panic_guard.panicked = false;
```
currantw, line 528: Thanks for catching this! 🫡

tests/Valkey.Glide.IntegrationTests/CompressionTests.cs
```diff
@@ -453,4 +453,430 @@ public async Task Compression_Setnx_CompressesValue(GlideClient _)
...
+            .WithCompression(CompressionConfig.Zstd())
+            .WithTls(TestConfiguration.TLS)
+            .Build();
+
+        await using var client = await GlideClient.CreateClient(clientConfig);
+
+        var statsBefore = BaseClient.GetStatistics();
+
+        string key = $"setex_custom_test_{Guid.NewGuid()}";
+        string value = new('s', LargeValueSize);
```
currantw: Should we just make this a static property of the class so we don't need to create a new "large" string every time? 🤷

tests/Valkey.Glide.IntegrationTests/CompressionTests.cs
```diff
@@ -453,4 +453,430 @@ public async Task Compression_Setnx_CompressesValue(GlideClient _)
...
+    [MemberData(nameof(Config.TestStandaloneClients), MemberType = typeof(TestConfiguration))]
+    public async Task Compression_SetExViaCustomCommand_CompressesValue(GlideClient _)
+    {
+        var clientConfig = new StandaloneClientConfigurationBuilder()
+            .WithAddress(TestConfiguration.STANDALONE_ADDRESS.Host, TestConfiguration.STANDALONE_ADDRESS.Port)
+            .WithCompression(CompressionConfig.Zstd())
+            .WithTls(TestConfiguration.TLS)
+            .Build();
+
+        await using var client = await GlideClient.CreateClient(clientConfig);
```
currantw: - Can we extract this client – perhaps to a test fixture? Seems like every method is re-creating a new client (with the exact same, duplicated configuration), when they could all just use the same one.
- Are we able to just pass `TestConfiguration.STANDALONE_ADDRESS` to `WithAddress` instead of separate host and port (just a touch simpler).
- Any idea why this is using TLS? By default, most of our tests don't, I believe. Is there some specific aspect of compression that we are trying to test that requires TLS?

tests/Valkey.Glide.IntegrationTests/CompressionTests.cs
```diff
@@ -453,4 +453,430 @@ public async Task Compression_Setnx_CompressesValue(GlideClient _)
...
+        Assert.Equal("OK", result?.ToString());
+
+        var statsAfter = BaseClient.GetStatistics();
+        Assert.True(statsAfter.TotalValuesCompressed > statsBefore.TotalValuesCompressed,
+            $"PSETEX via CustomCommand should compress value. Before: {statsBefore.TotalValuesCompressed}, After: {statsAfter.TotalValuesCompressed}");
+
+        // Verify value can be retrieved and decompressed
+        ValkeyValue retrieved = await client.GetAsync(key);
+        Assert.Equal(value, retrieved.ToString());
+    }
```
currantw, line 494: Should we also test TTL here?

tests/Valkey.Glide.IntegrationTests/CompressionTests.cs
```diff
@@ -453,4 +453,430 @@ public async Task Compression_Setnx_CompressesValue(GlideClient _)
...
+            $"SETEX via CustomCommand should compress value. Before: {statsBefore.TotalValuesCompressed}, After: {statsAfter.TotalValuesCompressed}");
+
+        // Verify value can be retrieved and decompressed
+        ValkeyValue retrieved = await client.GetAsync(key);
+        Assert.Equal(value, retrieved.ToString());
+
+        // Verify TTL was set
+        var ttl = await client.TimeToLiveAsync(key);
+        Assert.True(ttl.HasTimeToLive);
+        Assert.True(ttl.TimeToLive!.Value.TotalSeconds > 0 && ttl.TimeToLive!.Value.TotalSeconds <= 10);
```
currantw: Nit. Could use `InRange` instead? Apparently it gives a more useful error message on failure:


```suggestion
	Assert.InRange(ttl.TimeToLive!.Value, TimeSpan.FromSeconds(1), TimeSpan.FromSeconds(10));
```

(InRange uses inclusive bounds and Valkey TTL values are integer seconds, which is why the above uses `TimeSpan.FromSeconds(1)` instead of `TimeSpan.Zero`).

tests/Valkey.Glide.IntegrationTests/CompressionTests.cs
```diff
@@ -453,4 +453,430 @@ public async Task Compression_Setnx_CompressesValue(GlideClient _)
...
+            .WithTls(TestConfiguration.TLS)
+            .Build();
+
+        await using var client = await GlideClient.CreateClient(clientConfig);
+
+        // Ensure key doesn't exist
+        string key = $"setnx_custom_test_{Guid.NewGuid()}";
+#pragma warning disable IDE0058 // Expression value is never used
+        await client.DeleteAsync(key);
+#pragma warning restore IDE0058
```
currantw: Why is this necessary? We are using a random key, so can't we just assume it doesn't exist (like all the other tests)?

tests/Valkey.Glide.IntegrationTests/CompressionTests.cs
```diff
@@ -453,4 +453,430 @@ public async Task Compression_Setnx_CompressesValue(GlideClient _)
...
+        var clientConfig = new StandaloneClientConfigurationBuilder()
+            .WithAddress(TestConfiguration.STANDALONE_ADDRESS.Host, TestConfiguration.STANDALONE_ADDRESS.Port)
+            .WithCompression(CompressionConfig.Zstd())
+            .WithTls(TestConfiguration.TLS)
+            .Build();
+
+        await using var client = await GlideClient.CreateClient(clientConfig);
+
+        string key = $"incrby_test_{Guid.NewGuid()}";
+        await client.SetAsync(key, "100");
```
currantw: Is it even necessary to add the key – won't this test (and others like it) fail regardless of whether the key exists?


```suggestion
```

tests/Valkey.Glide.IntegrationTests/CompressionTests.cs
```diff
@@ -453,4 +453,430 @@ public async Task Compression_Setnx_CompressesValue(GlideClient _)
...
+
+        await using var client = await GlideClient.CreateClient(clientConfig);
+
+        string key = $"incr_custom_test_{Guid.NewGuid()}";
+        await client.SetAsync(key, "100");
+
+        var exception = await Assert.ThrowsAsync<Errors.RequestException>(async () =>
+            await client.CustomCommand(["INCR", key]));
+
+        Assert.Contains("compression", exception.Message, StringComparison.OrdinalIgnoreCase);
```
currantw: Can these "Incompatible Commands via CustomCommand" be parametrized to reduce duplicate? Each set of data could just be the custom command arguments. (Assuming, as mentioned in a previous comment, that the setup also isn't needed).

This would also let us add tests for _all_ applicable custom commands. Looks like we are currently not testing:
- BITPOS
- BITFIELD
- BITFIELD_RO
- BITOP
- LCS
- SUBSTR*

rust/src/lib.rs
currantw: This seems to duplicate some of the logic you added above. Should we extract to a helper?

currantw, line 1: > Batch/pipeline compression for CustomCommand is not yet implemented

I assume that this because it is not implemented in core yet?

currantw (commented):
Reviewed to `8e41966`.

tests/Valkey.Glide.IntegrationTests/CompressionTests.cs
alexr-bq: Addressed all

tests/Valkey.Glide.IntegrationTests/CompressionTests.cs
alexr-bq: We wouldn't want it to fail for the wrong reason

rust/src/lib.rs
alexr-bq: Done

alexr-bq: Yes, but this isn't good and I'm addressing it.

alexr-bq: Pulled in latest main and added support for batch

tests/Valkey.Glide.IntegrationTests/CompressionTests.cs
alexr-bq: Done

PR_COMPRESSION_SUPPORT.md
currantw, line 1: Woops, did you mean to add this?

rust/src/ffi.rs
```diff
@@ -713,19 +715,78 @@ pub struct BatchOptionsInfo {
...
     }
+
     Ok(cmd)
 }
 
+/// Resolves the actual RequestType for a CustomCommand by parsing the command name from the first argument.
+fn resolve_custom_command_type(args: &[Vec<u8>]) -> RequestType {
+    if args.is_empty() {
+        return RequestType::CustomCommand;
+    }
```
currantw, line 788: Under what circumstances might the args be empty?

tests/Valkey.Glide.IntegrationTests/CompressionTests.cs
```diff
@@ -46,64 +78,44 @@ public async Task Compression_Zstd_Standalone_CompressesAndDecompresses(GlideCli
...
 
         var statsAfter = BaseClient.GetStatistics();
         Assert.True(statsAfter.TotalValuesCompressed > statsBefore.TotalValuesCompressed);
     }
 
-    [Theory(DisableDiscoveryEnumeration = true)]
-    [MemberData(nameof(Config.TestClusterClients), MemberType = typeof(TestConfiguration))]
-    public async Task Compression_Zstd_Cluster_CompressesAndDecompresses(GlideClusterClient _)
+    [Fact]
+    public async Task Compression_Zstd_Cluster_CompressesAndDecompresses()
```
currantw, line 106: Should we also test LZ4 with cluster for completeness?

PR_COMPRESSION_SUPPORT.md
alexr-bq: Oops, nope.

rust/src/ffi.rs
alexr-bq: No reasonable / expected case

tests/Valkey.Glide.IntegrationTests/CompressionTests.cs
alexr-bq: Added

rust/src/ffi.rs
currantw: Should we panic or otherwise fail here? Would just returning `CustomCommand` here potentially lead to even hrder-to-debug failures later?

sources/Valkey.Glide/CompressionConfig.cs
```diff
@@ -36,17 +36,26 @@ public sealed class CompressionConfig
...
     /// <param name="backend">The compression backend to use.</param>
     /// <param name="compressionLevel">Optional compression level. If null, uses backend default.</param>
     /// <param name="minCompressionSize">Minimum value size to compress.</param>
+    /// <param name="maxDecompressedSize">Maximum allowed size for decompressed data. If null, uses default (512 MB).</param>
     /// <exception cref="ArgumentOutOfRangeException">Thrown if minCompressionSize is less than 16 bytes.</exception>
     public CompressionConfig(
         CompressionBackend backend,
         int? compressionLevel = null,
-        nuint minCompressionSize = DefaultMinCompressionSize)
+        nuint minCompressionSize = DefaultMinCompressionSize,
```
currantw, line 57: Know this was added previously ... but should `MinCompressionSize` be handled the same? (i.e. by the Rust core, rather than being exposed here)?

currantw: Also not ideal that the min size in an `nuint` while the max size is a `ulong` 😬

alexr-bq: This would involve a protobuf change in the core and I think needs to be out of scope of this PR.

currantw: If that's the case, then it certainly is! Thanks for looking into it.

currantw (approved):

prateek-kumar-improving (approved):

Issue #345, opened 2026-04-22
CI/CD: Extract reusable test workflow and introduce nightly Full Matrix Test


jamesx-improving:
## Problem

The CI (`tests.yml`) and CD (`cd.yml`) workflows define their test matrices and test steps independently. This allows the CD post-publish tests to exercise OS/arch/server-version combinations that CI never covers, creating blind spots that only surface on release day.

This was demonstrated in the v1.0.0-rc3 CD run ([#24592944915](https://github.com/valkey-io/valkey-glide-csharp/actions/runs/24592944915)), where 5 of 6 test jobs failed due to issues that CI never had a chance to catch.

## Proposal

Extract the test job into a **reusable workflow**, parameterized by a `profile` that controls the test matrix size. CI, CD, and a new nightly Full Matrix Test (FMT) workflow all call this reusable workflow with different profiles.

```
CI (PR/push)        ── profile: standard ──▶
CD (post-publish)   ── profile: smoke    ──▶  reusable test workflow
Nightly FMT         ── profile: full     ──▶
```

### Profiles

| Profile | Platform breadth | Server versions | Test depth | Typical callers |
|---|---|---|---|---|
| `smoke` | All OS × arch | One (latest stable) | Minimal subset — enough to confirm native interop and packaging | CD post-publish validation |
| `standard` | Narrower subset | Multiple | Full test suite | CI on PR/push |
| `full` | All OS × arch | All supported | Full test suite | Nightly FMT |

Ordering: `smoke` ⊆ `full` is the hard requirement.

The `standard` profile does not need to cover every platform or server version — it is scoped for fast PR feedback. The `full` profile (nightly FMT) is the one that must cover 100% of what CD's `smoke` profile tests, ensuring no platform-level issue reaches release day undetected.

Note: since valkey-glide ships platform-specific native binaries (Rust core via FFI), the `smoke` profile must cover all platforms. Packaging issues — wrong native binary bundled, missing RID, broken FFI — only manifest on the affected platform. The smoke profile is minimal in test *depth*, not platform *breadth*.

### Reusable workflow responsibilities

- The test matrix (single source of truth, filtered by profile)
- All test execution steps (environment setup, server install, test run, report upload)

An optional `package-version` input allows CD to swap source project references for the published NuGet package.

### Key invariant

[...]

PR #356, opened 2026-04-29
CI: Add reusable test workflow

jamesx-improving:
### Summary

Add a reusable test workflow (`_reusable-test.yml`) that centralizes all test matrix logic and test execution steps. This is Phase 1 of the CI/CD test unification effort — the workflow is created but not yet wired into any callers.

Importantly it introduces the concept of profiles ; 3 profiles were added to the test workflow: smoke, standard, full. Each profile specify a test matrix to run. Ex: the profile `full` will run all configuration, essentially a Ful Matrix test.

### Issue Link

This pull request is linked to issue: https://github.com/valkey-io/valkey-glide-csharp/issues/345

### How it works

  - _reusable-test.yml accepts a profile: smoke, standard, or full.
  - It then use `load_matrices.py`, to load in matrices based on the given profiles
  - Two matrix jobs, runner-tests and container-tests, are spun up based on the profiles.
  - `.github/json_matrices/` is our "run order". To configure profiles or add new OS, you only need to change that folder.

### Limitations

- This is Phase 1 only: the reusable workflow exists but is not called by any workflow yet. Phase 2 will refactor `tests.yml` → `ci.yml`, update `cd.yml`, and create `nightly.yml`
- The `tags` field in `os-matrix.json` is retained for now since `cd.yml` and `create-test-matrices` still depend on it. It will be removed in Phase 2.
- `create-test-matrices` composite action is not deleted yet (still used by `tests.yml`). Will be removed in Phase 2.

### Testing

- YAML validated (parses without errors)
- No existing workflows are modified, so current CI/CD behavior is unchanged
- The reusable workflow will be integration-tested in Phase 2 when callers are wired up

### Related Issues

- Phase 2 PR (upcoming): refactor CI/CD callers to use the reusable workflow, create nightly FMT workflow, remove `tags` from `os-matrix.json`, delete `create-test-matrices`

### Checklist

- [x] ~~`CHANGELOG.md`, `README.md`, `DEVELOPER.md`, and other documentation files are updated.~~ N/A — CI-only change, no user-facing documentation impact.

### DIFF
--- a/.github/json_matrices/README.md
+++ b/.github/json_matrices/README.md
@@ -9,17 +9,21 @@ This directory contains JSON configuration files that define **what** gets teste
 | `os-matrix.json` | OS/runner platforms (VMs and containers) |
 | `server-matrix.json` | Valkey and Redis server versions |
 | `version-matrix.json` | .NET SDK versions |
+| `profiles.json` | Per-profile settings (e.g., test filters) |
 | `load_matrices.py` | Filters all matrices by profile, outputs to `$GITHUB_OUTPUT` |
 
 ## Profiles
 
 Each entry in the JSON files has a `profiles` array that declares which test profiles include it:
 
-| Profile | When used | Scope |
-|---------|-----------|-------|
-| `smoke` | Quick validation (e.g., PR checks) | Minimal: 1 server, core platforms, default .NET version |
-| `standard` | Regular CI (e.g., push to main) | Moderate: common platforms and servers |
-| `full` | Nightly / manual | Everything, including containers |
+
+| Profile    | When used                                                                  |
+| ---------- | -------------------------------------------------------------------------- |
+| `smoke`    | For quick validation. Runs on wide platforms but only against CommandTests |
+| `standard` | Standard tests run on common platform                                      |
+| `full`     | Nightly / manual run on all platforms and all tests                        |
+
+Per-profile settings like test filters are defined in `profiles.json`.
 
 ### Example
 
@@ -57,6 +61,7 @@ The script outputs four JSON arrays:
 | `container-host-matrix` | Container hosts (entries with `IMAGE`) matching the profile |
 | `server-matrix` | Server versions matching the profile |
 | `dotnet-matrix` | .NET versions matching the profile |
+| `test-filter` | Test filter expression from `profiles.json` (empty = run all tests) |
 
 These feed into the `test` and `test-container` job strategy matrices.
 
--- a/.github/json_matrices/README.md
+++ b/.github/json_matrices/README.md
@@ -46,24 +46,7 @@ This host runs in all three profiles. A host with `"profiles": ["full"]` only ru
 
 ## How it works
 
-The reusable workflow `_reusable-test.yml` calls the script:
-
-```yaml
-- name: Load and filter matrices
-  run: python3 load_matrices.py ${{ inputs.profile }}
-```
-
-The script outputs four JSON arrays:
-
-| Output | Description |
-|--------|-------------|
-| `host-matrix` | VM hosts (entries without `IMAGE`) matching the profile |
-| `container-host-matrix` | Container hosts (entries with `IMAGE`) matching the profile |
-| `server-matrix` | Server versions matching the profile |
-| `dotnet-matrix` | .NET versions matching the profile |
-| `test-filter` | Test filter expression from `profiles.json` (empty = run all tests) |
-
-These feed into the `test` and `test-container` job strategy matrices.
+The reusable workflow `_reusable-test.yml` calls the `load_matrices.py` script for a specified profile, which generates JSON arrays representing the corresponding host, server, .NET version, and test filter.s
 
 ### Local testing
 
--- a/.github/json_matrices/README.md
+++ b/.github/json_matrices/README.md
@@ -18,7 +18,7 @@ Each entry in the JSON files has a `profiles` array that declares which test pro
 | Profile | When used | Scope |
 |---------|-----------|-------|
 | `smoke` | Quick validation (e.g., PR checks) | Minimal: 1 server, core platforms, default .NET version |
-| `standard` | Regular CI (e.g., push to main) | Moderate: `always`-run platforms and servers |
+| `standard` | Regular CI (e.g., push to main) | Moderate: common platforms and servers |
 | `full` | Nightly / manual | Everything, including containers |
 
 ### Example
--- a/.github/json_matrices/README.md
+++ b/.github/json_matrices/README.md
@@ -1,0 +1,82 @@
+# CI/CD Test Matrices
+
+This directory contains JSON configuration files that define **what** gets tested and a Python script that filters them by profile.
+
+## Files
+
+| File | Purpose |
+|------|---------|
+| `os-matrix.json` | OS/runner platforms (VMs and containers) |
+| `server-matrix.json` | Valkey and Redis server versions |
+| `version-matrix.json` | .NET SDK versions |
+| `load_matrices.py` | Filters all matrices by profile, outputs to `$GITHUB_OUTPUT` |
+
+## Profiles
+
+Each entry in the JSON files has a `profiles` array that declares which test profiles include it:
+
+| Profile | When used | Scope |
+|---------|-----------|-------|
+| `smoke` | Quick validation (e.g., PR checks) | Minimal: 1 server, core platforms, default .NET version |
+| `standard` | Regular CI (e.g., push to main) | Moderate: `always`-run platforms and servers |
+| `full` | Nightly / manual | Everything, including containers |
+
+### Example
+
+```json
+{
+    "OS": "ubuntu",
+    "RUNNER": "ubuntu-24.04",
+    "profiles": ["smoke", "standard", "full"]
+}
+```
+
+This host runs in all three profiles. A host with `"profiles": ["full"]` only runs in the full suite.
+
+## Adding a new platform or server
+
+1. Add an entry to the appropriate JSON file.
+2. Set `profiles` to the profiles it should run in.
+3. For OS entries, set `tags` for CD pipeline inclusion (see below).
+4. No code changes needed — the Python script filters generically.
+
+## How it works
+
+The reusable workflow `_reusable-test.yml` calls the script:
+
+```yaml
+- name: Load and filter matrices
+  run: python3 load_matrices.py ${{ inputs.profile }}
+```
+
+The script outputs four JSON arrays:
+
+| Output | Description |
+|--------|-------------|
+| `host-matrix` | VM hosts (entries without `IMAGE`) matching the profile |
+| `container-host-matrix` | Container hosts (entries with `IMAGE`) matching the profile |
+| `server-matrix` | Server versions matching the profile |
+| `dotnet-matrix` | .NET versions matching the profile |
+
+These feed into the `test` and `test-container` job strategy matrices.
+
+### Local testing
+
+```bash
+cd .github/json_matrices
+python3 load_matrices.py smoke
+python3 load_matrices.py standard
+python3 load_matrices.py full
+```
+
+Without `$GITHUB_OUTPUT` set, the script prints results to stdout.
+
+## Tags (CD pipeline)
+
+The `tags` field in `os-matrix.json` is separate from profiles. It controls which platforms the CD pipeline (`cd.yml`) builds binaries for:
+
+- `"ci"` — included in CI test runs
+- `"cd"` — included in CD binary builds
+- `"container"` — container-based host
+
+A host can have tags but empty profiles (e.g., Windows ARM is tagged `"cd"` for binary builds but has no test profiles since it can't run the server).
--- a/.github/json_matrices/load_matrices.py
+++ b/.github/json_matrices/load_matrices.py
@@ -1,0 +1,74 @@
+#!/usr/bin/env python3
+"""
+Load and filter CI test matrices based on a profile (smoke, standard, full).
+
+Outputs JSON arrays to $GITHUB_OUTPUT for use in GitHub Actions matrix strategies.
+
+To add a new or update a run configuration, change json_matrices with the appropriate profiles.
+"""
+
+import json
+import os
+import sys
+from pathlib import Path
+
+VALID_PROFILES = ("smoke", "standard", "full")
+
+
+def load_json(filename: str) -> list | dict:
+    return json.loads(Path(filename).read_text())
+
+
+def filter_by_profile(entries: list, profile: str) -> list:
+    return [e for e in entries if profile in e.get("profiles", [])]
+
+
+def main() -> None:
+    if len(sys.argv) != 2 or sys.argv[1] not in VALID_PROFILES:
+        print(f"Usage: {sys.argv[0]} <{'|'.join(VALID_PROFILES)}>", file=sys.stderr)
+        sys.exit(1)
+
+    profile = sys.argv[1]
+
+    os_matrix = filter_by_profile(load_json("os-matrix.json"), profile)
+    host = [h for h in os_matrix if "IMAGE" not in h]
+    container_host = [h for h in os_matrix if "IMAGE" in h]
+
+    server = filter_by_profile(load_json("server-matrix.json"), profile)
+    
+    dotnet_entries = filter_by_profile(load_json("version-matrix.json"), profile)
+    dotnet = [e["version"] for e in dotnet_entries]
+
+    # Validate
+    if profile == "smoke" and len(server) != 1:
+        print(f"ERROR: Expected exactly 1 smoke server, found {len(server)}", file=sys.stderr)
+        sys.exit(1)
+    for name, value in [("HOST_MATRIX", host), ("SERVER_MATRIX", server), ("DOTNET_MATRIX", dotnet)]:
+        if not value:
+            print(f"ERROR: {name} is empty for profile={profile}", file=sys.stderr)
+            sys.exit(1)
+
+    # Summary
+    print(f"Profile: {profile}")
+    print(f"VM hosts: {len(host)}")
+    print(f"Container hosts: {len(container_host)}")
+    print(f"Servers: {len(server)}")
+    print(f"Dotnet versions: {len(dotnet)}")
+
+    # Write to GITHUB_OUTPUT
+    github_output = os.environ.get("GITHUB_OUTPUT")
+    if github_output:
+        with open(github_output, "a") as f:
+            f.write(f"host-matrix={json.dumps(host, separators=(',', ':'))}\n")
+            f.write(f"container-host-matrix={json.dumps(container_host, separators=(',', ':'))}\n")
+            f.write(f"server-matrix={json.dumps(server, separators=(',', ':'))}\n")
+            f.write(f"dotnet-matrix={json.dumps(dotnet, separators=(',', ':'))}\n")
+    else:
+        print(f"host-matrix={json.dumps(host, indent=2)}")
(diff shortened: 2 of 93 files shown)


.github/workflows/_reusable-test.yml
```diff
@@ -0,0 +1,312 @@
...
+              echo "ERROR: package-artifact-name requires package-version to also be set"
+              exit 1
+          fi
+
+      - uses: actions/checkout@v6
+
+      - name: Load and filter matrices
+        id: load
+        working-directory: .github/json_matrices
+        run: python3 load_matrices.py ${{ inputs.profile }}
```
xShinnRyuu: ${{ inputs.profile }} is interpolated directly into a run: shell command. Since this is a workflow_call input, the caller controls the value. If a future caller passes an unsanitized value, or if this is ever changed to workflow_dispatch, this becomes a command injection vector.

I think the best choice here is to make it an environment variable. 

Reference: https://securitylab.github.com/resources/github-actions-untrusted-input/

.github/workflows/_reusable-test.yml
```diff
@@ -0,0 +1,312 @@
...
+    outputs:
+      host-matrix: ${{ steps.load.outputs.host-matrix }}
+      container-host-matrix: ${{ steps.load.outputs.container-host-matrix }}
+      server-matrix: ${{ steps.load.outputs.server-matrix }}
+      dotnet-matrix: ${{ steps.load.outputs.dotnet-matrix }}
+    steps:
+      - name: Validate inputs
+        shell: bash
+        run: |
+          if [[ -n "${{ inputs.package-artifact-name }}" && -z "${{ inputs.package-version }}" ]]; then
```
xShinnRyuu: Both `${{ inputs.package-artifact-name }}` and `${{ inputs.package-version }}` have the same issue as ${{ inputs.profile }} below

DEVELOPER.md
```diff
@@ -381,6 +381,16 @@ Performance benchmarking for the C# client can be performed using [resp-bench](h
 
 Refer to the [resp-bench README](https://github.com/ikolomi/resp-bench/blob/main/README.md) and [C# benchmark docs](https://github.com/ikolomi/resp-bench/blob/main/docs/BENCHMARKS_CSHARP.md) for setup and usage instructions.
 
+## Updating CI Test Matrices
+
+To update our Github workflow test runs, edit the JSON files in [`.github/json_matrices/`](.github/json_matrices/). 
+
+It serves as a "test order", describing available properties and which `profiles` they belong to (`smoke`, `standard`, `full`). No workflow or script changes are needed.
+
+Ex: To add a new platform to the `standard` profile, add a new entry to `.github/json_matrices/os_matrices.json` with `profiles: ['standard']`
```
xShinnRyuu: typo: `os_matrices.json` should be `os-matrix.json` i think

.github/workflows/_reusable-test.yml
```diff
@@ -0,0 +1,312 @@
...
+          dotnet test
+          --configuration Debug
+          --framework net${{ matrix.dotnet }}
+          ${{ inputs.profile == 'smoke' && '--filter "FullyQualifiedName~CommandTests"' || '' }}
+          --logger "html;LogFileName=TestReport.html"
+          --logger "trx;LogFileName=TestResults.trx"
+          --logger "console;verbosity=detailed"
+          --results-directory .
+
+      - name: Derive failures.json (from TRX)
```
xShinnRyuu: This `Derive failures.json (from TRX)` step is duplicated on line 274 as well. Is it possible to refactor this?

xShinnRyuu (dismissed):
The request change is really only for the security gap using the raw `inputs` on run commands.

.github/json_matrices/os-matrix.json
```diff
@@ -6,39 +6,43 @@
         "ARCH": "x64",
         "TARGET": "x86_64-unknown-linux-gnu",
         "tags": ["ci", "cd"],
-        "run": "always"
+        "profiles": ["smoke", "standard", "full"]
```
jeremyprime, line 9: I see `standard` only on a couple platforms and `smoke` on many of them, maybe those 2 should be swapped.  `smoke` should be the one that only runs a couple platforms to perform a quick validation.

Should also double check `server-matrix.json`.

Aryex: So actually there's a smoke test filter that I missed. Smoke tests covers a wider platforms but only run CommandTests so it is shallow.

https://github.com/valkey-io/valkey-glide-csharp/pull/356/changes#diff-a21e815279508224f10b5bb96009c315ddb6ad7c8e60c7fa38e2ef7eb13357caR150

That asid, I will move this to the profiles level as well to stay consistent with our json configuration approach.

Aryex: Addressed in https://github.com/valkey-io/valkey-glide-csharp/pull/356/changes/a9399d2077f856c50b647224ec2bc7f65159a39d

.github/workflows/_reusable-test.yml
Aryex: Addressed in https://github.com/valkey-io/valkey-glide-csharp/pull/356/commits/63ac46c3fb67276c221cd1cb8223427d6a290c05

.github/workflows/_reusable-test.yml
Aryex: Addressed in https://github.com/valkey-io/valkey-glide-csharp/pull/356/commits/63ac46c3fb67276c221cd1cb8223427d6a290c05

DEVELOPER.md
Aryex: Fixed

.github/workflows/_reusable-test.yml
```diff
@@ -0,0 +1,304 @@
...
+          JOB_NAME: "net${{ matrix.dotnet }}, server ${{ matrix.server.version }}, ${{ matrix.host.TARGET }}"
+          DOTNET: ${{ matrix.dotnet }}
+          SERVER_TYPE: ${{ matrix.server.type }}
+          SERVER_VERSION: ${{ matrix.server.version }}
+          HOST_OS: ${{ matrix.host.OS }}
+          HOST_ARCH: ${{ matrix.host.ARCH }}
+          HOST_RUNNER: ${{ matrix.host.RUNNER }}
+          RUN_ID: ${{ github.run_id }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+        run: .github/workflows/scripts/derive_failures.sh
```
xShinnRyuu: Is `derive_failures.sh` suppose to be `create_failures_json.sh` here?

Aryex: fixed.

xShinnRyuu (approved):

.github/json_matrices/profiles.json
```diff
@@ -0,0 +1,11 @@
+{
+    "smoke": {
+        "test-filter": "FullyQualifiedName~CommandTests"
+    },
+    "standard": {
+        "test-filter": ""
+    },
+    "full": {
+        "test-filter": ""
+    }
```
currantw, line 10: Perhaps a `test-filters` array would be better for future extensibility?

.github/json_matrices/os-matrix.json
```diff
@@ -6,39 +6,43 @@
...
+        "profiles": ["smoke", "full"]
     },
     {
         "OS": "macos",
         "NAMED_OS": "darwin",
         "RUNNER": "macos-15-intel",
         "ARCH": "x64",
         "TARGET": "x86_64-apple-darwin",
-        "tags": ["ci", "cd"]
+        "tags": ["ci", "cd"],
```
currantw, line 44: I was just gonna comment that we should remove these tags and use this new profile system instead – but it seems that @jamesx-improving has already thought of this and it is planned for Phase 2. Huzzah! 🎉

docs/ci-cd.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+This host runs in all three profiles. A host with `"profiles": ["full"]` only runs in the full suite.
+
+## Adding a new platform or server
+
+1. Add an entry to the appropriate JSON file.
+2. Set `profiles` to the profiles it should run in.
+3. For OS entries, set `tags` for CD pipeline inclusion (see below).
+4. No code changes needed — the Python script filters generically.
```
currantw: Not sure exactly what this means? The actions above definitely seem like "code changes"..?

.github/json_matrices/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+| Output | Description |
+|--------|-------------|
+| `host-matrix` | VM hosts (entries without `IMAGE`) matching the profile |
+| `container-host-matrix` | Container hosts (entries with `IMAGE`) matching the profile |
+| `server-matrix` | Server versions matching the profile |
+| `dotnet-matrix` | .NET versions matching the profile |
+| `test-filter` | Test filter expression from `profiles.json` (empty = run all tests) |
+
+These feed into the `test` and `test-container` job strategy matrices.
```
currantw: Perhaps we could just condense this a bit? Seems like we are duplicating information about how the script actually work, which is probably best left to the script documentation? 🤷 Also, is it outputting four _or five_ JSON arrays?


```suggestion
## How it works

The reusable workflow `_reusable-test.yml` calls the `load_matrices.py` script for a specified profile, which generates JSON arrays representing the corresponding host, server, .NET version, and test filter.s
```

docs/ci-cd.md
```diff
@@ -0,0 +1,87 @@
...
+### Local testing
+
+```bash
+cd .github/json_matrices
+python3 load_matrices.py smoke
+python3 load_matrices.py standard
+python3 load_matrices.py full
+```
+
+Without `$GITHUB_OUTPUT` set, the script prints results to stdout.
```
currantw: This also seems like information that can just be part of the script documentation?

docs/ci-cd.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Tags (CD pipeline)
+
+The `tags` field in `os-matrix.json` is separate from profiles. It controls which platforms the CD pipeline (`cd.yml`) builds binaries for:
+
+- `"ci"` — included in CI test runs
+- `"cd"` — included in CD binary builds
+- `"container"` — container-based host
+
+A host can have tags but empty profiles (e.g., Windows ARM is tagged `"cd"` for binary builds but has no test profiles since it can't run the server).
```
currantw: Do we want to flag this section with a TODO that links to the Phase 2 issue, so that we make sure we remove it when we tackle that one?

docs/ci-cd.md
currantw, line 1: I like the idea of adding this documentation. I certainly think it is helpful! However, I wonder if it would be better off in the docs folder? I'm not sure what is _actually_ more likely to get noticed by developers and, increasingly, agents, but I am wary of having README files sprinkled throughout the code base. 🤷

DEVELOPER.md
```diff
@@ -381,6 +381,16 @@ Performance benchmarking for the C# client can be performed using [resp-bench](h
...
 
+## Updating CI Test Matrices
+
+To update our Github workflow test runs, edit the JSON files in [`.github/json_matrices/`](.github/json_matrices/). 
+
+It serves as a "test order", describing available properties and which `profiles` they belong to (`smoke`, `standard`, `full`). No workflow or script changes are needed.
+
+Ex: To add a new platform to the `standard` profile, add a new entry to `.github/json_matrices/os_matrix.json` with `profiles: ['standard']`
+
+See the [CI/CD Test Matrices README](.github/json_matrices/README.md) for details.
```
currantw: I'm not sure if this warrants its own sections? Perhaps a generic GitHub Workflows or CI/CD section might work? It could include a link to that documentation, and we could add links to other relevant docs as we write them?

.github/json_matrices/load_matrices.py
```diff
@@ -0,0 +1,81 @@
...
+
+To add a new or update a run configuration, change json_matrices with the appropriate profiles.
+"""
+
+import json
+import os
+import sys
+from pathlib import Path
+
+VALID_PROFILES = ("smoke", "standard", "full")
```
currantw: We already load `profiles.json`. Can we just get these from there, rather than duplicating here?

.github/json_matrices/load_matrices.py
```diff
@@ -0,0 +1,81 @@
...
+    test_filter = profile_settings.get("test-filter", "")
+
+    # Validate
+    if profile == "smoke" and len(server) != 1:
+        print(f"ERROR: Expected exactly 1 smoke server, found {len(server)}", file=sys.stderr)
+        sys.exit(1)
+    for name, value in [("HOST_MATRIX", host), ("SERVER_MATRIX", server), ("DOTNET_MATRIX", dotnet)]:
+        if not value:
+            print(f"ERROR: {name} is empty for profile={profile}", file=sys.stderr)
+            sys.exit(1)
```
currantw: Nit. Perhaps better to just validate each matrix immediately after creating it? Might be a teensy bit of duplication, but seems a bit easier sequentially for me? Plus it avoid the need to load information if we are already going to fail (i.e. if `host` is empty, we don't need to bother loading server or dotnet matrices.

Does this also work is `container_host` is empty by `host` isn't? That seems like a valid potential case?

.github/json_matrices/load_matrices.py
```diff
@@ -0,0 +1,81 @@
...
+    print(f"Container hosts: {len(container_host)}")
+    print(f"Servers: {len(server)}")
+    print(f"Dotnet versions: {len(dotnet)}")
+    print(f"Test filter: {test_filter or '(none)'}")
+
+    # Write to GITHUB_OUTPUT
+    github_output = os.environ.get("GITHUB_OUTPUT")
+    if github_output:
+        with open(github_output, "a") as f:
+            f.write(f"host-matrix={json.dumps(host, separators=(',', ':'))}\n")
```
currantw: Any reason we need to explicitly specify these separators? Would the default work? If there is a specific reason (e.g. slightly more compact), perhaps extract to a constants and/or add a comment to explain why?

currantw: Perhaps simpler:


```suggestion
    outputs = {
        "host-matrix": json.dumps(host),
        "container-host-matrix": json.dumps(container_host),
        "server-matrix": json.dumps(server),
        "dotnet-matrix": json.dumps(dotnet),
        "test-filter": test_filter,
    }

    github_output = os.environ.get("GITHUB_OUTPUT")
    if github_output:
        with open(github_output, "a") as f:
            for key, value in outputs.items():
               print(f"{key}={value}", file=f)
    else:
        for key, value in outputs.items():
            print(f"{key}={value}")
```

(Also using print with `file` arg to get rid of the newline)

.github/workflows/_reusable-test.yml
```diff
@@ -0,0 +1,303 @@
+name: Reusable Test Workflow
+
+on:
+  workflow_call:
+    inputs:
+      profile:
+        required: true
+        type: string
+        description: "Test profile: smoke, standard, or full"
```
currantw: Nit. Probably should't couple this to the specific names 🤷 
```suggestion
        description: "Test profile"
```

.github/workflows/_reusable-test.yml
```diff
@@ -0,0 +1,303 @@
...
+          fi
+
+      - uses: actions/checkout@v6
+
+      - name: Load and filter matrices
+        id: load
+        working-directory: .github/json_matrices
+        env:
+          PROFILE: ${{ inputs.profile }}
+        run: python3 load_matrices.py "$PROFILE"
```
currantw: Perhaps worth separating validation?

.github/workflows/_reusable-test.yml
```diff
@@ -0,0 +1,303 @@
...
+        env:
+          PACKAGE_ARTIFACT_NAME: ${{ inputs.package-artifact-name }}
+          PACKAGE_VERSION: ${{ inputs.package-version }}
+        run: |
+          if [[ -n "$PACKAGE_ARTIFACT_NAME" && -z "$PACKAGE_VERSION" ]]; then
+              echo "ERROR: package-artifact-name requires package-version to also be set"
+              exit 1
+          fi
+
+      - uses: actions/checkout@v6
```
currantw: We only need to checkout one file, right?

```yml
- uses: actions/checkout@v6
  with:
    sparse-checkout: .github/json_matrices
```

.github/workflows/_reusable-test.yml
```diff
@@ -0,0 +1,303 @@
...
+
+      - name: Load and filter matrices
+        id: load
+        working-directory: .github/json_matrices
+        env:
+          PROFILE: ${{ inputs.profile }}
+        run: python3 load_matrices.py "$PROFILE"
+
+  runner-tests:
+    name: net${{ matrix.dotnet }}, server ${{ matrix.server.version }}, ${{ matrix.host.TARGET }}
```
currantw: While we are at it, maybe a more readable name would be nice? Even just a `test –` prefix would be nice 🤷

.github/workflows/_reusable-test.yml
```diff
@@ -0,0 +1,303 @@
...
+          echo "Job running with the following matrix configuration:"
+          echo "$MATRIX_JSON"
+          echo "Profile: $PROFILE"
+          echo "Package version: $PACKAGE_VERSION"
+
+      - name: Set up dotnet ${{ matrix.dotnet }}
+        uses: actions/setup-dotnet@v5
+        with:
+          dotnet-version: |
+            9
```
currantw: Why do we need to install two versions?

.github/workflows/_reusable-test.yml
```diff
@@ -0,0 +1,303 @@
...
+          --logger "trx;LogFileName=TestResults.trx"
+          --logger "console;verbosity=detailed"
+          --results-directory .
+
+      - name: Create failures.json from TRX
+        if: always()
+        shell: bash
+        env:
+          TRX_FILE: TestResults.trx
+          JOB_NAME: "net${{ matrix.dotnet }}, server ${{ matrix.server.version }}, ${{ matrix.host.TARGET }}"
```
currantw: Maybe save this as env var rather than duplicating here?

.github/workflows/scripts/create_failures_json.sh
currantw, line 1: Where does this come from? I didn't see it mentioned in the issue description. Why is it needed?

Aryex: It was added from this comment. https://github.com/valkey-io/valkey-glide-csharp/pull/356#discussion_r3170973153

DEVELOPER.md
Aryex: I the intention for this part is a quick high-level descriptions on how our CI works and how to change it. It links to `github/json_matrices/README.md` as the comprehensive docs.

.github/json_matrices/load_matrices.py
Aryex: good point. Going to use assertion as we assume that profiles should not produce empty configuration matrices.

.github/workflows/_reusable-test.yml
Aryex: Looks like the project uses C# 13 features that require the .net 9 SDK to build even tho we are tageting .net 8.

.github/workflows/_reusable-test.yml
Aryex: Not sure which you are referring to. The JOB_NAME is unique between the runner and container tests.

docs/ci-cd.md
Aryex: I updated the doc to be more concise. pls take another look.

alexr-bq (approved):

Issue #299, opened 2026-04-13
[Task] Add JSON module command support


prateek-kumar-improving:
JSON module commands have already been added in GO/Python client. Add the same commands for C#.

PR #358, opened 2026-04-29
C# JSON commands

alexr-bq:
### Summary

Adds JSON module C# Commands as a wrapper interface around client.

### Issue Link

This pull request is linked to issue: https://github.com/valkey-io/valkey-glide-csharp/issues/299

This PR adds comprehensive support for the Valkey JSON module commands in the C# client, achieving full parity with the Java client implementation.

**Supported Commands (23 total):**

| Category | Commands |
|----------|----------|
| Core | `JSON.SET`, `JSON.GET`, `JSON.MGET`, `JSON.DEL`, `JSON.FORGET`, `JSON.CLEAR`, `JSON.TYPE` |
| Numeric | `JSON.NUMINCRBY`, `JSON.NUMMULTBY` |
| String | `JSON.STRAPPEND`, `JSON.STRLEN` |
| Array | `JSON.ARRAPPEND`, `JSON.ARRINSERT`, `JSON.ARRINDEX`, `JSON.ARRLEN`, `JSON.ARRPOP`, `JSON.ARRTRIM` |
| Object | `JSON.OBJLEN`, `JSON.OBJKEYS` |
| Boolean | `JSON.TOGGLE` |
| Debug | `JSON.DEBUG MEMORY`, `JSON.DEBUG FIELDS` |
| Other | `JSON.RESP` |

**Key Features:**
- Full JSONPath support (e.g., `$`, `$.field`, `$..recursive`)
- Legacy path support for backward compatibility
- Conditional set operations (`NX`, `XX`)
- Formatting options for `JSON.GET` (indent, newline, space)
- Range options for `JSON.ARRINDEX`
- Batch/pipeline support via `GlideJsonBatch` static methods
- Works with both `GlideClient` (standalone) and `GlideClusterClient` (cluster)

## Implementation

The implementation follows the established patterns in the codebase and mirrors the Java client's approach:

**Source Files:**
- `GlideJson.cs` - Core commands (SET, GET, MGET, DEL, FORGET, CLEAR, TYPE, NUMINCRBY, NUMMULTBY, STRAPPEND, STRLEN)
- `GlideJson.ArrayCommands.cs` - Array operations (ARRAPPEND, ARRINSERT, ARRINDEX, ARRLEN, ARRPOP, ARRTRIM)
- `GlideJson.ObjectCommands.cs` - Object, boolean, debug, and RESP operations
- `GlideJson.Options.cs` - Option classes (`SetCondition`, `GetOptions`, `ArrIndexOptions`)
- `GlideJson.Batch.cs` - Batch support via `GlideJsonBatch` static methods

**Key Implementation Details:**

1. **Static Method Pattern**: All commands are implemented as static async methods on the `GlideJson` class, accepting `BaseClient` as the first parameter. This allows the same methods to work with both standalone and cluster clients.

2. **Batch Support**: The `GlideJsonBatch` class provides static methods matching Java's `JsonBatch` pattern:
   ```csharp
   Batch batch = new Batch(true);
   GlideJsonBatch.Set(batch, "doc", "$", "{\"a\": 1}");
   GlideJsonBatch.Get(batch, "doc", "$");
   object?[]? result = await client.Exec(batch, false);
   ```

3. **Cluster Client Handling**: The `ExecuteCommandAsync` helper properly unwraps `ClusterValue<T>` responses and handles null cases correctly.

4. **Return Types**: Commands return `ValkeyValue`, `ValkeyResult`, or `long` depending on the command semantics, matching the expected response types from the server.

## Limitations

1. **Module Dependency**: JSON commands require the Valkey JSON module to be loaded on the server. Commands will fail with an error if the module is not available.

2. **No StackExchange.Redis Compatibility Layer**: JSON commands are exposed only through the `GlideJson` static class, not through the `IDatabase` abstraction layer, as StackExchange.Redis does not have native JSON module support.

## Testing

**Test Files:**
- `JsonCommandTests.cs` - Core command tests (~2300 lines)
- `JsonArrayCommandTests.cs` - Array command tests (~300 lines)
- `JsonObjectCommandTests.cs` - Object/boolean/debug command tests (~200 lines)
- `JsonBatchTests.cs` - Batch operation tests (~600 lines)

**Test Coverage:**
- All 23 commands tested with both standalone and cluster clients
- JSONPath and legacy path variations
- Edge cases (empty arrays, non-existent keys, type mismatches)
- Conditional operations (NX/XX)
- Batch operations with multiple commands
- Error handling for missing module

### Checklist


.github/workflows/tests.yml
```diff
@@ -197,7 +197,7 @@ jobs:
               run: |
                   dotnet test tests/Valkey.Glide.IntegrationTests/ \
                     --framework net${{ matrix.dotnet }} \
-                    --filter "FullyQualifiedName~ValkeySearchCommand" \
+                    --filter "FullyQualifiedName~ValkeySearchCommand|FullyQualifiedName~JsonCommandTests|FullyQualifiedName~JsonArrayCommandTests|FullyQualifiedName~JsonObjectCommandTests" \
```
alexr-bq: Should be corrected to use modules namespace after merging with Ft refactor

sources/Valkey.Glide/ServerModules/GlideJson.Batch.cs
```diff
@@ -0,0 +1,594 @@
...
+    /// </summary>
+    /// <typeparam name="T">The batch type.</typeparam>
+    /// <param name="batch">The batch to add the command to.</param>
+    /// <param name="key">The key of the JSON document.</param>
+    /// <param name="path">The path within the JSON document.</param>
+    /// <param name="increment">The amount to increment by.</param>
+    /// <returns>The batch for chaining.</returns>
+    public static T NumIncrBy<T>(T batch, GlideString key, GlideString path, double increment)
+        where T : BaseBatch<T>
+        => batch.CustomCommand([GlideJson.JsonNumIncrBy, key, path, increment.ToString()]);
```
xShinnRyuu: `increment.ToString()` may cause issues for locales the use the `,` as decimals, which Valkey will reject or misinterpret

sources/Valkey.Glide/ServerModules/GlideJson.Batch.cs
```diff
@@ -0,0 +1,594 @@
...
+    /// </summary>
+    /// <typeparam name="T">The batch type.</typeparam>
+    /// <param name="batch">The batch to add the command to.</param>
+    /// <param name="key">The key of the JSON document.</param>
+    /// <param name="path">The path within the JSON document.</param>
+    /// <param name="multiplier">The amount to multiply by.</param>
+    /// <returns>The batch for chaining.</returns>
+    public static T NumMultBy<T>(T batch, GlideString key, GlideString path, double multiplier)
+        where T : BaseBatch<T>
+        => batch.CustomCommand([GlideJson.JsonNumMultBy, key, path, multiplier.ToString()]);
```
xShinnRyuu: `multiplier.ToString()` may cause issues for locales the use the `,` as decimals, which Valkey will reject or misinterpret

sources/Valkey.Glide/ServerModules/GlideJson.cs
```diff
@@ -0,0 +1,529 @@
...
+    /// <param name="path">The JSONPath or legacy path within the JSON document.</param>
+    /// <param name="multiplier">The amount to multiply by.</param>
+    /// <returns>
+    /// When a JSONPath is provided, returns a JSON string array of new values (or "null" for non-numeric matches).
+    /// When a legacy path is provided, returns the new value as a string.
+    /// </returns>
+    /// <seealso href="https://valkey.io/commands/json.nummultby/"/>
+    public static async Task<ValkeyValue> NumMultByAsync(BaseClient client, ValkeyKey key, ValkeyValue path, double multiplier)
+    {
+        GlideString[] args = [JsonNumMultBy, ToGlideString(key), ToGlideString(path), multiplier.ToString()];
```
xShinnRyuu: `multiplier.ToString()` may cause issues for locales the use the `,` as decimals, which Valkey will reject or misinterpret

sources/Valkey.Glide/ServerModules/GlideJson.cs
```diff
@@ -0,0 +1,529 @@
...
+    /// <param name="path">The JSONPath or legacy path within the JSON document.</param>
+    /// <param name="increment">The amount to increment by.</param>
+    /// <returns>
+    /// When a JSONPath is provided, returns a JSON string array of new values (or "null" for non-numeric matches).
+    /// When a legacy path is provided, returns the new value as a string.
+    /// </returns>
+    /// <seealso href="https://valkey.io/commands/json.numincrby/"/>
+    public static async Task<ValkeyValue> NumIncrByAsync(BaseClient client, ValkeyKey key, ValkeyValue path, double increment)
+    {
+        GlideString[] args = [JsonNumIncrBy, ToGlideString(key), ToGlideString(path), increment.ToString()];
```
xShinnRyuu: `increment.ToString()` may cause issues for locales the use the `,` as decimals, which Valkey will reject or misinterpret

sources/Valkey.Glide/ServerModules/GlideJson.cs
```diff
@@ -0,0 +1,529 @@
...
+    /// </summary>
+    private static GlideString ToGlideString(ValkeyKey key) => (GlideString)key;
+
+    /// <summary>
+    /// Converts a ValkeyValue to GlideString for command arguments.
+    /// </summary>
+    private static GlideString ToGlideString(ValkeyValue value)
+    {
+        if (value.IsNull)
+            return new GlideString(string.Empty);
```
xShinnRyuu: Are we sure we want to return nulls as empty strings? `null` and `empty string` is not the same thing

tests/Valkey.Glide.IntegrationTests/JsonArrayCommandTests.cs
```diff
@@ -0,0 +1,307 @@
...
+/// </summary>
+[Collection("GlideTests")]
+public class JsonArrayCommandTests(TestConfiguration config)
+{
+    public TestConfiguration Config { get; } = config;
+
+    /// <summary>
+    /// Checks if the JSON module is available on the server.
+    /// </summary>
+    private static async Task<bool> IsJsonModuleAvailable(BaseClient client)
```
xShinnRyuu: This `IsJsonModuleAvailable()` function is duplicated across `JsonCommandTests.cs`, `JsonArrayCommandTests.cs`, `JsonObjectCommandTests.cs`, and `JsonBatchTests.cs` (with slight variations for standalone vs cluster in the batch tests).

Can we extract this to a shared test utility.

sources/Valkey.Glide/ServerModules/GlideJson.cs
```diff
@@ -0,0 +1,529 @@
...
+    /// <summary>
+    /// Alias for <see cref="DelAsync(BaseClient, ValkeyKey, ValkeyValue)"/>.
+    /// Deletes the JSON value at the specified path in the key.
+    /// </summary>
+    /// <param name="client">The Glide client to use for the command.</param>
+    /// <param name="key">The key where the JSON document is stored.</param>
+    /// <param name="path">The JSONPath or legacy path within the JSON document.</param>
+    /// <returns>The number of paths deleted.</returns>
+    /// <seealso href="https://valkey.io/commands/json.forget/"/>
+    public static async Task<long> ForgetAsync(BaseClient client, ValkeyKey key, ValkeyValue path)
```
affonsov: if it is an alias why not just delegate to the delAsync instead of re-implementing

sources/Valkey.Glide/ServerModules/GlideJson.cs
```diff
@@ -0,0 +1,529 @@
...
+            // For single-node responses, return the value (which may be null)
+            return result.HasSingleData ? result.SingleValue : null;
+        }
+        throw new ArgumentException("Unsupported client type. Expected GlideClient or GlideClusterClient.", nameof(client));
+    }
+
+    /// <summary>
+    /// Converts an object result to ValkeyValue.
+    /// </summary>
+    private static ValkeyValue ToValkeyValue(object? result)
```
affonsov: Some commands use ValkeyResult.Create and other ToValkeyValue. Can this be cleaned up to use only one or another?

sources/Valkey.Glide/ServerModules/GlideJson.Batch.cs
affonsov, line 1: Is it missing forget on batch?

sources/Valkey.Glide/ServerModules/GlideJson.Options.cs
```diff
@@ -0,0 +1,181 @@
+// Copyright Valkey GLIDE Project Contributors - SPDX Identifier: Apache-2.0
+
+namespace Valkey.Glide.ServerModules;
+
+public static partial class GlideJson
```
affonsov, line 8: missing documentation

tests/Valkey.Glide.IntegrationTests/ServerModules/JsonArrayCommandTests.cs
```diff
@@ -0,0 +1,307 @@
...
+    {
+        await SkipIfJsonModuleNotAvailable(client);
+
+        string key = GetUniqueKey();
+        string jsonValue = "{\"arr\":[1,2,3]}";
+
+        _ = await GlideJson.SetAsync(client, key, "$", jsonValue);
+        ValkeyResult result = await GlideJson.ArrAppendAsync(client, key, "$.arr", "4", "5");
+
+        Assert.NotNull(result);
```
affonsov, line 32: A lot of batch tests are just verifying if the results are not null. Is that enough for these tests?

affonsov (approved):

sources/Valkey.Glide/ServerModules/GlideJson.cs
alexr-bq: Fixing to throw exception here

sources/Valkey.Glide/ServerModules/GlideJson.cs
alexr-bq: They are two different result types with two different usages

.github/workflows/tests.yml
```diff
@@ -197,7 +197,7 @@ jobs:
               run: |
                   dotnet test tests/Valkey.Glide.IntegrationTests/ \
                     --framework net${{ matrix.dotnet }} \
-                    --filter "FullyQualifiedName~ValkeySearchCommand" \
+                    --filter "FullyQualifiedName~ValkeySearchCommand|FullyQualifiedName~JsonCommandTests|FullyQualifiedName~JsonArrayCommandTests|FullyQualifiedName~JsonObjectCommandTests|FullyQualifiedName~JsonBatchTests" \
```
currantw: I'ver just updated this to `~ServerModules` in mine, which should cover both of our tests (provided the tests are in that namespace).

rust/src/ffi.rs
```diff
@@ -321,6 +321,7 @@ pub(crate) unsafe fn create_connection_request(
             },
         ),
         read_only: config.read_only,
+        node_discovery_mode: glide_core::client::NodeDiscoveryMode::Standard,
```
currantw: How is this change related to this work?

scripts/validate_examples.py
```diff
@@ -59,6 +60,7 @@ class ExamplesValidator:
         "Valkey.Glide.Commands.Options.InfoOptions",
         "Valkey.Glide.Commands.Options.BitFieldOptions",
         "Valkey.Glide.Commands.Options.BitFieldOptions.Encoding",
+        "Valkey.Glide.ServerModules.GlideJson",
```
currantw, line 63: I'm not really sure my we are using `GlideJson` for the name here. It seems like other clients do this? But that doesn't really make sense to me, at least based on my understanding:

- The module is called [valkey-json](https://github.com/valkey-io/valkey-json)
- The commands are all prefixed in Valkey with `JSON.*`

To me, this is a case where we should just call the module `Json` – even if it diverges from the other clients – unless there is a clear reason why `GlideJson` is better that I am missing?

currantw: @edlng Any idea why the name GlideJson was used instead of just Json?

rust/src/ffi.rs
alexr-bq: It's not, slipped in with some conflict resolution, will revert

sources/Valkey.Glide/ServerModules/GlideJson.Options.cs
```diff
@@ -0,0 +1,184 @@
...
+/// <summary>
+/// Options and enums for JSON commands.
+/// </summary>
+public static partial class GlideJson
+{
+    /// <summary>
+    /// Specifies the condition for JSON.SET command execution.
+    /// </summary>
+    /// <seealso href="https://valkey.io/commands/json.set/"/>
+    public enum JsonSetCondition
```
currantw: Do we need `Json` prefix since this is already inside `GlideJson`?

sources/Valkey.Glide/ServerModules/GlideJson.Options.cs
```diff
@@ -0,0 +1,184 @@
...
+        /// Converts options to command arguments.
+        /// </summary>
+        /// <returns>An array of <see cref="GlideString"/> containing the command arguments.</returns>
+        internal GlideString[] ToArgs()
+        {
+            List<GlideString> args = [];
+
+            if (Indent is not null)
+            {
+                args.Add("INDENT");
```
currantw: Should add this to `ValkeyLiterals`. Same for "NEWLINE", "SPACE", and any others in this PR.

sources/Valkey.Glide/ServerModules/GlideJson.Options.cs
```diff
@@ -0,0 +1,184 @@
...
+        public long Start { get; init; } = 0;
+
+        /// <summary>
+        /// The end index (exclusive). If null, searches to end of array.
+        /// </summary>
+        /// <remarks>
+        /// Negative values indicate offsets from the end of the array.
+        /// When null, the search continues to the end of the array.
+        /// </remarks>
+        public long? End { get; init; }
```
currantw, line 148: Is -1 a better default? Could add a constant like I did in `IndexRange`?

sources/Valkey.Glide/ServerModules/GlideJson.Options.cs
```diff
@@ -0,0 +1,184 @@
...
+    /// var fromStart = GlideJson.ArrIndexOptions.FromStart(2);
+    /// var inRange = GlideJson.ArrIndexOptions.InRange(0, 5);
+    ///
+    /// // Using object initializer
+    /// var options = new GlideJson.ArrIndexOptions { Start = 1, End = 10 };
+    /// </code>
+    /// </example>
+    /// </remarks>
+    /// <seealso href="https://valkey.io/commands/json.arrindex/">valkey.io</seealso>
+    public sealed class ArrIndexOptions
```
currantw: See existing `IndexRange`. That support reversable ranges (i.e start can be greater than end), but probably should align naming between then? 🤷

sources/Valkey.Glide/ServerModules/GlideJson.Options.cs
```diff
@@ -0,0 +1,184 @@
...
+        /// When null, the search continues to the end of the array.
+        /// </remarks>
+        public long? End { get; init; }
+
+        /// <summary>
+        /// Creates options with a start index.
+        /// </summary>
+        /// <param name="start">The starting index (inclusive).</param>
+        /// <returns>A new <see cref="ArrIndexOptions"/> instance.</returns>
+        public static ArrIndexOptions FromStart(long start) => new() { Start = start };
```
currantw: If we have this, probably good to have `ToEnd` to symmetry?

sources/Valkey.Glide/ServerModules/GlideJson.Options.cs
```diff
@@ -0,0 +1,184 @@
...
+
+        /// <summary>
+        /// Creates options with a start and end index.
+        /// </summary>
+        /// <param name="start">The starting index (inclusive).</param>
+        /// <param name="end">The ending index (exclusive).</param>
+        /// <returns>A new <see cref="ArrIndexOptions"/> instance.</returns>
+        public static ArrIndexOptions InRange(long start, long end) => new() { Start = start, End = end };
```
currantw: Similar static builder is called `Between` in `IndexRange`

.github/workflows/tests.yml
alexr-bq: Moved tests into ServerModules

sources/Valkey.Glide/ServerModules/GlideJson.Options.cs
alexr-bq: This is to avoid name conflict with another SetCondition class we have.

sources/Valkey.Glide/ServerModules/GlideJson.ArrayCommands.cs
```diff
@@ -0,0 +1,224 @@
...
+    /// </returns>
+    /// <seealso href="https://valkey.io/commands/json.arrappend/"/>
+    /// <example>
+    /// <code>
+    /// await GlideJson.SetAsync(client, "mykey", "$", "{\"arr\":[1,2,3]}");
+    /// ValkeyResult result = await GlideJson.ArrAppendAsync(client, "mykey", "$.arr", "4", "5");
+    /// // (long[])result == [5] (new length of the array)
+    /// </code>
+    /// </example>
+    public static async Task<ValkeyResult> ArrAppendAsync(BaseClient client, ValkeyKey key, ValkeyValue path, params ValkeyValue[] values)
```
currantw: Should we use `IEnumerable<ValkeyValue>` instead for consistency? Same comment for other methods as well.

sources/Valkey.Glide/ServerModules/GlideJson.Options.cs
alexr-bq: Was for name conflict resolution with SetCondition but shouldn't actually be needed because of the nesting here, reverted

sources/Valkey.Glide/ServerModules/GlideJson.Options.cs
alexr-bq: Leaving to null -> don't send an end, which seems to be the proper behaviour

sources/Valkey.Glide/ServerModules/GlideJson.Options.cs
alexr-bq: Changing name

sources/Valkey.Glide/ServerModules/GlideJson.ArrayCommands.cs
currantw: Can this return `long?[]` instead?

sources/Valkey.Glide/ServerModules/GlideJson.ArrayCommands.cs
```diff
@@ -0,0 +1,224 @@
...
+    /// <param name="key">The key where the JSON document is stored.</param>
+    /// <param name="path">The JSONPath or legacy path within the JSON document.</param>
+    /// <param name="index">The index before which to insert the values. Negative indices count from the end.</param>
+    /// <param name="values">The JSON values to insert into the array.</param>
+    /// <returns>
+    /// When a JSONPath is provided, returns an array of new array lengths (or null for non-array matches).
+    /// When a legacy path is provided, returns the new array length.
+    /// </returns>
+    /// <seealso href="https://valkey.io/commands/json.arrinsert/"/>
+    public static async Task<ValkeyResult> ArrInsertAsync(BaseClient client, ValkeyKey key, ValkeyValue path, long index, params ValkeyValue[] values)
```
currantw: Can this return `long?[]` as well?

sources/Valkey.Glide/ServerModules/GlideJson.ArrayCommands.cs
```diff
@@ -0,0 +1,224 @@
...
+    /// </summary>
+    /// <param name="client">The Glide client to use for the command.</param>
+    /// <param name="key">The key where the JSON document is stored.</param>
+    /// <param name="path">The JSONPath or legacy path within the JSON document.</param>
+    /// <returns>
+    /// When a JSONPath is provided, returns an array of lengths (or null for non-array matches).
+    /// When a legacy path is provided, returns the array length.
+    /// </returns>
+    /// <seealso href="https://valkey.io/commands/json.arrlen/"/>
+    public static async Task<ValkeyResult> ArrLenAsync(BaseClient client, ValkeyKey key, ValkeyValue path)
```
currantw: Can this return `long?[]`?

sources/Valkey.Glide/ServerModules/GlideJson.ArrayCommands.cs
```diff
@@ -0,0 +1,224 @@
...
+    /// <param name="client">The Glide client to use for the command.</param>
+    /// <param name="key">The key where the JSON document is stored.</param>
+    /// <param name="path">The JSONPath or legacy path within the JSON document.</param>
+    /// <param name="index">The index of the element to pop. Negative indices count from the end. Default is -1 (last element).</param>
+    /// <returns>
+    /// When a JSONPath is provided, returns an array of popped elements (or null for non-array/empty matches).
+    /// When a legacy path is provided, returns the popped element.
+    /// </returns>
+    /// <seealso href="https://valkey.io/commands/json.arrpop/"/>
+    public static async Task<ValkeyResult> ArrPopAsync(BaseClient client, ValkeyKey key, ValkeyValue path, long index = -1)
```
currantw: Can this return `ValkeyValue[]`?

scripts/validate_examples.py
alexr-bq: One potential reason - naming conflict with very common System.Text.Json which might be used in conjunction with this?

But I'm not tied to it, if you don't think that's a very good reason I can switch GlideJson -> Json

Go and Node.js use GlideJson

sources/Valkey.Glide/ServerModules/GlideJson.ArrayCommands.cs
```diff
@@ -0,0 +1,224 @@
...
+    /// <param name="key">The key where the JSON document is stored.</param>
+    /// <param name="path">The JSONPath or legacy path within the JSON document.</param>
+    /// <param name="start">The start index (inclusive).</param>
+    /// <param name="end">The end index (inclusive). Negative indices count from the end.</param>
+    /// <returns>
+    /// When a JSONPath is provided, returns an array of new array lengths (or null for non-array matches).
+    /// When a legacy path is provided, returns the new array length.
+    /// </returns>
+    /// <seealso href="https://valkey.io/commands/json.arrtrim/"/>
+    public static async Task<ValkeyResult> ArrTrimAsync(BaseClient client, ValkeyKey key, ValkeyValue path, long start, long end)
```
currantw: Can this be `long?[]`?

sources/Valkey.Glide/ServerModules/GlideJson.Batch.cs
```diff
@@ -0,0 +1,627 @@
...
+/// Batch batch = new Batch(true);
+/// GlideJsonBatch.Set(batch, "doc", ".", "{\"a\": 1.0, \"b\": 2}");
+/// GlideJsonBatch.Get(batch, "doc");
+/// object?[]? result = await client.Exec(batch, false);
+/// // result[0] == "OK" (result of Set)
+/// // result[1] == "{\"a\": 1.0, \"b\": 2}" (result of Get)
+/// </code>
+/// </example>
+/// <seealso href="https://valkey.io/commands/?group=json">Valkey JSON Commands</seealso>
+public static class GlideJsonBatch
```
currantw, line 28: Rather than a separate `GlideJsonBatch` module, would it make sense to just put back and client in the same `GlideJson` static class? Than users could just call `GlideJson.Set(batch, ...)`. Seems like a more unified approach? 🤷

sources/Valkey.Glide/ServerModules/GlideJson.ArrayCommands.cs
alexr-bq: Yeah, I'm switching these so when it returns single values it will wrap in an array. Let me know if that makes sense to you.

sources/Valkey.Glide/ServerModules/GlideJson.cs
```diff
@@ -0,0 +1,524 @@
...
+    {
+        if (result is null)
+            return ValkeyValue.Null;
+        if (result is GlideString gs)
+            return gs;
+        return result.ToString() ?? string.Empty;
```
currantw: Should we throw an exception here? Or explicitly handle other types? I think only `double`, `long`, and maybe `bool` could be returned by the server?

sources/Valkey.Glide/ServerModules/GlideJson.cs
```diff
@@ -0,0 +1,524 @@
...
+    /// <param name="value">The JSON value to set (must be valid JSON-encoded string).</param>
+    /// <returns>"OK" if the value was set successfully, or <see cref="ValkeyValue.Null"/> if the condition was not met.</returns>
+    /// <seealso href="https://valkey.io/commands/json.set/"/>
+    /// <example>
+    /// <code>
+    /// ValkeyValue result = await GlideJson.SetAsync(client, "mykey", "$", "{\"name\":\"John\"}");
+    /// // result == "OK"
+    /// </code>
+    /// </example>
+    public static Task<ValkeyValue> SetAsync(BaseClient client, ValkeyKey key, ValkeyValue path, ValkeyValue value)
```
currantw: This can just return `Task`, right?

Valkey docs say it will return:

> [Simple string reply](https://valkey.io/topics/protocol#simple-strings): 'OK' on success.
> [Nil reply](https://valkey.io/topics/protocol#bulk-strings): If the NX or XX condition is not met.

So is "or <see cref="ValkeyValue.Null"/> if the condition was not met." accurate?

sources/Valkey.Glide/ServerModules/GlideJson.cs
```diff
@@ -0,0 +1,541 @@
...
+
+    /// <summary>
+    /// Gets the JSON value(s) at the specified path(s) in the key.
+    /// </summary>
+    /// <param name="client">The Glide client to use for the command.</param>
+    /// <param name="key">The key where the JSON document is stored.</param>
+    /// <param name="paths">The JSONPath or legacy path(s) within the JSON document.</param>
+    /// <returns>The JSON value(s), or <see cref="ValkeyValue.Null"/> if the key does not exist.</returns>
+    /// <seealso href="https://valkey.io/commands/json.get/"/>
+    public static async Task<ValkeyValue> GetAsync(BaseClient client, ValkeyKey key, ValkeyValue[] paths)
```
currantw: IEnumerable?


```suggestion
    public static async Task<ValkeyValue> GetAsync(BaseClient client, ValkeyKey key, IEnumerable<ValkeyValue> paths)
```

sources/Valkey.Glide/ServerModules/GlideJson.cs
```diff
@@ -0,0 +1,541 @@
...
+    /// <param name="client">The Glide client to use for the command.</param>
+    /// <param name="key">The key where the JSON document is stored.</param>
+    /// <param name="path">The JSONPath or legacy path within the JSON document.</param>
+    /// <returns>The number of values cleared.</returns>
+    /// <seealso href="https://valkey.io/commands/json.clear/"/>
+    public static async Task<long> ClearAsync(BaseClient client, ValkeyKey key, ValkeyValue path)
+    {
+        GlideString[] args = [JsonClear, ToGlideString(key), ToGlideString(path)];
+        object? result = await ExecuteCommandAsync(client, args);
+        return (long)(result ?? 0L);
```
currantw: Under what circumstances would `result` be `null` here?

sources/Valkey.Glide/ServerModules/GlideJson.cs
```diff
@@ -0,0 +1,541 @@
...
+    /// </summary>
+    /// <param name="client">The Glide client to use for the command.</param>
+    /// <param name="key">The key where the JSON document is stored.</param>
+    /// <param name="path">The JSONPath or legacy path within the JSON document.</param>
+    /// <returns>
+    /// An array of type values for each matching path. Returns null for non-existent paths.
+    /// Types: "null", "boolean", "string", "number", "integer", "object", "array".
+    /// </returns>
+    /// <seealso href="https://valkey.io/commands/json.type/"/>
+    public static async Task<ValkeyValue[]?> TypeAsync(BaseClient client, ValkeyKey key, ValkeyValue path)
```
currantw: - Should this be `ValkeyValue[]` instead? Valkey docs say it uses "Empty array if the document key does not exist."
- Or, since there is a finite number of values ('nil', 'boolean', 'string', 'number', 'integer', 'object', 'array', according to docs), should we add an Enum to return here? So return type could be `Json.Type?[]` perhaps?

sources/Valkey.Glide/ServerModules/GlideJson.cs
```diff
@@ -0,0 +1,541 @@
...
+    /// <param name="client">The Glide client to use for the command.</param>
+    /// <param name="key">The key where the JSON document is stored.</param>
+    /// <param name="path">The JSONPath or legacy path within the JSON document.</param>
+    /// <param name="increment">The amount to increment by.</param>
+    /// <returns>
+    /// When a JSONPath is provided, returns a JSON string array of new values (or "null" for non-numeric matches).
+    /// When a legacy path is provided, returns the new value as a string.
+    /// </returns>
+    /// <seealso href="https://valkey.io/commands/json.numincrby/"/>
+    public static async Task<ValkeyValue> NumIncrByAsync(BaseClient client, ValkeyKey key, ValkeyValue path, double increment)
```
currantw, line 548: Should we parse this to `double?[]`?

sources/Valkey.Glide/ServerModules/GlideJson.cs
```diff
@@ -0,0 +1,541 @@
...
+    /// <param name="client">The Glide client to use for the command.</param>
+    /// <param name="key">The key where the JSON document is stored.</param>
+    /// <param name="path">The JSONPath or legacy path within the JSON document.</param>
+    /// <param name="multiplier">The amount to multiply by.</param>
+    /// <returns>
+    /// When a JSONPath is provided, returns a JSON string array of new values (or "null" for non-numeric matches).
+    /// When a legacy path is provided, returns the new value as a string.
+    /// </returns>
+    /// <seealso href="https://valkey.io/commands/json.nummultby/"/>
+    public static async Task<ValkeyValue> NumMultByAsync(BaseClient client, ValkeyKey key, ValkeyValue path, double multiplier)
```
currantw, line 579: Similarly, should this return `double?[]`?

sources/Valkey.Glide/ServerModules/GlideJson.cs
```diff
@@ -0,0 +1,541 @@
...
+    /// </summary>
+    /// <param name="client">The Glide client to use for the command.</param>
+    /// <param name="key">The key where the JSON document is stored.</param>
+    /// <param name="path">The JSONPath or legacy path within the JSON document.</param>
+    /// <param name="value">The JSON string value to append (must be a valid JSON string, e.g., "\"suffix\"").</param>
+    /// <returns>
+    /// An array of new string lengths for each matching path. Returns null for non-string matches.
+    /// </returns>
+    /// <seealso href="https://valkey.io/commands/json.strappend/"/>
+    public static async Task<long?[]?> StrAppendAsync(BaseClient client, ValkeyKey key, ValkeyValue path, ValkeyValue value)
```
currantw: What is the difference here between returning a null array and an empty array? Do we need both?

sources/Valkey.Glide/ServerModules/GlideJson.cs
```diff
@@ -0,0 +1,541 @@
...
+    /// <param name="client">The Glide client to use for the command.</param>
+    /// <param name="key">The key where the JSON document is stored.</param>
+    /// <param name="value">The JSON string value to append (must be a valid JSON string, e.g., "\"suffix\"").</param>
+    /// <returns>The new string length.</returns>
+    /// <seealso href="https://valkey.io/commands/json.strappend/"/>
+    public static async Task<long> StrAppendAsync(BaseClient client, ValkeyKey key, ValkeyValue value)
+    {
+        GlideString[] args = [JsonStrAppend, ToGlideString(key), ToGlideString(value)];
+        object? result = await ExecuteCommandAsync(client, args);
+        return (long)(result ?? 0L);
```
currantw: Under what conditions would this be `null`?

sources/Valkey.Glide/ServerModules/GlideJson.cs
```diff
@@ -0,0 +1,541 @@
...
+    /// Gets the length of the string value at the specified path in the key.
+    /// </summary>
+    /// <param name="client">The Glide client to use for the command.</param>
+    /// <param name="key">The key where the JSON document is stored.</param>
+    /// <param name="path">The JSONPath or legacy path within the JSON document.</param>
+    /// <returns>
+    /// An array of string lengths for each matching path. Returns null for non-string matches.
+    /// </returns>
+    /// <seealso href="https://valkey.io/commands/json.strlen/"/>
+    public static async Task<long?[]?> StrLenAsync(BaseClient client, ValkeyKey key, ValkeyValue path)
```
currantw, line 693: Similar to previous comment, could this just be `long?[]`? What does the `null` case add?

sources/Valkey.Glide/ServerModules/GlideJson.cs
alexr-bq: There's an argument for it but I've confirmed that this implementation is consistent with all other glide clients.

sources/Valkey.Glide/ServerModules/GlideJson.cs
alexr-bq: There's an argument for it but I've confirmed that this implementation is consistent with all other glide clients.

sources/Valkey.Glide/ServerModules/GlideJson.Batch.cs
alexr-bq: I can see arguments either way (not polluting main class with batch functions) but this way is more consistent with other glide clients.

scripts/validate_examples.py
currantw: > One potential reason - naming conflict with very common System.Text.Json which might be used in conjunction with this?

Hmmm. That seems like a decent reason.

But it's easy enough to add an alias in C# (it seems to me):

```
// Alias a type
using GlideJson = Valkey.Glide.ServerModules.Json;

// Now use it as:
await GlideJson.SetAsync(client, "key", "$", "{}");
```

So it seems to me like it would be better to have this nice `Json.*` syntax _and_ be consistent with `Ft.*`, so that would be my preference. But don't feel strongly enough about it to really push it too much 😉

sources/Valkey.Glide/ServerModules/GlideJson.ObjectCommands.cs
```diff
@@ -0,0 +1,327 @@
...
+        if (result is null)
+            return null;
+        if (result is object?[] arr)
+            return [.. arr.Select(o => o?.ToString() ?? string.Empty)];
+        return [];
+    }
+
+    #endregion
+
+    #region JSON.TOGGLE
```
currantw: Maybe makes more sense for this to be in `GlideJson.cs`?

sources/Valkey.Glide/ServerModules/GlideJson.cs
```diff
@@ -0,0 +1,690 @@
...
+    /// <remarks>
+    /// <example>
+    /// <code>
+    /// await GlideJson.SetAsync(client, "key1", "$", "{\"a\":1}");
+    /// await GlideJson.SetAsync(client, "key2", "$", "{\"a\":2}");
+    /// var values = await GlideJson.MGetAsync(client, ["key1", "key2"], "$.a");  // ["[1]", "[2]"]
+    /// </code>
+    /// </example>
+    /// </remarks>
+    public static async Task<ValkeyValue[]> MGetAsync(BaseClient client, IEnumerable<ValkeyKey> keys, ValkeyValue path)
```
currantw: - For consistency with SER, the general pattern in the GLIDE client has been to just call this `GetAsync` (no "M") and just have it as a different variant.
- I also noticed there didn't seem to be a MSetAsync – was that intentional?

sources/Valkey.Glide/ServerModules/GlideJson.ObjectCommands.cs
```diff
@@ -0,0 +1,327 @@
...
+    /// </example>
+    /// </remarks>
+    public static async Task<ValkeyValue[]?[]?> ObjKeysAsync(BaseClient client, ValkeyKey key, ValkeyValue path)
+    {
+        GlideString[] args = [JsonObjKeys, ToGlideString(key), ToGlideString(path)];
+        object? result = await ExecuteCommandAsync(client, args);
+        return ConvertToNestedValkeyValueArray(result);
+    }
+
+    private static ValkeyValue[]?[]? ConvertToNestedValkeyValueArray(object? result)
```
currantw: Can this just return `ValkeyValue[][]`? 🤔

sources/Valkey.Glide/ServerModules/GlideJson.ObjectCommands.cs
```diff
@@ -0,0 +1,327 @@
...
+    public static async Task<long?> DebugMemoryAsync(BaseClient client, ValkeyKey key)
+    {
+        GlideString[] args = [JsonDebug, ValkeyLiterals.MEMORY, ToGlideString(key)];
+        object? result = await ExecuteCommandAsync(client, args);
+        return result is null ? null : (long)result;
+    }
+
+    #endregion
+
+    #region JSON.DEBUG FIELDS
```
currantw, line 203: I don't think we really have many other debug commands implemented right now? Do we need these? Are they implemented in other clients?

sources/Valkey.Glide/ServerModules/GlideJson.cs
alexr-bq: Changing to an exception

sources/Valkey.Glide/ServerModules/GlideJson.cs
alexr-bq: Outer ? (nullable array): null when the key doesn't exist
Inner ? (nullable elements): null for paths where the value isn't the expected type (e.g., not a string for STRLEN)

Updating documentation to match.

alexr-bq: This applies to read commands, write commands throw exceptions such that this type was not appropriate and has been changed.

sources/Valkey.Glide/ServerModules/GlideJson.cs
alexr-bq: Fixing first point

No other language has json mset. I'm not sure why, leaving it out for consistency for now, will ask around and perhaps raise an issue.

sources/Valkey.Glide/ServerModules/GlideJson.ObjectCommands.cs
alexr-bq: The outer nullable is necessary but the inner isn't. Fixing.

sources/Valkey.Glide/ServerModules/GlideJson.ObjectCommands.cs
alexr-bq: Yes, other clients have them.

yipin-chen (approved):

Issue #123, opened 2025-10-28
[C#][Flaky Test] ConfigGetAsync_ReturnsConfigurationPerNode


xShinnRyuu:
### Test Name

Valkey.Glide.IntegrationTests.ClusterClientTests.ConfigGetAsync_ReturnsConfigurationPerNode(client: GlideClusterClient { 0x600003490F60 RESP2 })

### Test Location

alkey-glide-csharp/valkey-glide-csharp/tests/Valkey.Glide.IntegrationTests/ClusterClientTests.cs:line 266

### Failure Permlink

https://github.com/valkey-io/valkey-glide-csharp/actions/runs/18862990152/job/53825160585#step:7:2552

### System Information

aarch64-apple-darwin

### Language and Version

C# net8.0

### Engine Version

v6.2

### Logs

```
Failed Valkey.Glide.IntegrationTests.ClusterClientTests.ConfigGetAsync_ReturnsConfigurationPerNode(client: GlideClusterClient { 0x600003490F60 RESP2 }) [79 ms]
  Error Message:
   Assert.Contains() Failure: Filter not matched in collection
Collection: [["maxmemory-policy"] = "allkeys-lru"]
  Stack Trace:
     at Valkey.Glide.IntegrationTests.ClusterClientTests.ConfigGetAsync_ReturnsConfigurationPerNode(GlideClusterClient client) in /Users/runner/work/valkey-glide-csharp/valkey-glide-csharp/tests/Valkey.Glide.IntegrationTests/ClusterClientTests.cs:line 313
   at Valkey.Glide.IntegrationTests.ClusterClientTests.ConfigGetAsync_ReturnsConfigurationPerNode(GlideClusterClient client) in /Users/runner/work/valkey-glide-csharp/valkey-glide-csharp/tests/Valkey.Glide.IntegrationTests/ClusterClientTests.cs:line 329
--- End of stack trace from previous location ---
 ```

### Glide Version

main

PR #376, opened 2026-05-07
Refactor existing CI/CD workflow

### LINKED ISSUE #345: CI/CD: Extract reusable test workflow and introduce nightly Full Matrix Test

### LINKED ISSUE #382: Add Redis 7.1 and Redis 7.2 to full CI test matrix
## Summary

The [official Valkey GLIDE documentation](https://glide.valkey.io/overview/#supported-engine-versions) lists Redis 6.2, 7.0, 7.1, and 7.2 as supported engine versions. However, the CI test matrix in does not cover Redis 7.1 and Redis 7.2.

## Current state

```json
[
    { "type": "redis", "version": "7.0", "profiles": ["full"] },
    { "type": "redis", "version": "6.2", "profiles": ["standard", "full"] }
]
```

## Proposed change

Add Redis 7.1 and Redis 7.2 to the `full` profile in `server-matrix.json`:

```json
{ "type": "redis", "version": "7.1", "profiles": ["full"] },
{ "type": "redis", "version": "7.2", "profiles": ["full"] }
```

## Motivation

Ensuring CI coverage matches documented supported versions reduces the risk of shipping regressions against Redis 7.1 and 7.2 deployments.

Aryex:
### Summary

This PR refactor our existing ci/cd pipelines to use the resuable test.yml workflows.

### Issue Link

Closes #345 
Closes #382 

- Renamed `_reusable-test.yml` to `test.yml`. 
- Replaced `tests.yml` with `ci.yml` which runs on each PR.
- Refactored `cd.yml` to use our reusable workflow, running the full test profile.
- Cleanup matrix configurations from dead properties.
- Added missing redis tests to full matrix environment.
- Added modules tests to test.yml. It only runs in full profile.

### Checklist

## Summary by CodeRabbit

* **Chores**
  * Removed the smoke test profile; CI now uses only standard and full profiles
  * Simplified test matrices and stopped emitting per-profile test-filter outputs
  * Renamed/streamlined CD workflow, added a new CI workflow, and consolidated test execution to a reusable test workflow; retired the older standalone workflow

* **Tests**
  * Added a dedicated module integration test job with artifact publishing
  * Improved TRX result parsing for more robust failure reporting


.github/workflows/tests.yml
```diff
@@ -1,354 +0,0 @@
...
-              uses: actions/upload-artifact@v7
-              with:
-                  name: test-reports-dotnet-${{ matrix.dotnet }}-${{ matrix.server.type }}-${{ matrix.server.version }}-${{ matrix.host.OS }}-${{ matrix.host.ARCH }}
-                  path: |
-                      TestReport.html
-                      TestResults.trx
-                      failures.json
-                      valkey-glide/utils/clusters/**
-
-    tests-modules:
```
currantw, line 159: This block is needed for run module tests (JSON and FT/search). These run off Valkey servers that are spun up through Docker so they can be loaded with JSON and FT modules automatically. Are these preserved through this refacactor?

(As was worried when these were added that have a "parallel" set of tests might create issues for us. Seems like that might have arrived more quickly than I anticipated!)

I think fixing more holistically this is ultimately going to be a bigger issue. In my estimation, we probably need to first update `cluster_manager.py` so that it can spin up servers with modules enabled, and then update all our developer documentation and GitHub workflows so that they can test modules using the same logic/code paths as other tests.

But this is a bigger issue than this PR! For now, can you just make sure that these modules tests continue to run across all profiles, and then perhaps raise issue(s) for a separate refactor?

.github/workflows/ci.yml
currantw, line 1: The old workflow has a `cancel-in-progress: true`. I think we should preserve this so that older workflows are automatically cancelled if new changes are pushed.

.github/workflows/cd.yml
```diff
@@ -208,93 +209,17 @@ jobs:
...
-              uses: actions/upload-artifact@v7
-              with:
-                  name: test-reports-${{ matrix.host.TARGET }}
-                  path: |
-                      TestReport.html
-                      valkey-glide/utils/clusters/**
+        needs: [publish, set-release-version]
+        uses: ./.github/workflows/_reusable-test.yml
+        with:
+            profile: smoke
```
currantw: Why did we decide to go with `smoke` for CD? Given how rarely this runs and how important it is, why not at least run `standard`?

.github/workflows/cd.yml
```diff
@@ -41,8 +41,9 @@ jobs:
             - name: load-platform-matrix
               id: load-platform-matrix
               working-directory: .github/json_matrices
+              # Select all VM runners (non-container) — each produces a unique native binary for the NuGet package.
               run: |
-                  jq -c "[.[] | select(any(.tags[] == \"cd\"; .))]" < os-matrix.json | awk '{ printf "PLATFORM_MATRIX=%s\n", $0 }' | tee -a $GITHUB_OUTPUT
+                  jq -c '[.[] | select(has("IMAGE") | not)]' < os-matrix.json | awk '{ printf "PLATFORM_MATRIX=%s\n", $0 }' | tee -a $GITHUB_OUTPUT
```
currantw: Nit. Even with the explanatory comment, I had a hard figuring out what was going on here (certainly no expert in bash. Maybe an intermediary variable would help? 🤷 

```bash
PLATFORM_MATRIX=$(jq -c '[.[] | select(has("IMAGE") | not)]' < os-matrix.json)
echo "PLATFORM_MATRIX=$PLATFORM_MATRIX" | tee -a "$GITHUB_OUTPUT"
```

.github/workflows/cd.yml
```diff
@@ -208,93 +209,17 @@ jobs:
...
-            - name: Upload test reports
-              if: always()
-              continue-on-error: true
-              uses: actions/upload-artifact@v7
-              with:
-                  name: test-reports-${{ matrix.host.TARGET }}
-                  path: |
-                      TestReport.html
-                      valkey-glide/utils/clusters/**
+        needs: [publish, set-release-version]
```
currantw, line 220: Was it intentional to remove `load-platform-matrix` from this list?

.github/json_matrices/os-matrix.json
currantw: It also looks like `NAMED_OS` is now no longer used anywhere. Might as well remove?

.github/json_matrices/os-matrix.json
```diff
@@ -5,7 +5,6 @@
         "RUNNER": "ubuntu-24.04",
         "ARCH": "x64",
         "TARGET": "x86_64-unknown-linux-gnu",
-        "tags": ["ci", "cd"],
         "profiles": ["smoke", "standard", "full"]
```
currantw, line 9: Even though I'm broadly familiar with them, these names are still a bit confusing for me. Perhaps while we are already refactoring, we could consider some renames for clarity:

- `RUNNER` to `GITHUB_RUNNER`
- `TARGET` to `RUST_TARGET`

.github/workflows/cd.yml
currantw, line 1: Broader question about this refactoring:
- Have we actually verified that the coverage is output ("Run coverage)?
- What about the test reports – where can these actually be found ("Upload test reports")?
- Is this documented anywhere that I'm missing – should it be?
- Does it make sense for the coverage and test report steps to always run (at least every .NET 8.0, Ubuntu, x64 test with Valkey 9 in the former case)?

currantw: Seems like we run all the tests over again for coverage! That doesn't make sense to me, and it certainly slows our jobs down! Would you be able to raise an issue for this?

.github/workflows/ci.yml
xShinnRyuu: +1

.github/workflows/cd.yml
xShinnRyuu: +1

xShinnRyuu (commented):
Other than what Taylor has pointed out i'm okay with this PR.

.github/json_matrices/os-matrix.json
Aryex: Removed

.github/workflows/cd.yml
Aryex: The tetst result should be uploaded to the github action's summary page (scroll to the bottom).

However, as disscussed we will need to review the coverage so I'm going to disable it for now.

.github/workflows/cd.yml
Aryex: It does not require it. _reusable-test.yml loads its own platform matrix.

.github/workflows/ci.yml
Aryex: fixed

.github/workflows/cd.yml
Aryex: fixed

.github/workflows/cd.yml
Aryex: updated to standard

.github/workflows/tests.yml
Aryex: Added back. It will now run on standard and full

.github/json_matrices/os-matrix.json
Aryex: I would refrain from this for now. It touches quite a few files.

currantw: Sure. I might just create a PR for this after it's merged then, if I'm doing CI/CD work anyway. 🤷

.github/workflows/tests.yml
currantw: Hmmm. I don't like that are aren't running them as part of smoke tests right now. These are significant command groups that we now support, so skipping them in smoke tests seems like completed skipping all the sorted set command tests or all the geospatial command tests. 😬 

That being said, I think it is reasonable given where this is at...

Can you raise an issue to integrate the modules tests with the regular test infrastructure. and tag it here with an explanatory "TODO #123" comment? As mentioned, I think that issue will require a decent bit of design as well.

Thoughts?

.github/workflows/ci.yml
currantw: Thanks! 🙏

.github/workflows/cd.yml
currantw: Thank you! 🎉

.github/workflows/cd.yml
currantw: Thanks for supporting bash amateurs 😉

.github/workflows/cd.yml
currantw: Thanks!

Only slightly related but noticed this. It seems like the `load-platform-matrix` job checks out _the entire repo_, and then only uses it to load the PLATFORM_MATRIX env var from `os-matrix.json` – which is now only used by the  `create-binaries` job.

Are we able to update `load-platform-matrix` to use `sparse-checkout`, so it only fetches `os-matrix.json` file it needs?

```yml
- uses: actions/checkout@v4
  with:
    sparse-checkout: .github/json_matrices
    sparse-checkout-cone-mode: false
```

.github/json_matrices/os-matrix.json
currantw: Thanks! 🧹

.github/workflows/cd.yml
currantw: Thank you 🙏

.github/workflows/tests.yml
currantw, line 1: Can we remove the `create-test-matrices` action as well now? Seems like it is only used by `tests` workflow (which is now removed)?

currantw: Removed in my PR, so comment is already outdated now.

.github/workflows/_reusable-test.yml
```diff
@@ -303,3 +299,77 @@ jobs:
...
+            - uses: actions/checkout@v6
+              with:
+                  submodules: true
+
+            - name: Set up dotnet ${{ matrix.dotnet }}
+              uses: actions/setup-dotnet@v5
+              with:
+                  dotnet-version: |
+                      9
+                      ${{ matrix.dotnet }}
```
currantw, line 318: Do we need to install both dotnet versions here, or would just one of them (matrix.dotnet, I assume) suffice?

.github/workflows/cd.yml
Aryex: added

.github/workflows/tests.yml
Aryex: Added https://github.com/valkey-io/valkey-glide-csharp/issues/390

.github/workflows/_reusable-test.yml
currantw: Thanks for removing.

currantw (approved):
Looks good. Thanks! ✅

xShinnRyuu (approved):

PR #392, opened 2026-05-12
Support custom socket address resolution

affonsov:
### Summary

Ports the custom socket address resolution feature from valkey-glide to the C# client. Allows users to provide a callback that intercepts and rewrites server host/port pairs before the client establishes connections - useful for routing through proxies or custom DNS resolution.

### Issue Link

This pull request is linked to issue: N/A (porting from upstream valkey-glide)

- New public delegate `AddressResolverDelegate` in `ConnectionConfiguration`:
  ```csharp
  public delegate (string host, ushort port) AddressResolverDelegate(string host, ushort port);
  ```
- New `WithAddressResolver()` builder method on both `StandaloneClientConfigurationBuilder` and `ClusterClientConfigurationBuilder`
- Feature is fully opt-in - `null` by default, no behaviour change for existing clients
- If the resolver throws an exception or returns invalid values (empty host, port 0, oversized hostname), the client falls back to the original address and logs an error (consistent with Java client behavior)

### Implementation

Follows the direct FFI callback pattern (same as Go PR valkey-io/valkey-glide#5891), not the socket listener/registry pattern used by Python and Node.

- Updated `valkey-glide` submodule to `b4b34a1dc16d` to include the core Rust `AddressResolver` trait (PR #5328) and the FFI layer changes (PR #5876)
- Added `AddressResolverCallback` C function pointer type and `FFIAddressResolver` struct to `rust/src/lib.rs`
- Updated `create_client` Rust FFI function with an optional 5th `address_resolver` parameter
- C# side marshals the managed delegate to a native function pointer and passes it to the Rust layer; the delegate is held as a field to prevent GC collection

### Limitations

- The resolver is called once per address at connection time; it is not called on reconnection attempts (same behaviour as other language bindings)
- Maximum resolved hostname length is 1024 bytes

### Testing

- Unit tests in `ConnectionConfigurationTests.cs` covering builder configuration, defaults, and edge cases
- Integration tests in `AddressResolverTests.cs` covering:
  - Identity resolver is invoked and client connects successfully (standalone and cluster)
  - Resolver that throws an exception falls back to original address
  - Resolver returning empty host falls back to original address
  - Resolver returning port zero falls back to original address
  - Resolver returning oversized hostname falls back to original address

### Related Issues

- valkey-io/valkey-glide#5328 - Java: Support custom socket address resolution (core Rust infra)
- valkey-io/valkey-glide#5876 - Python: Support custom socket address resolution (FFI layer)
- valkey-io/valkey-glide#5890 - Node: Support custom socket address resolution
- valkey-io/valkey-glide#5891 - Go: Support custom socket address resolution

### Checklist

- [x] ~~This Pull Request is related to one issue.~~


tests/Valkey.Glide.UnitTests/AddressResolverTests.cs
currantw, line 1: There is an existing `ConnectionConfigurationTests` that contains tests for the various builder properties and methods. Probably best to just move this to a new "Address Resolver Tests" region in there instead of a new file?

rust/src/lib.rs
```diff
@@ -97,6 +97,55 @@ pub type FailureCallback = unsafe extern "C-unwind" fn(
     error_type: RequestErrorType,
 ) -> ();
 
+/// Callback for resolving server addresses before connection.
+/// Returns the resolved port (0 on error), and writes the resolved host into `resolved_host_buf`.
+/// `resolved_host_len` is set to the number of bytes written.
```
currantw: Maybe worth specifying what happens to the host on error?

sources/Valkey.Glide/ConnectionConfiguration.cs
```diff
@@ -877,6 +887,27 @@ public T WithClientSideCache(ClientSideCacheConfig clientSideCacheConfig)
...
+            get => Config.AddressResolver;
+            set => Config.AddressResolver = value;
+        }
+
+        /// <inheritdoc cref="AddressResolver" />
+        public T WithAddressResolver(AddressResolverDelegate addressResolver)
+        {
+            AddressResolver = addressResolver;
+            return (T)this;
+        }
```
currantw, line 915: For other options like `WithClientSideCache`, we don't support "unsetting" the value to null, or providing explicit `null` option – we call `ArgumentNullException.ThrowIfNull`. Should we do the same here for consistency?

rust/src/lib.rs
```diff
@@ -97,6 +97,55 @@ pub type FailureCallback = unsafe extern "C-unwind" fn(
...
+
+impl std::fmt::Debug for FFIAddressResolver {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "FFIAddressResolver")
+    }
+}
+
+impl redis::AddressResolver for FFIAddressResolver {
+    fn resolve(&self, host: &str, port: u16) -> (String, u16) {
+        let mut buf = vec![0u8; 1024];
```
currantw: Probably a good idea to extract a constant for the max hostname size?

 ```rust
 const MAX_RESOLVED_HOST_LEN: usize = 1024;
 ```

rust/src/lib.rs
```diff
@@ -97,6 +97,55 @@ pub type FailureCallback = unsafe extern "C-unwind" fn(
...
+                &mut len,
+            )
+        };
+        if resolved_port == 0 || len == 0 || len > buf.len() {
+            return (host.to_string(), port);
+        }
+        match std::str::from_utf8(&buf[..len]) {
+            Ok(h) => (h.to_string(), resolved_port),
+            Err(_) => (host.to_string(), port),
+        }
```
currantw, line 171: Seems like we should _at least_ log some errors here – not just silently pass back the original post and host? An error here would be very difficult otherwise for a user to debug. I would personally prefer raising an error. What do the other clients do?

sources/Valkey.Glide/ConnectionConfiguration.cs
```diff
@@ -20,6 +20,15 @@ public abstract class ConnectionConfiguration
...
     public static readonly long CertificateMaxSize = 10 * 1024 * 1024; // 10 MB
 
+    /// <summary>
+    /// Callback for resolving server addresses before connection.
+    /// Allows intercepting and rewriting host/port pairs before the client connects.
+    /// </summary>
+    /// <param name="host">The configured host.</param>
+    /// <param name="port">The configured port.</param>
+    /// <returns>The resolved (host, port) to use for the actual connection.</returns>
+    public delegate (string host, int port) AddressResolverDelegate(string host, int port);
```
currantw: Should this instead be `ushort` for port? Otherwise, we probably need to check for and handle possible overflow errors before casting to `ushort` and passing to the FFI layer.

sources/Valkey.Glide/Client/BaseClient.cs
```diff
@@ -201,7 +201,27 @@ protected static async Task<T> CreateClient<T>(BaseClientConfiguration config, F
         }
 
         Message message = client.MessageContainer.GetMessageForCall();
-        CreateClientFfi(request.ToPtr(), successCallbackPointer, failureCallbackPointer, pubsubCallbackPointer);
+
+        IntPtr addressResolverPointer = IntPtr.Zero;
+        if (config.Request.AddressResolver != null)
+        {
+            client._addressResolverDelegate = (hostPtr, hostLen, port, bufPtr, bufLen, outLen) =>
```
currantw, line 210: Suggestion from Kiro (one that seems reasonable to me):

**Exception handling in the callback**: If `config.Request.AddressResolver` throws an exception, it will propagate through the FFI boundary into Rust, which is undefined behavior. The callback should wrap the user delegate in a try/catch:

 ```csharp
 client._addressResolverDelegate = (hostPtr, hostLen, port, bufPtr, bufLen, outLen) =>
 {
     try
     {
         string host = Marshal.PtrToStringUTF8(hostPtr, (int)hostLen) ?? string.Empty;
         var (resolvedHost, resolvedPort) = config.Request.AddressResolver(host, port);
         // ... rest of logic
     }
     catch
     {
         return 0; // Signal error to Rust, which will fall back to original address
     }
 };
 ```

tests/Valkey.Glide.UnitTests/AddressResolverTests.cs
currantw, line 1: I think we also need integration tests for this? Should include successful connection with address resolution and various error scenarios.

sources/Valkey.Glide/Client/BaseClient.cs
```diff
@@ -201,7 +201,27 @@ protected static async Task<T> CreateClient<T>(BaseClientConfiguration config, F
...
+                string host = Marshal.PtrToStringUTF8(hostPtr, (int)hostLen) ?? string.Empty;
+                var (resolvedHost, resolvedPort) = config.Request.AddressResolver(host, port);
+                if (resolvedPort <= 0 || string.IsNullOrEmpty(resolvedHost))
+                    return 0;
+                byte[] bytes = System.Text.Encoding.UTF8.GetBytes(resolvedHost);
+                if (bytes.Length > (int)bufLen)
+                    return 0;
+                Marshal.Copy(bytes, 0, bufPtr, bytes.Length);
+                unsafe { *(UIntPtr*)outLen = (UIntPtr)bytes.Length; }
+                return (ushort)resolvedPort;
```
currantw: Same delegate issue mentioned in elsewhere, I think. Should this already be `ushort` so we can avoid any possible errors from casting from `int`?

sources/Valkey.Glide/Client/BaseClient.cs
```diff
@@ -201,7 +201,27 @@ protected static async Task<T> CreateClient<T>(BaseClientConfiguration config, F
...
         Message message = client.MessageContainer.GetMessageForCall();
-        CreateClientFfi(request.ToPtr(), successCallbackPointer, failureCallbackPointer, pubsubCallbackPointer);
+
+        IntPtr addressResolverPointer = IntPtr.Zero;
+        if (config.Request.AddressResolver != null)
+        {
+            client._addressResolverDelegate = (hostPtr, hostLen, port, bufPtr, bufLen, outLen) =>
+            {
+                string host = Marshal.PtrToStringUTF8(hostPtr, (int)hostLen) ?? string.Empty;
+                var (resolvedHost, resolvedPort) = config.Request.AddressResolver(host, port);
```
currantw: Might be worth extracting an `addressResolver` variable so that the delegate doesn't need to capture (and prevent GC of) the entire `config.Request`. 🤷

sources/Valkey.Glide/Client/BaseClient.cs
```diff
@@ -201,7 +201,27 @@ protected static async Task<T> CreateClient<T>(BaseClientConfiguration config, F
...
 
         Message message = client.MessageContainer.GetMessageForCall();
-        CreateClientFfi(request.ToPtr(), successCallbackPointer, failureCallbackPointer, pubsubCallbackPointer);
+
+        IntPtr addressResolverPointer = IntPtr.Zero;
+        if (config.Request.AddressResolver != null)
+        {
+            client._addressResolverDelegate = (hostPtr, hostLen, port, bufPtr, bufLen, outLen) =>
+            {
+                string host = Marshal.PtrToStringUTF8(hostPtr, (int)hostLen) ?? string.Empty;
```
currantw: Under what conditions might `PtrToStringUTF8` return null? Should we raise en error here rather than silently defaulting to the empty string (which might lead to further errors later?)

tests/Valkey.Glide.UnitTests/AddressResolverTests.cs
currantw: Fixed! ✅ Moved to a new `#region Address Resolver Tests` in `ConnectionConfigurationTests.cs`.

rust/src/lib.rs
currantw: Fixed! ✅ Added doc comment on the `resolve` method documenting fallback to original `(host, port)` on error.

sources/Valkey.Glide/ConnectionConfiguration.cs
currantw: Not applicable. Java allows null (`Optional<AddressResolver>`) and the property setter should remain permissive for consistency with other GLIDE clients. The `WithAddressResolver` builder method is the recommended API.

rust/src/lib.rs
currantw: Fixed! ✅ Extracted `const MAX_RESOLVED_HOST_LEN: usize = 1024;` and also switched from heap allocation (`vec!`) to stack allocation (`[0u8; MAX_RESOLVED_HOST_LEN]`).

rust/src/lib.rs
currantw: Fixed! ✅ Added error logging on the C# side for each failure case (empty host, oversized host, invalid port, exception). Checked Java — it also falls back to the original address on error (does not raise). The Rust `AddressResolver` trait returns `(String, u16)` with no `Result`, so errors can't propagate — fallback is the intended cross-client behavior.

sources/Valkey.Glide/ConnectionConfiguration.cs
currantw: Fixed! ✅ Changed delegate to `ushort` throughout — overflow is now impossible at the type level.

sources/Valkey.Glide/Client/BaseClient.cs
currantw: Fixed! ✅ Added try/catch wrapping the entire callback. Logs the exception at Error level and returns 0 to trigger Rust's fallback to the original address.

tests/Valkey.Glide.UnitTests/AddressResolverTests.cs
currantw: Fixed! ✅ Added `AddressResolverTests.cs` with integration tests covering identity resolver, exception fallback, empty host fallback, port zero fallback, and oversized hostname fallback — for both standalone and cluster modes.

sources/Valkey.Glide/Client/BaseClient.cs
currantw: Fixed! ✅ Delegate uses `ushort` throughout now.

sources/Valkey.Glide/Client/BaseClient.cs
currantw: Fixed! ✅ Extracted `var addressResolver = config.Request.AddressResolver;` local variable before the lambda.

sources/Valkey.Glide/Client/BaseClient.cs
currantw: Fixed! ✅ Replaced `?? string.Empty` with the null-forgiving operator (`!`). `PtrToStringUTF8(ptr, len)` only returns null if the pointer is `IntPtr.Zero`, which can't happen here (Rust passes a valid `&str` pointer). If it somehow did, the `NullReferenceException` would be caught by the try/catch and logged.

alexr-bq (approved):

sources/Valkey.Glide/Client/BaseClient.cs
```diff
@@ -201,7 +201,57 @@ protected static async Task<T> CreateClient<T>(BaseClientConfiguration config, F
...
+
+                    byte[] bytes = System.Text.Encoding.UTF8.GetBytes(resolvedHost);
+                    if (bytes.Length > (int)bufLen)
+                    {
+                        var msg = $"Resolved host length ({bytes.Length} bytes) exceeds maximum length ({bufLen} bytes) for {host}:{port}";
+                        Logger.Log(Level.Error, "AddressResolver", msg);
+                        return 0;
+                    }
+
+                    if (resolvedPort <= 0)
```
xShinnRyuu: nit: This should be `resolvedPort == 0` since port values cannot be negative.


Feel free to resolve this comment

xShinnRyuu (approved):

sources/Valkey.Glide/Client/BaseClient.cs
currantw: Thanks. Fixed at `f82a0e8`. ✅

Issue #435, opened 2026-06-29
feat(compatibility): Add commands for compatibility with other GLIDE clients


currantw:
### Describe the feature

Support for the following commands was added to the Valkey GLIDE Java, Node, Python, and Go clients in [#6010](https://github.com/valkey-io/valkey-glide/issues/6010). The Valkey GLIDE C# must also implement them:

- [ ] [MIGRATE](https://valkey.io/commands/migrate/)
- [ ] [RESET](https://valkey.io/commands/reset/)
- [ ] [CLIENT PAUSE](https://valkey.io/commands/client-pause/)
- [ ] [CLIENT UNPAUSE](https://valkey.io/commands/client-unpause/)
- [ ] [CLIENT CACHING](https://valkey.io/commands/client-caching/)
- [ ] [CLIENT TRACKING](https://valkey.io/commands/client-tracking/)
- [ ] [CLIENT TRACKINGINFO](https://valkey.io/commands/client-trackinginfo/)
- [ ] [MONITOR](https://valkey.io/commands/monitor/)
- [ ] [SAVE](https://valkey.io/commands/save/)
- [ ] [BGSAVE](https://valkey.io/commands/bgsave/)
- [ ] [BGREWRITEAOF](https://valkey.io/commands/bgrewriteaof/)
- [ ] [REPLICAOF](https://valkey.io/commands/replicaof/)
- [ ] [FAILOVER](https://valkey.io/commands/failover/)
- [ ] [LATENCY HISTORY](https://valkey.io/commands/latency-history/)
- [ ] [LATENCY LATEST](https://valkey.io/commands/latency-latest/)
- [ ] [LATENCY RESET](https://valkey.io/commands/latency-reset/)
- [ ] [MEMORY DOCTOR](https://valkey.io/commands/memory-doctor/)
- [ ] [MEMORY MALLOC-STATS](https://valkey.io/commands/memory-malloc-stats/)
- [ ] [MEMORY PURGE](https://valkey.io/commands/memory-purge/)
- [ ] [MEMORY STATS](https://valkey.io/commands/memory-stats/)

### Use Case

See [#6010](https://github.com/valkey-io/valkey-glide/issues/6010).

### Proposed Solution

- Implement each command following the same approach as the other Valkey GLIDE clients.
- Add methods for StackExchange.Redis compatibility as necessary.

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might cause a breaking change

### Related

[...]

PR #447, opened 2026-07-12
Add `MIGRATE` command support

currantw:
## Summary

Add support for the `MIGRATE` command, enabling atomic key transfer between Valkey instances. Includes single-key (standalone + cluster) and multi-key (standalone-only) variants, batch/pipeline support, SER-compatible `KeyMigrateAsync`, and credential-safe `MigrateOptions` with `IDisposable` support.

## Issue Link

Part of #435.

- `MigrateAsync(ValkeyKey key, MigrateOptions options)` on `BaseClient` (both standalone and cluster)
- `MigrateAsync(IEnumerable<ValkeyKey> keys, MigrateOptions options)` on `GlideClient` (standalone-only)
- `MigrateOptions` class with:
  - Primary constructor: `(string host, ushort port, ushort destinationDb, TimeSpan timeout)`
  - Fluent builders: `WithCopy()`, `WithReplace()`, `WithAuth(password)`, `WithAuth(username, password)`
  - `IDisposable` — clears password from memory on dispose
  - `ObjectDisposedException` guards on all public methods after disposal
  - `sealed` to prevent security bypass via inheritance
- `KeyMigrateAsync` on `IDatabaseAsync` for StackExchange.Redis compatibility
- Batch support: single-key on `BaseBatch`, multi-key on `Batch`

## Implementation

- `MigrateOptions.ToArgs(GlideString[] keys)` builds the full wire-format argument array, handling both single-key and multi-key formats
- `Request.MigrateAsync` uses `Cmd<object, bool>` because the FFI returns `string` for "OK" but `GlideString` for "NOKEY"
- Password stored as `char[]` internally; `ToString()` omits credentials; `ClearPassword()` zeroes the array
- Renamed existing unused `MigrateOptions` flags enum → `KeyMigrateOptions` (moved to `abstract_Enums/SER/`)
- Moved standalone-only batch methods (`Move`, `Copy` with destDb) from `BaseBatch` into `Batch` directly
- Removed redundant `SelectAsync` from `IBatchStandalone` (already in `IBatchConnectionManagementCommands`)
- Removed unused `StringOKToBool` and `ObjectOKToBool` helpers from `Request.cs`

## Limitations

- `KeyMigrateAsync` (SER) does not expose AUTH/AUTH2 (matching StackExchange.Redis)
- Credential scrubbing is best-effort: `ToArgs` creates `new string(Password)` which persists until GC (same limitation as `ServerCredentials`)

## Testing

- **Unit tests**: `MigrateOptionsTests` (arg building, validation, dispose guards, password clearing), command args in `CommandTests`
- **Integration tests**: single-key success (both clients via `TestClients`), multi-key success (standalone), COPY flag, REPLACE flag, invalid host error, non-existent key NOKEY, empty keys validation
- **Batch tests**: shared single-key in `SharedBatchTests`, multi-key in `StandaloneClientTests`

## Related Issues

- https://github.com/valkey-io/valkey-glide/issues/5108
- https://github.com/valkey-io/valkey-glide/pull/6063
- #454 (follow-up: RENAME/MSET batch converter breaking change)

## Checklist


dev/scripts/validate_examples.py
```diff
@@ -44,6 +44,7 @@ class ExamplesValidator:
 
     # Namespace imports to include when --add-imports is specified.
     _USING_NAMESPACES = [
+        "System.Net",
```
currantw, line 47: Think this is a common enough import that we don't need to explicitly include it in all examples. 🤷

sources/Valkey.Glide/abstract_Enums/SER/KeyMigrateOptions.cs
```diff
@@ -1,12 +1,12 @@
...
 
 namespace Valkey.Glide;
 
 /// <summary>
-/// Additional options for the MIGRATE command.
+/// Additional options for the <see cref="IDatabaseAsync.KeyMigrateAsync"/> command.
 /// </summary>
 [Flags]
-public enum MigrateOptions
+public enum KeyMigrateOptions
```
currantw, line 9: This does NOT exactly match SER, where this enum is instead called `MigrateOptions`. My decision was to prioritize consistent naming for GLIDE (where this enum is not used) over SER. It is a simple rename for users.

sources/Valkey.Glide/Client/BaseClient.GenericCommands.cs
```diff
@@ -120,6 +120,10 @@ public async Task<ValkeyValue[]> SortAsync(ValkeyKey key, long skip = 0, long ta
     public async Task<long> SortAndStoreAsync(ValkeyKey destination, ValkeyKey key, long skip = 0, long take = -1, Order order = Order.Ascending, SortType sortType = SortType.Numeric, ValkeyValue by = default, IEnumerable<ValkeyValue>? get = null)
         => await Command(Request.SortAndStoreAsync(destination, key, skip, take, order, sortType, by, get?.ToArray()));
 
+    /// <inheritdoc cref="IBaseClient.MigrateAsync(ValkeyKey, MigrateOptions)"/>
+    public async Task<bool> MigrateAsync(ValkeyKey key, MigrateOptions options)
+        => await Command(Request.MigrateAsync([key], options));
```
currantw, line 125: Different interface than other Valkey GLIDE clients. All options contained in `MigrateOptions`. I think that this is more consistent with the pattern in the C# client, which tends to try to encapsulate parameters in `Options` (which also helps with validation).

sources/Valkey.Glide/Client/IBaseClient.GenericCommands.cs
```diff
@@ -556,6 +556,24 @@ public partial interface IBaseClient
...
+    /// <remarks>
+    /// <example>
+    /// <code>
+    /// await client.SetAsync("key", "value");
+    /// using var options = new MigrateOptions("desthost", 6379, 0, TimeSpan.FromSeconds(5));
+    /// var migrated = await client.MigrateAsync("key", options);
+    /// </code>
+    /// </example>
+    /// </remarks>
+    Task<bool> MigrateAsync(ValkeyKey key, MigrateOptions options);
```
currantw, line 575: Returning `bool` instead of `string` is also different from other clients, but it consistent with the pattern in C# (which also doesn't tend to return "OK" on success).

sources/Valkey.Glide/Commands/Options/MigrateOptions.cs
```diff
@@ -0,0 +1,190 @@
...
+/// Options for the MIGRATE command.
+/// </summary>
+/// <remarks>
+/// <para>See <a href="https://valkey.io/commands/migrate/">valkey.io</a></para>
+/// </remarks>
+/// <param name="host">The host address of the destination server.</param>
+/// <param name="port">The port number of the destination server.</param>
+/// <param name="destinationDb">The database number on the destination server.</param>
+/// <param name="timeout">The timeout for the migration.</param>
+public class MigrateOptions(string host, ushort port, int destinationDb, TimeSpan timeout) : IDisposable
```
currantw: Design decisions here:

- Includes features similar to `ServerCredentials` that help protect sensitive information, which was part of previous C# security releases. `MigrateOptions` is disposable (so password doesn't stay in memory), stores password as char array (so it can be explicitly cleared), and overrides `ToString` to avoid exposing it.
- Use `WithAuth` helpers to eliminate the possibility of invalid states (e.g. it is not possible for the user to specify a username without also specifying a password.).

sources/Valkey.Glide/Commands/Options/MigrateOptions.cs
```diff
@@ -0,0 +1,190 @@
...
+
+    /// <summary>
+    /// When <see langword="true"/>, do not remove the key from the local instance.
+    /// </summary>
+    public bool Copy { get; set; } = false;
+
+    /// <summary>
+    /// When <see langword="true"/>, replace existing key on the remote instance.
+    /// </summary>
+    public bool Replace { get; set; } = false;
```
currantw: Should we also add `WithCopy` and `WithReplace` builders and make these `private set` for consistency?

sources/Valkey.Glide/Commands/Options/MigrateOptions.cs
```diff
@@ -0,0 +1,190 @@
...
+
+        List<GlideString> args =
+        [
+            Host,
+            Port.ToGlideString(),
+            isMultiKey ? "" : keys.First(),
+            DestinationDb.ToGlideString(),
+
+            // TODO: Update once #446 is merged.
+            Timeout.TotalMilliseconds.ToGlideString()
```
currantw: Reminder comment for myself here. I'll resolve only once I've addressed this, after #446 is merged.

sources/Valkey.Glide/Internals/Request.cs
```diff
@@ -39,8 +38,17 @@ private static Cmd<T, bool> Boolean<T>(RequestType request, GlideString[] args)
...
+        => new(request, args, false, response => response == "OK");
+
+    /// <summary>
+    /// Create a Cmd which returns a Boolean value based on the response being OK or not.
+    /// </summary>
+    /// <param name="request">The request type</param>
+    /// <param name="args">The command arguments</param>
+    /// <returns>A command that converts the response to a boolean value (true if response equals OK)</returns>
+    private static Cmd<object, bool> ObjectOKToBool(RequestType request, GlideString[] args)
+        => new(request, args, false, response => (response is string s && s == "OK") || (response is GlideString gs && gs.ToString() == "OK"));
```
currantw: This is needed because the C# response handler currently returns `GlideString` for "NOKEY" but `string` for "OK", so we need to be able to handle both.

sources/Valkey.Glide/Internals/Request.cs
```diff
@@ -30,7 +30,6 @@ private static Cmd<T, T> Simple<T>(RequestType request, GlideString[] args, bool
     private static Cmd<T, bool> Boolean<T>(RequestType request, GlideString[] args)
         => new(request, args, false, response => Convert.ToInt64(response) == 1);
 
-    private static readonly Func<string, bool> IsOkConverter = response => response == "OK";
```
currantw, line 35: Moved inline because it is only used in one place below. 🤷

sources/Valkey.Glide/Pipeline/BaseBatch.GenericCommands.cs
```diff
@@ -81,12 +81,9 @@ public abstract partial class BaseBatch<T>
     public T Copy(ValkeyKey source, ValkeyKey destination, bool replace = false)
         => AddCmd(CopyAsync(source, destination, replace));
 
-    /// <inheritdoc cref="IBatchStandalone.Copy(ValkeyKey, ValkeyKey, int, bool)" />
-    public T Copy(ValkeyKey source, ValkeyKey destination, int destinationDatabase, bool replace = false)
-        => AddCmd(CopyAsync(source, destination, destinationDatabase, replace));
-
-    /// <inheritdoc cref="IBatchStandalone.Move(ValkeyKey, int)" />
-    public T Move(ValkeyKey key, int database) => AddCmd(MoveAsync(key, database));
```
currantw, line 89: Moved these standalone only methods to `Batch`. They don't belong here!

sources/Valkey.Glide/Pipeline/BaseBatch.GenericCommands.cs
```diff
@@ -122,34 +119,35 @@ public T SortReadOnly(ValkeyKey key, SortOptions? options)
     public T Wait(long numreplicas, TimeSpan timeout) => AddCmd(WaitAsync(numreplicas, timeout));
 
     // Explicit interface implementations for IBatchGenericCommands
-    IBatch IBatchGenericCommands.Delete(ValkeyKey key) => Delete(key);
```
currantw, line 125: Just made this alphabetical

sources/Valkey.Glide/Pipeline/Batch.cs
```diff
@@ -24,10 +27,22 @@ namespace Valkey.Glide.Pipeline;
...
+    public Batch Copy(ValkeyKey source, ValkeyKey destination, int destinationDatabase, bool replace = false)
+        => AddCmd(Request.CopyAsync(source, destination, destinationDatabase, replace));
+
+    /// <inheritdoc cref="IBatchStandalone.Migrate(IEnumerable{ValkeyKey}, MigrateOptions)" />
+    public Batch Migrate(IEnumerable<ValkeyKey> keys, MigrateOptions options)
+        => AddCmd(Request.MigrateAsync(keys, options));
+
+    /// <inheritdoc cref="IBatchStandalone.Move(ValkeyKey, int)" />
+    public Batch Move(ValkeyKey key, int database)
+        => AddCmd(Request.MoveAsync(key, database));
```
currantw, line 40: As mentioned in another comment, there methods were all previous incorrectly declared in `BaseBathc.GenericCommands`, even though they are standalone-only. Moved them to here where they belong.

tests/Valkey.Glide.IntegrationTests/StandaloneClientTests.cs
```diff
@@ -225,40 +226,69 @@ public async Task KeyCopy_Move(GlideClient client)
     }
 
     [Theory]
-    [InlineData(true)]
-    [InlineData(false)]
-    public async Task BatchKeyCopyAndKeyMove(bool isAtomic)
```
currantw, line 230: Split into separate tests for each method.

tests/Valkey.Glide.UnitTests/Options/MigrateOptionsTests.cs
```diff
@@ -0,0 +1,36 @@
+// Copyright Valkey GLIDE Project Contributors - SPDX Identifier: Apache-2.0
+
+using Valkey.Glide.Commands.Options;
+
+namespace Valkey.Glide.UnitTests;
+
+public class MigrateOptionsTests
```
currantw, line 9: We don't have any other unit tests for `Options` classes yet, but think it makes sense to start adding them for better coverage.

tests/Valkey.Glide.UnitTests/CommandTests.cs
```diff
@@ -189,46 +189,49 @@ public void ValidateCommandArgs() => Assert.Multiple(
             () => Assert.Equal(["SSCAN", "", "0"], Request.SetScanAsync("", 0).GetArgs()),
 
             // Generic Commands
+            () => Assert.Equal(["COPY", "src", "dest", "DB", "1", "REPLACE"], Request.CopyAsync("src", "dest", 1, true).GetArgs()),
```
currantw, line 203: Again, just alphabetized. Just focus on new MIGRATE tests.

sources/Valkey.Glide/Commands/Options/MigrateOptions.cs
```diff
@@ -0,0 +1,190 @@
...
+    /// <returns>This instance for fluent chaining.</returns>
+    public MigrateOptions WithAuth(string password)
+    {
+        ArgumentNullException.ThrowIfNull(password, nameof(password));
+
+        Username = null;
+        Password = password.ToCharArray();
+
+        return this;
+    }
```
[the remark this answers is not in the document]
currantw: Addressed in a518ef4b.

sources/Valkey.Glide/Commands/Options/MigrateOptions.cs
```diff
@@ -0,0 +1,190 @@
...
+    public MigrateOptions WithAuth(string username, string password)
+    {
+        ArgumentNullException.ThrowIfNull(username, nameof(username));
+        ArgumentNullException.ThrowIfNull(password, nameof(password));
+
+        Username = username;
+        Password = password.ToCharArray();
+
+        return this;
+    }
```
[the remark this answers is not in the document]
currantw: Addressed in a518ef4b.

sources/Valkey.Glide/Abstract/Database.GenericCommands.cs
```diff
@@ -223,6 +223,27 @@ public Task<bool> KeyMoveAsync(ValkeyKey key, int database, CommandFlags flags =
...
+
+        (string host, ushort port) = Utils.SplitEndpoint(toServer);
+        var options = new MigrateOptions(host, port, toDatabase, TimeSpan.FromMilliseconds(timeoutMilliseconds))
+        {
+            Copy = migrateOptions.HasFlag(KeyMigrateOptions.Copy),
+            Replace = migrateOptions.HasFlag(KeyMigrateOptions.Replace),
+        };
+
+        return await MigrateAsync(key, options);
+    }
```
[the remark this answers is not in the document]
currantw: Addressed in a518ef4b.

tests/Valkey.Glide.TestUtils/Options.cs
```diff
@@ -0,0 +1,19 @@
+// Copyright Valkey GLIDE Project Contributors - SPDX Identifier: Apache-2.0
+
+using Valkey.Glide.Commands.Options;
+
+namespace Valkey.Glide.TestUtils;
+
+/// <summary>
+/// Helper methods for building command options in tests.
+/// </summary>
+public static class Options
```
currantw, line 10: Helper class. `BuildMigrateOptions` simplifies test by only requiring the caller to specify the arguments that are actually necessary for the test. Just this method for now, but I have other similar ones I plan to move her in future PRs.

sources/Valkey.Glide/Commands/Options/MigrateOptions.cs
```diff
@@ -0,0 +1,190 @@
...
+        {
+            args.Add(ValkeyLiterals.REPLACE);
+        }
+
+        if (Username is not null && Password is not null)
+        {
+            args.Add(ValkeyLiterals.AUTH2);
+            args.Add(Username);
+            args.Add(new string(Password));
+        }
```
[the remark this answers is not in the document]
currantw: Addressed in a518ef4b.

sources/Valkey.Glide/abstract_Enums/SER/KeyMigrateOptions.cs
affonsov, line 9: changing public enum name is a breaking change

sources/Valkey.Glide/Commands/Options/MigrateOptions.cs
```diff
@@ -0,0 +1,220 @@
...
+    /// <summary>
+    /// Returns a string representation with sensitive data omitted.
+    /// </summary>
+    public override string ToString()
+        => $"MigrateOptions {{ Host = {Host}, Port = {Port}, DestinationDb = {DestinationDb}, Timeout = {Timeout}, Copy = {Copy}, Replace = {Replace} }}";
+
+    /// <summary>
+    /// Clears sensitive data.
+    /// </summary>
+    public void Dispose()
```
xShinnRyuu, line 144: Can we add in a unit test to see dispose will clear the password array?

sources/Valkey.Glide/Commands/Options/MigrateOptions.cs
```diff
@@ -0,0 +1,220 @@
...
+/// Options for the MIGRATE command.
+/// </summary>
+/// <remarks>
+/// <para>See <a href="https://valkey.io/commands/migrate/">valkey.io</a></para>
+/// </remarks>
+/// <param name="host">The host address of the destination server.</param>
+/// <param name="port">The port number of the destination server.</param>
+/// <param name="destinationDb">The database number on the destination server.</param>
+/// <param name="timeout">The timeout for the migration.</param>
+public class MigrateOptions(string host, ushort port, int destinationDb, TimeSpan timeout) : IDisposable
```
xShinnRyuu: This might be a slight security violation

When a class like `MigrateOptions` implements `IDisposable` to securely clear sensitive data (passwords stored as `char[]`), leaving it unsealed creates a potential security bypass vector.

Here's what could happen:
```
// Malicious or poorly implemented derived class
public class InsecureMigrateOptions : MigrateOptions
{
    private char[]? _passwordCopy;

    public InsecureMigrateOptions(string host, ushort port, int db, TimeSpan timeout)
        : base(host, port, db, timeout) { }

    public new MigrateOptions WithAuth(string password)
    {
        // Secretly keep a copy of the password
        _passwordCopy = password.ToCharArray();
        return base.WithAuth(password);
    }

    // Override Dispose to prevent clearing
    protected override void Dispose(bool disposing)
    {
        // Intentionally skip the base.Dispose() call
        // or call it but keep our own copy
        GC.SuppressFinalize(this);
    }

    // Later access the password even after "disposal"
    public string? GetStoredPassword() => _passwordCopy != null ? new string(_passwordCopy) : null;
}
```
Why This Matters

1. Security Contract Violation: The base `MigrateOptions` class promises that `Dispose()` will clear credentials from memory. A derived class can break this contract.
2. Unexpected Credential Persistence: Code using the derived class through the base interface expects credentials to be cleared, but they persist in memory longer than intended.
3. Accidental Misuse: Even without malicious intent, a derived class might override `Dispose()` for legitimate reasons but forget to call `base.Dispose()`.


The Solution: Sealed Classes
```
public sealed class MigrateOptions : IDisposable  // Add 'sealed'
{
    // ... existing implementation
}
```
Making the class sealed:

- Prevents inheritance: No derived classes can be created
- Guarantees disposal behavior: The credential-clearing logic cannot be bypassed
- Makes security assumptions explicit: The class is designed as a complete, final implementation

When to Seal Credential-Handling Classes

Seal classes that:

- Store sensitive data (passwords, keys, tokens)
- Implement security-critical disposal logic
- Are not designed for inheritance
- Need predictable cleanup behavior

sources/Valkey.Glide/Commands/Options/MigrateOptions.cs
currantw: Good call. Added in 849c46ab.

sources/Valkey.Glide/Commands/Options/MigrateOptions.cs
currantw: Good catch (and interesting)! Make `sealed` in ea0d41b7.

sources/Valkey.Glide/abstract_Enums/SER/KeyMigrateOptions.cs
currantw: The enum previously had zero source-code consumers since the `MIGRATE` command was never implemented. It must just have been bulk-copied from SER? So nobody using GLIDE would have had any reason to use it, although it was technically exposed?

affonsov: source-code usage is fine, we need to think in user usage.
In theory it was released, so someone could be using right? 
I think you could use [Obsolete] and point to the new enum. Take a look at that

currantw: > In theory it was released, so someone could be using right?

In _theory_, yes.

> I think you could use [Obsolete] and point to the new enum. Take a look at that

Looked into this. Don't believe that this would work because of the new GLIDE class `MigrateOptions`, so there would be a naming conflict. 

Some other options:

- We could mark the SER `MigrateOptions` enum as `Obselete` and then hold off implementing the `MIGRATE` command for C# until release 2.0. This would mean that the C# client would be missing functionality that is available in the other GLIDE clients _and_ in StackExchange.Redis for quite some time, although users could obviously still use a custom command if they really needed.
- We could name the GLIDE class something like `MigrateCommandOptions` instead. This would avoid the naming conflict, but it would be inconsistent with every other `Options` class in the codebase. And because both these types could actually be used after this upcoming release, we would effectively be stuck with these names forevermore.

Let me know what you think!

sources/Valkey.Glide/Commands/Options/MigrateOptions.cs
currantw: Fixed!

sources/Valkey.Glide/Abstract/Database.GenericCommands.cs
```diff
@@ -223,6 +223,33 @@ public Task<bool> KeyMoveAsync(ValkeyKey key, int database, CommandFlags flags =
...
+        EndPoint toServer,
+        int toDatabase = 0,
+        int timeoutMilliseconds = 0,
+        KeyMigrateOptions migrateOptions = KeyMigrateOptions.None,
+        CommandFlags flags = CommandFlags.None)
+    {
+        GuardClauses.ThrowIfCommandFlags(flags);
+
+        (string host, ushort port) = Utils.SplitEndpoint(toServer);
+        using var options = new MigrateOptions(host, port, (ushort)toDatabase, TimeSpan.FromMilliseconds(timeoutMilliseconds));
```
xShinnRyuu, line 238: What is the range for `toDatabase`? `ushort` is 0-65535 I believe will this cause silent truncation issues?

Im assuming the `toDatabase` is a port value 🤔 and this is a non-issue.

sources/Valkey.Glide/Internals/Request.cs
xShinnRyuu: Does this fact need to be added in the comment section above?

tests/Valkey.Glide.TestUtils/Options.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    /// <summary>
+    /// Builds and returns migrate options for testing.
+    /// If not specified, required arguments are populated w ith default values.
```
xShinnRyuu: type `w ith`

sources/Valkey.Glide/Commands/Options/MigrateOptions.cs
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    #endregion
+    #region Internal Methods
+
+    /// <summary>
+    /// Converts the options to command arguments.
+    /// </summary>
+    internal GlideString[] ToArgs(GlideString[] keys)
+    {
```
xShinnRyuu, line 175: suggestion to add in a small guard if `ToArgs()` was ever called after `dispose()` already occurred.
```
// Throw if already disposed - fail fast
ObjectDisposedException.ThrowIf(_disposed, this);
```

xShinnRyuu (approved):

sources/Valkey.Glide/Abstract/Database.GenericCommands.cs
currantw: `toDatabase` is for Valkey's numbered database feature, which is between 0 and 15: 
https://valkey.io/blog/numbered-databases/

tests/Valkey.Glide.TestUtils/Options.cs
currantw: Fixed in 551b1960.

sources/Valkey.Glide/Commands/Options/MigrateOptions.cs
currantw: Good call. Done in a0561919

sources/Valkey.Glide/abstract_Enums/SER/KeyMigrateOptions.cs
affonsov: Is there a naming conflict? they are under different namespace right?

affonsov (approved):

sources/Valkey.Glide/abstract_Enums/SER/KeyMigrateOptions.cs
affonsov: Clarified, there is no code usage of MigrateOptions so there is no risk of any user on using it.

sources/Valkey.Glide/Internals/Request.cs
currantw: Thanks for your comments! This lead me down a bit of a rabbit-hole, which lead to me discovering and raising this issue: https://github.com/valkey-io/valkey-glide-csharp/issues/454. Take a look. I think the changes should address your concerns!

PR #455, opened 2026-07-15
Add `NodeDiscoveryMode` configuration option for standalone clients

### LINKED ISSUE #131: Support for connecting via Envoyproxy or Twemproxy
## Summary
StackExchange.Redis provides ability to connect to fronting redis/valkey [proxy](https://github.com/valkey-io/valkey-glide-csharp/blob/main/sources/Valkey.Glide/abstract_Enums/Proxy.cs)

## Requirements:
- disable topology discovery, route commands to endpoints without slot/shard calculation
- restrict commands to those supported by proxy (remove INFO requirement on connection)

jamesx-improving:
## Summary

Adds a `NodeDiscoveryMode` configuration option for standalone clients, controlling how the client discovers node roles and topology during connection initialization. Three modes are supported:

- **`Standard`** (default): Verify node roles via `INFO REPLICATION`, use only the provided addresses. This is the current behavior — fully backward compatible.
- **`Static`**: Skip role detection entirely. Trust the provided addresses as-is; the first address is treated as the primary. Designed for proxy compatibility (e.g. Envoy/Twemproxy) or known-static topologies where `INFO` is unavailable.
- **`DiscoverAll`**: Discover the full topology (primary + all replicas) from any starting node. Provide any single node address and the client will find and connect to all other nodes.

This brings the C# binding in line with the other four language bindings (Rust core, Python, Java, Node.js, Go) implemented in the main valkey-glide repo.

## Issue Link

Closes #131.

- New public `NodeDiscoveryMode` enum (`Standard = 0`, `Static = 1`, `DiscoverAll = 2`) on `ConnectionConfiguration`.
- New `NodeDiscoveryMode` property and `WithNodeDiscoveryMode(...)` fluent builder method on `StandaloneClientConfigurationBuilder`.
- **`Static` mode** skips `INFO REPLICATION` during connection setup and assumes the first provided address is the primary — enabling connectivity through proxies such as Envoy that don't support the `INFO` command (the original ask in #131).
- **`DiscoverAll` mode** discovers the full standalone topology from any starting node.
- **`Standard` mode** (default) maintains the current behavior with zero changes. Existing code continues to work without modification.
- The option is standalone-only; it is ignored in cluster mode, which has its own topology discovery mechanism.

## Implementation

Unlike the main valkey-glide repo (which edits the protobuf schema + glide-core), the C# repo consumes `glide-core` directly through its own thin Rust FFI shim, which was already hardcoding `node_discovery_mode: NodeDiscoveryMode::default()`. This PR wires the option through the existing FFI chain:

- **`sources/Valkey.Glide/ConnectionConfiguration.cs`**: public `NodeDiscoveryMode` enum, field on the internal `ConnectionConfig` record + `ToFfi()` wiring, and the standalone builder property/method.
- **`sources/Valkey.Glide/Internals/FFI.structs.cs`**: new parameter on the FFI `ConnectionConfig` constructor and a `NodeDiscoveryMode` field on the `ConnectionRequest` struct, placed right after `ReadOnly` to match the Rust layout.
- **`rust/src/ffi.rs`**: new `#[repr(u32)]` FFI `NodeDiscoveryMode` enum, a field on the FFI struct, and a match mapping it to `glide_core::client::NodeDiscoveryMode` (replacing the hardcoded `::default()`).

Reviewers may want to pay extra attention to the FFI struct layout alignment: the C# `enum : uint` (4 bytes) mirrors the Rust `#[repr(u32)]` enum, and the field is inserted at the same position on both sides (between `read_only` and `has_client_side_cache_config`), following the existing `TlsMode`/`Protocol` field pattern.

## Limitations

- **`Static` mode + `ClientName`**: Do not set `ClientName` when using `Static` mode with a proxy, as the proxy may not support the `CLIENT SETNAME` command sent during connection setup.
- **`DiscoverAll` mode**: Discovery happens only at client creation time — no periodic re-discovery. Replicas added after client creation are not discovered until the client is recreated.
- **`DiscoverAll` + read-only**: These modes are mutually exclusive (read-only skips the `INFO REPLICATION` call that discovery requires); rejected by glide-core at client creation time.
- **Standalone mode only**: `NodeDiscoveryMode` is ignored in cluster mode.

## Testing

- Added unit tests in `ConnectionConfigurationTests.cs` covering the default (`Standard`), each mode via the builder and via the property setter, and an enum-discriminant assertion that guards the `{0,1,2}` values against the Rust FFI / glide-core protobuf contract.
- `task build`, `task test:unit` (534/534 passing), `task lint:csharp`, and `task lint:rust` (fmt, clippy `-D warnings`, cargo-deny, cargo-doc) all pass locally.

## Related Issues

- Mirrors valkey-io/valkey-glide#5724 (NodeDiscoveryMode across the other bindings).

## Checklist

### DIFF
--- a/sources/Valkey.Glide/ConnectionConfiguration.cs
+++ b/sources/Valkey.Glide/ConnectionConfiguration.cs
@@ -60,7 +60,7 @@ public abstract class ConnectionConfiguration
         public TimeSpan? PubSubReconciliationInterval;
         public CompressionConfig? CompressionConfig;
         public bool ReadOnly;
-        public NodeDiscoveryMode NodeDiscoveryMode;
+        public NodeDiscoveryMode NodeDiscoveryMode = NodeDiscoveryMode.Standard;
         public ClientSideCacheConfig? ClientSideCacheConfig;
         public AddressResolverDelegate? AddressResolver;
 
@@ -223,40 +223,6 @@ public abstract class ConnectionConfiguration
         /// Use RESP3 to communicate with the server nodes.
         /// </summary>
         RESP3 = 1,
-    }
-
-    /// <summary>
-    /// Controls how the client discovers node roles and topology in standalone mode.
-    /// <para />
-    /// This option is only relevant for standalone clients; it is ignored in cluster mode, which
-    /// has its own topology discovery mechanism.
-    /// </summary>
-    public enum NodeDiscoveryMode : uint
-    {
-        /// <summary>
-        /// Default: verify node roles via <c>INFO REPLICATION</c>, use only the provided addresses.
-        /// This is the fully backward-compatible behavior.
-        /// </summary>
-        Standard = 0,
-        /// <summary>
-        /// Skip role detection entirely. Trust the provided addresses as-is; the first address is
-        /// treated as the primary. Use when connecting through a proxy (e.g., Envoy) or when the
-        /// topology is known and static.
-        /// <para />
-        /// <b>Note</b>: Do not set <see cref="ClientConfigurationBuilder{T}.ClientName"/> when using
-        /// this mode with a proxy, as the proxy may not support the <c>CLIENT SETNAME</c> command
-        /// sent during connection setup.
-        /// </summary>
-        Static = 1,
-        /// <summary>
-        /// Discover the full topology (primary + all replicas) from any starting node. Provide any
-        /// single node address (primary or replica) and the client will find and connect to all
-        /// other nodes.
-        /// <para />
-        /// Discovery happens only at client creation time; topology changes after creation are not
-        /// re-discovered until the client is recreated.
-        /// </summary>
-        DiscoverAll = 2,
     }
 
     #endregion
@@ -974,12 +940,7 @@ public abstract class ConnectionConfiguration
         /// <summary>
         /// Controls how the client discovers node roles and topology during connection
         /// initialization. If not set, defaults to <see cref="NodeDiscoveryMode.Standard" />.
-        /// <para />
-        /// <see cref="NodeDiscoveryMode.DiscoverAll" /> is mutually exclusive with read-only mode
-        /// (read-only skips the <c>INFO REPLICATION</c> call that discovery requires); this
-        /// combination is rejected at client creation time.
-        /// <para />
-        /// See also <seealso cref="NodeDiscoveryMode" />.
+        /// See <see cref="NodeDiscoveryMode" /> for the available modes and their details.
         /// </summary>
         public NodeDiscoveryMode NodeDiscoveryMode
         {
--- a/sources/Valkey.Glide/ConnectionConfiguration.cs
+++ b/sources/Valkey.Glide/ConnectionConfiguration.cs
@@ -60,6 +60,7 @@ public abstract class ConnectionConfiguration
         public TimeSpan? PubSubReconciliationInterval;
         public CompressionConfig? CompressionConfig;
         public bool ReadOnly;
+        public NodeDiscoveryMode NodeDiscoveryMode;
         public ClientSideCacheConfig? ClientSideCacheConfig;
         public AddressResolverDelegate? AddressResolver;
 
@@ -83,6 +84,7 @@ public abstract class ConnectionConfiguration
                 (uint?)PubSubReconciliationInterval?.TotalMilliseconds,
                 CompressionConfig?.ToFfi(),
                 ReadOnly,
+                NodeDiscoveryMode,
                 ClientSideCacheConfig?.ToFfi()
             );
     }
@@ -221,6 +223,40 @@ public abstract class ConnectionConfiguration
         /// Use RESP3 to communicate with the server nodes.
         /// </summary>
         RESP3 = 1,
+    }
+
+    /// <summary>
+    /// Controls how the client discovers node roles and topology in standalone mode.
+    /// <para />
+    /// This option is only relevant for standalone clients; it is ignored in cluster mode, which
+    /// has its own topology discovery mechanism.
+    /// </summary>
+    public enum NodeDiscoveryMode : uint
+    {
+        /// <summary>
+        /// Default: verify node roles via <c>INFO REPLICATION</c>, use only the provided addresses.
+        /// This is the fully backward-compatible behavior.
+        /// </summary>
+        Standard = 0,
+        /// <summary>
+        /// Skip role detection entirely. Trust the provided addresses as-is; the first address is
+        /// treated as the primary. Use when connecting through a proxy (e.g., Envoy) or when the
+        /// topology is known and static.
+        /// <para />
+        /// <b>Note</b>: Do not set <see cref="ClientConfigurationBuilder{T}.ClientName"/> when using
+        /// this mode with a proxy, as the proxy may not support the <c>CLIENT SETNAME</c> command
+        /// sent during connection setup.
+        /// </summary>
+        Static = 1,
+        /// <summary>
+        /// Discover the full topology (primary + all replicas) from any starting node. Provide any
+        /// single node address (primary or replica) and the client will find and connect to all
+        /// other nodes.
+        /// <para />
+        /// Discovery happens only at client creation time; topology changes after creation are not
+        /// re-discovered until the client is recreated.
+        /// </summary>
+        DiscoverAll = 2,
     }
 
     #endregion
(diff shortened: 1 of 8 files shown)


sources/Valkey.Glide/ConnectionConfiguration.cs
```diff
@@ -223,6 +225,40 @@ public enum Protocol : uint
         RESP3 = 1,
     }
 
+    /// <summary>
+    /// Controls how the client discovers node roles and topology in standalone mode.
+    /// <para />
+    /// This option is only relevant for standalone clients; it is ignored in cluster mode, which
+    /// has its own topology discovery mechanism.
+    /// </summary>
+    public enum NodeDiscoveryMode : uint
```
currantw: The organization for this project is kinda all over the place, but would `abstract_Enums/` be a better spot for this? Plus, this file is already 1000+ files, so would probably be nice to pull this out 🤷

sources/Valkey.Glide/ConnectionConfiguration.cs
```diff
@@ -223,6 +225,40 @@ public enum Protocol : uint
...
+        /// <summary>
+        /// Discover the full topology (primary + all replicas) from any starting node. Provide any
+        /// single node address (primary or replica) and the client will find and connect to all
+        /// other nodes.
+        /// <para />
+        /// Discovery happens only at client creation time; topology changes after creation are not
+        /// re-discovered until the client is recreated.
+        /// </summary>
+        DiscoverAll = 2,
+    }
```
currantw: Are they any existing Valkey GLIDE docs for `NodeDiscoveryMode` that we can reference here? In my mind, it would be better to include detailed information on this (e.g. don't use `CLIENT SETNAME` with `Static`, that `DiscoverAll` only discovers at client creation time) in common documentation that we could _reference_ from here, and then keep the documentation on this `enum` itself relatively simple/high-level and just point users to that common documentation for more details. 🤷

sources/Valkey.Glide/ConnectionConfiguration.cs
```diff
@@ -934,6 +970,31 @@ public StandaloneClientConfigurationBuilder() : base(false) { }
...
 
+        #region Node Discovery Mode
+        /// <summary>
+        /// Controls how the client discovers node roles and topology during connection
+        /// initialization. If not set, defaults to <see cref="NodeDiscoveryMode.Standard" />.
+        /// <para />
+        /// <see cref="NodeDiscoveryMode.DiscoverAll" /> is mutually exclusive with read-only mode
+        /// (read-only skips the <c>INFO REPLICATION</c> call that discovery requires); this
+        /// combination is rejected at client creation time.
+        /// <para />
```
currantw: Nit. As with previous comment. might be worth centralizing this information in one place? 🤷

sources/Valkey.Glide/ConnectionConfiguration.cs
```diff
@@ -60,6 +60,7 @@ internal record ConnectionConfig
         public TimeSpan? PubSubReconciliationInterval;
         public CompressionConfig? CompressionConfig;
         public bool ReadOnly;
+        public NodeDiscoveryMode NodeDiscoveryMode;
```
currantw: Nit. Is it worth being explicit with the default here?


```suggestion
        public NodeDiscoveryMode NodeDiscoveryMode = NodeDiscoveryMode.Standard.
```

sources/Valkey.Glide/Internals/FFI.structs.cs
```diff
@@ -1130,6 +1132,8 @@ private struct ConnectionRequest
         [MarshalAs(UnmanagedType.U1)]
         public bool ReadOnly;
```
currantw, line 1140: Nit. Would you mind separating these for me while you're here? I'm just too pedantic 😆 
Just feel free to resolve this comment yourself.

```suggestion
[MarshalAs(UnmanagedType.U1)]
public bool HasCompressionConfig;
public CompressionConfig CompressionConfig;

[MarshalAs(UnmanagedType.U1)]
public bool ReadOnly;
```

tests/Valkey.Glide.UnitTests/ConnectionConfigurationTests.cs
currantw, line 1: Is it possible to have integration tests for this? If we can't directly differentiate between the cases, it would at least be good to test creating an actual client with each enum options?

tests/Valkey.Glide.UnitTests/ConnectionConfigurationTests.cs
```diff
@@ -268,6 +268,48 @@ public void RefreshTopologyFromInitialNodes_False()
...
+    [Theory]
+    [InlineData(NodeDiscoveryMode.Standard)]
+    [InlineData(NodeDiscoveryMode.Static)]
+    [InlineData(NodeDiscoveryMode.DiscoverAll)]
+    public void WithNodeDiscoveryMode_SetsMode(NodeDiscoveryMode mode)
+    {
+        var builder = new StandaloneClientConfigurationBuilder()
+            .WithNodeDiscoveryMode(mode);
+        Assert.Equal(mode, builder.Build().Request.NodeDiscoveryMode);
+    }
```
currantw, line 288: Could define `NodeDiscoverModes` in `TestUtils.Data`:

```csharp
    public static TheoryData<NodeDiscoveryMode> NodeDiscoveryModes
        => new(Enum.GetValues<NodeDiscoveryMode>());
```

and then use that for parametrization here:

```csharp
    [Theory]
    [MemberData(nameof(AllNodeDiscoveryModes))]
    public void WithNodeDiscoveryMode_SetsMode(NodeDiscoveryMode mode)
```

That way, if we ever add new values (unlikely, but possible?), we won't need to update this.

tests/Valkey.Glide.UnitTests/ConnectionConfigurationTests.cs
```diff
@@ -268,6 +268,48 @@ public void RefreshTopologyFromInitialNodes_False()
...
+
+    [Fact]
+    public void NodeDiscoveryMode_EnumValues_MatchFfiContract()
+    {
+        // These discriminants must stay aligned with the Rust FFI `NodeDiscoveryMode` enum and
+        // the glide-core protobuf values.
+        Assert.Equal(0u, (uint)NodeDiscoveryMode.Standard);
+        Assert.Equal(1u, (uint)NodeDiscoveryMode.Static);
+        Assert.Equal(2u, (uint)NodeDiscoveryMode.DiscoverAll);
+    }
```
currantw, line 321: Hmmm. Do we have any tests for the actually FFI configuration output? Rather than verifying the value of the enum directly, perhaps it is better to verify that we are actually passing the correct values to the FFI layer? That seems more consistent with a connection configuraiton test, to my eye? 🤷

currantw (commented):
Reviewed to 3d75d15.

sources/Valkey.Glide/Internals/FFI.structs.cs
```diff
@@ -200,6 +200,12 @@ internal class ConnectionConfig : Marshallable
     {
         private ConnectionRequest _request;
 
+        /// <summary>
+        /// The node discovery mode marshalled into the underlying FFI request. Exposed for testing
+        /// that the value is correctly wired through to the FFI layer.
+        /// </summary>
+        internal NodeDiscoveryMode NodeDiscoveryMode => _request.NodeDiscoveryMode;
```
currantw, line 207: Nit. Would it make more sense to just make `_request` internal so we don't increase the interface?

currantw (commented):
Reviewed to `eb4a4ce`. Just one nit and one remaining unresolved comment.

tests/Valkey.Glide.UnitTests/ConnectionConfigurationTests.cs
jamesx-improving: Yep, added in `eb4a4ce` — `Connect_WithNodeDiscoveryMode_Succeeds` in `StandaloneClientTests.cs`, parametrized over all modes via the new `Data.NodeDiscoveryModes`. It creates a real `GlideClient` with each mode and asserts it connects + pings.

sources/Valkey.Glide/Internals/FFI.structs.cs
jamesx-improving: I looked at that — there's a snag: `_request` is a `ConnectionRequest`, which is a `private struct` (line 1077), so an `internal` field of a `private` type won't compile (CS0052). We'd have to promote the struct to `internal` too.

Once we do that, exposing `_request` actually widens the surface more than the accessor does: the raw struct has ~33 public fields (pointers, counts, has-flags, the whole marshalling layout), and tests would then couple to that internal layout. The single `internal NodeDiscoveryMode` property exposes just one intention-revealing member and keeps the raw struct private, so I'd lean toward keeping it as-is if that works for you.

currantw: Sure, that makes sense to me. We can re-assess in the future if we end up with a bunch more of these. 👍

currantw (approved):
Reviewed to 53f1755.
Looks good ✅

Aryex (approved):

PR #431, opened 2026-06-23
fix #430: Parse all values returned by XREADGROUP

fs-paul:
## Summary

Parse all values from XREADGROUP

## Issue Link

Closes #430 .

## Features and Behaviour Changes

No new features, only iterates through the outer array of a stream's data instead of grabbing the first one only

## Implementation

Updated both `ConvertSingleStreamRead` and `ConvertMultiStreamRead` as they both used the same logic. Replaced the `outerArray[0]` with a loop to build the NameValueEntry array.

## Limitations

N/A

## Testing

Ran the test suite locally

## Related Issues

N/A

## Checklist

- [x] This Pull Request is related to one issue.
- [x] Commit message has a detailed description of what changed and why.
- [x] Tests are added or updated and all checks pass.
- [x] `CHANGELOG.md`, `README.md`, `DEVELOPER.md`, and other documentation files are updated.
- [x] Destination branch is correct - `main` or release
- [x] Create merge commit if merging release branch into `main`, squash otherwise.

Issue #326, opened 2026-04-17
Add GLIDE interfaces for remaining stream commands


currantw:
### Summary

Add GLIDE-native interfaces (`IBaseClient`, `IStreamBaseCommands`, `IBatchStreamCommands`) for the remaining stream commands that currently only have SER-compatible implementations on `IDatabaseAsync`.

### Description

PR #324 enabled GLIDE interfaces for core stream commands (XADD, XREAD, XRANGE/XREVRANGE, XREADGROUP, XLEN, XDEL) but temporarily disabled the GLIDE-native methods for the remaining commands. The SER layer (`IDatabaseAsync`) works for all of them, but the GLIDE layer needs to be re-enabled with proper cleanup.

Commands needing GLIDE interfaces:

- XGROUP CREATE
- XGROUP SETID
- XGROUP DESTROY
- XGROUP CREATECONSUMER
- XGROUP DELCONSUMER
- XCLAIM / XCLAIM JUSTID
- XAUTOCLAIM / XAUTOCLAIM JUSTID
- XTRIM
- XACK
- XPENDING
- XINFO STREAM / XINFO STREAM FULL
- XINFO GROUPS
- XINFO CONSUMERS

The internal `Request` layer already supports all of these. The work is to uncomment and clean up the `IBaseClient`, `IStreamBaseCommands`, `IBatchStreamCommands`, and `BaseClient` methods, ensuring cross-client consistency with Java, Python, Go, and Node GLIDE clients.

### Proposed Solution

For each command, uncomment the GLIDE methods on `IBaseClient`/`IStreamBaseCommands`/`IBatchStreamCommands`, verify signatures match other GLIDE clients, and re-enable the corresponding integration tests.

### Effort & Severity Estimates

- Effort: medium
- Severity: medium

### Related

- #270 — Parent issue: command API cleanup
- #282 — Stream command cleanup parent issue
- #324 — PR that disabled these GLIDE methods
- #322 — `claimMinIdleTime` support for SER `StreamReadGroupAsync`

PR #464, opened 2026-07-20
chore: code cleanup following new command PRs (#436–#457)

currantw:
## Summary

Code quality cleanup following the batch of new command PRs (#436–#457). Standardizes code organization, naming, documentation patterns, and type definitions across the files modified in those PRs.

## Issue Link

Closes #435.

## Features and Behaviour Changes

- Removed obsolete `SaveType.ForegroundSave` enum variant (was `[Obsolete]` and threw `NotSupportedException`).
- Moved `CopyAsync(destinationDatabase)` and `MoveAsync` from `IGlideClient` to `IBaseClient` (both standalone and cluster clients need them).
- Fixed `IGlideClusterClient.PubSubCommands` namespace from `Valkey.Glide.Commands` to `Valkey.Glide` (was an incorrect partial interface split).

### Type Standardization

- Converted `ClientTrackingInfo`, `MemoryStats`, `MonitorMessage` to `sealed record` with internal constructors and `required init` properties.
- Converted `MemoryStatsDb`, `LatencyEntry`, `LatencyEventInfo` to `readonly record struct` (value types for allocation efficiency).

### Documentation

- Standardized XML doc summaries to use `<see href="...">COMMAND</see>` linking to Valkey docs.
- Added `<seealso href="..."/>` to all new API types and options classes.

### Code Organization

- Moved server management and generic command methods from `GlideClient.cs`/`GlideClusterClient.cs` into dedicated partial files (`*.ServerManagementCommands.cs`, `*.GenericCommands.cs`).
- Ordered methods alphabetically in all client, interface, batch, and request builder files.
- Added consistent `#region` usage (Constants, Command Builders, Response Converters, Argument Builders) to all `Request.*` partial files.
- Standardized region naming to "Private Fields" and "Constructors & Builders".

### Cleanup

- Renamed `TestUtils.Options` class to `Builders`; consolidated test helper methods.
- Removed unnecessary `.ToGlideString()` calls where implicit conversions exist.
- Added `APPEND`, `LIBRARYNAME`, `WITHCODE` to `ValkeyLiterals`; replaced string literals.
- Improved `GuardClauses.ThrowIfWhenExists` error message to use `nameof()`.
- Added TODO #462 comments for future consolidation of redundant command overrides.

## Limitations

:white_circle: None — this is a cleanup/refactoring PR with no new features.

## Testing

- All 547 unit tests pass.
[...]

PR #469, opened 2026-07-21
Fix `XREADGROUP` multi-field parsing and add stream read tests

### LINKED ISSUE #430: fix(StreamReadGroupAsync): Only parses the first field returned by XREADGROUP for each entry
### Describe the bug

When using StreamReadGroupAsync, only the first field of an entry if provided in entry.Values.

### Expected Behavior

Returned stream entries have all their fields parsed in the StreamEntry.Values property.

### Current Behavior

StreamEntry.Values always has one item

### Valkey GLIDE C# client version

Other

### Valkey GLIDE C# client version (Other)

1.1.0

### .NET version

Other

### .NET version (Other)

.NET 10

### Engine type and version

Valkey 9.0

### Engine type and version (Other)

_No response_

### Engine Environment

Local

### Engine Environment (Other)

_No response_

### OS

Other

### OS (Other)

Linux

### Logs

:white_circle: None

### Cluster information

:white_circle: None

### Reproduction Steps

1. Add an entry with 2+ fields
2. Create stream group
3. Read stream group

```cs

using Valkey.Glide;
using Valkey.Glide.Commands.Options;
using static Valkey.Glide.ConnectionConfiguration;

var config = new StandaloneClientConfigurationBuilder()
  .WithAddress("localhost", 6379)
  .Build();

await using var client = await GlideClient.CreateClient(config);

var key = "this_is_a_key";
var group = "bar";
var consumer = "baz";

await client.StreamAddAsync(key, [new("a", "0"), new("b", "1")]);

try
{
  var res = await client.CustomCommand(["XGROUP", "CREATE", key, group, "0", "MKSTREAM"]);
}
catch { }


var entries = await client.StreamReadGroupAsync(new StreamPosition(key, ">"), group, consumer, new StreamReadGroupOptions { Block = TimeSpan.FromSeconds(5), Count = 5000 });
Console.WriteLine(entries.Length);
foreach (var entry in entries)
{
  Console.WriteLine(entry.Id);
  Console.WriteLine(entry.Values.Length);
  Console.WriteLine(string.Join('\n', entry.Values.Select(x => $"{x.Name}:{x.Value}")));
  await client.CustomCommand(["XACK", key, group, entry.Id]);
}

```

Output is:

```
1
1782127177990-0
1
a:0
```
Stating 1 entry on the stream, with 1 field only

### Possible Solution

https://github.com/valkey-io/valkey-glide-csharp/blob/28ecc0635d7c2ecd8e17abe2cb7dd7819d2ebeef/sources/Valkey.Glide/Internals/Request.StreamCommands.cs#L161

Maybe the outerArray isn't an array with a single element

### Related

:white_circle: None

### Additional Information

:white_circle: None

currantw:
## Summary

Fix XREADGROUP multi-field parsing bug, add comprehensive StackExchange.Redis integration test and command unit tests, and simplify stream response converters by removing dead RESP2 code paths that glide-core already normalizes.

## Issue Link

Closes #430.

:white_circle: None — this is a bug fix with no public API changes.

## Implementation

- **Bug fix**: Iterate all field-value pairs in the entry array instead of only taking `outerArray[0]`. Affects both single-stream and multi-stream `XREADGROUP` responses.
- **Converter refactoring**:
  - Extracted `ConvertStreamEntryMap` as the shared parser for the normalized `{entry_id => [[field, value], ...]}` map format.
  - `ConvertSingleStreamReadResponse` now delegates to `ConvertMultiStreamReadResponse`.
  - Removed dead RESP2 array handling and dead XAUTOCLAIM fallback — glide-core normalizes these to maps.
- **Testability**: Exposed stream converters and `ParseMemoryStats` as `internal` for direct unit testing without going through `.Converter`.

## Limitations

None

## Testing

- 7 new SER integration tests for `StreamReadGroupAsync` (error cases, single/multi stream, count, noAck, position).
- 2 new unit tests for stream read response converters (`StreamRead_SingleConverter`, `StreamRead_MultiConverter`).
- 3 new unit tests for multi-stream XREADGROUP argument building.
- All 549 unit tests pass. All 238 stream integration tests pass.

## Related Issues

- #431 — Original PR from @fs-paul (this supersedes it)
- #326 — Tracking issue for GLIDE-native stream command interfaces

## Checklist

### DIFF
--- a/sources/Valkey.Glide/Internals/Request.ServerManagement.cs
+++ b/sources/Valkey.Glide/Internals/Request.ServerManagement.cs
@@ -22,7 +22,7 @@ internal partial class Request
 
     private const string MemoryStatsDbPrefix = "db.";
 
-    private static MemoryStats ParseMemoryStats(Dictionary<GlideString, object> map)
+    internal static MemoryStats ParseMemoryStats(Dictionary<GlideString, object> map)
     {
         Dictionary<int, MemoryStatsDb> db = [];
         foreach (KeyValuePair<GlideString, object> kvp in map)
--- a/sources/Valkey.Glide/Internals/Request.StreamCommands.cs
+++ b/sources/Valkey.Glide/Internals/Request.StreamCommands.cs
@@ -26,12 +26,12 @@ internal partial class Request
     }
 
     public static Cmd<object, StreamEntry[]> StreamReadAsync(StreamPosition position, StreamReadOptions options)
-        => StreamReadAsyncWithOptions([position], options, ConvertSingleStreamRead);
+        => new(RequestType.XRead, BuildStreamReadArgs([position], options), false, ConvertSingleStreamReadResponse, allowConverterToHandleNull: true);
 
-    public static Cmd<object, ValkeyStream[]> StreamReadAsync(IEnumerable<StreamPosition> streamPositions, StreamReadOptions options)
-        => StreamReadAsyncWithOptions([.. streamPositions], options, ConvertMultiStreamRead);
+    public static Cmd<object, ValkeyStream[]> StreamReadAsync(IEnumerable<StreamPosition> positions, StreamReadOptions options)
+        => new(RequestType.XRead, BuildStreamReadArgs([.. positions], options), false, ConvertMultiStreamReadResponse, allowConverterToHandleNull: true);
 
-    private static Cmd<object, TResult> StreamReadAsyncWithOptions<TResult>(StreamPosition[] streamPositions, StreamReadOptions options, Func<object, TResult> converter)
+    private static GlideString[] BuildStreamReadArgs(StreamPosition[] positions, StreamReadOptions options)
     {
         List<GlideString> args = [];
 
@@ -48,16 +48,16 @@ internal partial class Request
         }
 
         args.Add(ValkeyLiterals.STREAMS);
-        foreach (var sp in streamPositions)
+        foreach (var sp in positions)
         {
             args.Add(sp.Key.ToGlideString());
         }
-        foreach (var sp in streamPositions)
+        foreach (var sp in positions)
         {
             args.Add(sp.Position.ToGlideString());
         }
 
-        return new(RequestType.XRead, [.. args], false, converter, allowConverterToHandleNull: true);
+        return [.. args];
     }
 
     public static Cmd<object, StreamEntry[]> StreamRangeAsync(ValkeyKey key, StreamRangeOptions options)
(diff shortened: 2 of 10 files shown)


sources/Valkey.Glide/Internals/Request.ServerManagement.cs
```diff
@@ -22,7 +22,7 @@ public static Cmd<Dictionary<GlideString, object>, MemoryStats> MemoryStatsAsync
 
     private const string MemoryStatsDbPrefix = "db.";
 
-    private static MemoryStats ParseMemoryStats(Dictionary<GlideString, object> map)
+    internal static MemoryStats ParseMemoryStats(Dictionary<GlideString, object> map)
```
currantw, line 25: Just exposed so it can be tested directly. No other changes, just for consistency.

tests/Valkey.Glide.UnitTests/CommandTests.cs
```diff
@@ -1180,7 +1244,7 @@ public void MemoryStatsConverter_WithAllFields()
             ["overhead.db.hashtable.rehashing"] = ConvertLong,
         };
 
-        var stats = Request.MemoryStatsAsync().Converter(raw);
+        var stats = Request.ParseMemoryStats(raw);
```
currantw, line 1247: Exposed ParseMemoryStats directly so easier testing.

sources/Valkey.Glide/Internals/Request.StreamCommands.cs
currantw, line 1: Consolidated a lot of unnecessary and duplicated response parsing.

sources/Valkey.Glide/Internals/Request.StreamCommands.cs
```diff
@@ -961,37 +794,69 @@ private static StreamInfo ConvertStreamInfo(object response)
...
+                {
+                    continue;
+                }
+                values[i] = new NameValueEntry(
+                    (GlideString)fieldValues[0],
+                    (GlideString)fieldValues[1]
+                );
+            }
+
+            entries.Add(new StreamEntry(entryId, values));
```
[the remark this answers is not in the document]
currantw: The flattened format is intentionally not supported here — glide-core normalizes XREAD/XREADGROUP/XRANGE/XCLAIM/XAUTOCLAIM responses to `ArrayOfPairs` format (`[[field, value], ...]`), so the flat format never reaches this converter. The `continue` leaving default `NameValueEntry` entries is a valid point though — it would only happen with malformed data from glide-core, which shouldn't occur, but switching to a `List<NameValueEntry>` would be more defensive. I'll make that change.

tests/Valkey.Glide.IntegrationTests/StackExchange/StreamCommandTests.cs
```diff
@@ -121,27 +118,111 @@ public async Task StreamConsumerGroupSetPositionAsync_ReturnsBool(IDatabaseAsync
...
+        var streams = await db.StreamReadGroupAsync([new StreamPosition(key1, ">"), new StreamPosition(key2, ">")], "mygroup", "consumer1");
+        Assert.Equal(2, streams.Length);
+
+        var stream0 = streams[0];
+        Assert.Equal(stream0.Key, key1);
+        Assert.Equivalent(new NameValueEntry[] { new("f1", "v1"), new("f2", "v2") }, Assert.Single(stream0.Entries).Values);
+
+        var stream1 = streams[1];
+        Assert.Equal(stream1.Key, key2);
+        Assert.Equivalent(new NameValueEntry[] { new("f3", "v3"), new("f4", "v4") }, Assert.Single(stream1.Entries).Values);
```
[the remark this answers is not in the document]
currantw: Fixed in 48055530

sources/Valkey.Glide/Internals/Request.StreamCommands.cs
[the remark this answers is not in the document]
currantw: Fixed in e386040c.

sources/Valkey.Glide/Internals/Request.StreamCommands.cs
```diff
@@ -48,16 +48,16 @@ private static Cmd<object, TResult> StreamReadAsyncWithOptions<TResult>(StreamPo
...
-        foreach (var sp in streamPositions)
+        foreach (var sp in positions)
         {
             args.Add(sp.Key.ToGlideString());
         }
-        foreach (var sp in streamPositions)
+        foreach (var sp in positions)
         {
             args.Add(sp.Position.ToGlideString());
         }
```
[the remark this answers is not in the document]
currantw: Fixed in 1014e971.

sources/Valkey.Glide/Internals/Request.StreamCommands.cs
```diff
@@ -370,11 +169,45 @@ public static Cmd<object, bool> StreamCreateConsumerAsync(ValkeyKey key, ValkeyV
...
+
+        args.Add(ValkeyLiterals.STREAMS);
+        foreach (var sp in positions)
+        {
+            args.Add(sp.Key);
+        }
+        foreach (var sp in positions)
+        {
+            args.Add(sp.Position);
+        }
```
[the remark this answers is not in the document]
currantw: Fixed in 1014e971.

sources/Valkey.Glide/Internals/Request.StreamCommands.cs
```diff
@@ -370,11 +169,45 @@ public static Cmd<object, bool> StreamCreateConsumerAsync(ValkeyKey key, ValkeyV
...
+
+        if (options.NoAck)
+        {
+            args.Add(ValkeyLiterals.NOACK);
+        }
+
+        args.Add(ValkeyLiterals.STREAMS);
+        foreach (var sp in positions)
+        {
+            args.Add(sp.Key);
```
prateek-kumar-improving, line 202: Why does this not need `sp.Key.ToGlideString()`? This is used in the same file around line 57.
Same for the loop below.

sources/Valkey.Glide/Internals/Request.StreamCommands.cs
```diff
@@ -370,11 +169,45 @@ public static Cmd<object, bool> StreamCreateConsumerAsync(ValkeyKey key, ValkeyV
...
+        }
+        foreach (var sp in positions)
```
prateek-kumar-improving, line 204: Can we combine this loop with the loop above and `Add` both values in the same loop?

sources/Valkey.Glide/Internals/Request.StreamCommands.cs
```diff
@@ -961,37 +794,68 @@ private static StreamInfo ConvertStreamInfo(object response)
...
+            {
+                continue;
+            }
+
+            var entryId = entryKvp.Key;
+
+            var values = new List<NameValueEntry>();
+            for (int i = 0; i < outerArray.Length; i++)
+            {
+                if (outerArray[i] is object[] fieldValues && fieldValues.Length >= 2)
```
prateek-kumar-improving: Why is this check required?
If it is not true, it will return empty `values` List. Should it not throw an error instead of returning empty response? 
Maybe in an ideal scenario this case might not occur. If that is the case then it should be an error, right? Basically indicating something has changed on the server end.

prateek-kumar-improving:
Reviewed till this commit: https://github.com/valkey-io/valkey-glide-csharp/pull/469/changes/1014e9715510b53098d30e7effdece2fe99a4c72

sources/Valkey.Glide/Internals/Request.StreamCommands.cs
currantw: `ValkeyKey` (like `sp.Key` here) and `ValkeyValue` (like `sp.Position` below) both have implicit casts to `GlideString`, so this isn't necessary. I did a much broader cleanup in #464 to remove unnecessary `ToGlideString()` calls.

sources/Valkey.Glide/Internals/Request.StreamCommands.cs
currantw: I don't think so, because we need to add all the keys _then_ all the positions (e.g. `[ key1, key2, ..., pos1, pos2]`) _not_ than key/position pairs (e.g `[ key1, pos1, key2, pos2, ...]`).

It would be possile to use `Select` to make this shorter, which I considered:

```csharp
args.AddRange(positions.Select(sp => sp.Key));
args.AddRange(positions.Select(sp => sp.Position));
```

But (a) I don't think that actually any easier to read, and (b) it is less efficient because it requires more internal allocation.

sources/Valkey.Glide/Internals/Request.StreamCommands.cs
currantw: Good cal! Not required. Removed in 157103b1.

prateek-kumar-improving (approved):

affonsov (approved):

Issue #472, opened 2026-07-23
task(ffi): Auto-generate RequestType enum from Rust source


currantw:
### Description

The `RequestType` enum in `sources/Valkey.Glide/Internals/FFI.structs.cs` is maintained manually as a 1:1 copy of the Rust enum in `valkey-glide/glide-core/src/request_type.rs`. When the submodule is updated with new commands, the C# enum must be updated by hand, which is error-prone and easy to forget.

Add a code generation script that parses the Rust source and emits the C# enum, eliminating manual synchronization. The Rust enum uses a rigid format (`VariantName = 123,`) that is trivial to parse with regex.

**Proposed approach:**

- Add a Python script in `dev/scripts/` that reads `request_type.rs` and generates `RequestType.g.cs` (or updates the existing section in `FFI.structs.cs`).
- Expose it as an on-demand task command (e.g., `task codegen` or `task codegen:ffi`).
- Add a CI check that regenerates and diffs against the committed file, failing if stale.

### Checklist

- [ ] Write Python script to parse `valkey-glide/glide-core/src/request_type.rs` and emit the C# `RequestType` enum
- [ ] Add a Taskfile command to run the generator
- [ ] Add CI step to verify the generated output is up-to-date
- [ ] Remove the manual `// TODO: generate this with a bindings generator` comment

### Related

- `sources/Valkey.Glide/Internals/FFI.structs.cs:604` (current TODO)
- `valkey-glide/glide-core/src/request_type.rs` (Rust source of truth)
- The Go wrapper uses `cbindgen` for similar generation (`valkey-glide/ffi/README.md`)

### Additional Information

- The `RequestType` enum is the simplest candidate for generation. Other structs in `FFI.structs.cs` have marshaling attributes and pointer fields that would require more complex generation logic — those can be tackled separately if this proves useful.
- The Rust enum is annotated with `#[repr(C)]`, so the discriminant values are stable across the FFI boundary.

Issue #414, opened 2026-05-26
feat(client): Add `CLIENT INFO` command support


currantw:
### Describe the feature

Add `CLIENT INFO` command support to the Valkey GLIDE C# client. The [`CLIENT INFO`](https://valkey.io/commands/client-info/) command returns information and statistics about the current client connection in a human-readable format. 

It is supported by Valkey GLIDE core (request type `ClientInfo`). However, the request is currently routed to a random node by default, so resolving this issue may also require a core change. This command is not supported by any other Valkey GLIDE client at the moment.

### Use Case

Users need to inspect the current connection's metadata (e.g., client ID, name, flags, database, subscriptions, connection age) for diagnostics, monitoring, and debugging purposes.

### Proposed Solution

- Implement GLIDE core changes as needed to update default routing.
- Add GLIDE-native `CLIENT INFO` methods to the client interfaces (`IBaseClient`, `IClusterClient`) and implement in `GlideClient` and `GlideClusterClient`.

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might cause a breaking change

### Related

- #277 — `CLIENT LIST` command support
- #275 — Connection management command API cleanup.
- #270 — Parent issue: command API cleanup.

### Additional Information

- `CLIENT INFO` was introduced in Redis 6.2 / Valkey 7.2.
- The command takes no arguments and returns a single line in the same format as `CLIENT LIST`.
- The glide-core request type is `ClientInfo = 306`.

currantw:
There are various complication caused by adding cluster routing for `CLIENT INFO`. It is unclear what the desired behaviour is from a user perspective. Based on discussion with @jeremyprime, we will defer this until there is a more clear need for it. Moving to backlog.

currantw:
Based on work completed on this command before it was moved to backlog, there are proposed interfaces on [this branch](https://github.com/currantw/valkey-glide-csharp/tree/414-client-info) that may be useful as a starting point for future assignee. In particular, the `ClientInfo` type and related enums may be helpful.

PR #473, opened 2026-07-23
chore: Triage and clean up TODO comments

### LINKED ISSUE #325: chore: Triage and clean up TODO comments across the codebase
### Summary

The codebase has accumulated ~45 TODO comments across C#, Rust FFI, nuspec, and utility files (excluding the read-only `valkey-glide/` submodule). These should be triaged: resolved, linked to tracking issues, or removed.

### Description

A sweep of the repository found TODO comments in the following categories:

- **Already linked to issues** (~17): References to #269, #205, #287, #322, #280. Verify still relevant; remove if resolved.
- **Documentation** (~2): Missing pubsub docs on client classes, missing wiki HelpLink.
- **Code cleanup / refactoring** (~8): Vague cleanup markers, visibility concerns, and minor improvements.
- **FFI / Internals** (~3): Bindings generator, missing config params, incomplete FFI notes.
- **NuSpec packaging** (~5): Missing metadata fields, changelog, conditional binary builds.
- **Tests** (~6): Stale test code, missing assertions, parallelization notes.
- **Utility script** (~1): License parsing logic improvement.

### Proposed Solution

For each TODO:

1. If it references an existing open issue, verify it's still relevant. Keep or remove accordingly.
2. If it's a small, actionable fix, resolve it directly.
3. If it's a larger effort, create a dedicated issue and update or remove the TODO.
4. If it's stale or no longer applicable, remove it.

### Effort & Severity Estimates

- Effort: medium
- Severity: low

### Related

- #269, #205, #287, #322, #280

currantw:
## Summary

Triage and clean up TODO comments across the codebase. Adds a `check-todos` script and CI workflow to enforce that all TODOs reference an open GitHub issue with a description, then resolves or annotates all existing TODOs.

## Issue Link

Closes #325.

### TODO checker infrastructure

- Add `dev/scripts/check_todos.py` — validates that every TODO comment follows the format `TODO #<number>: <description>`.
  - Description must be at least 10 characters.
  - Referenced issue must be open.
- Add `dev/conf/check-todos-ignore` — fnmatch-based ignore file for paths excluded from validation.
- Add `.github/workflows/check-todos.yml` — CI workflow that runs the checker on PRs and pushes.
- Add `task check:todos` command.
- Document the TODO format in `DEVELOPER.md`.

### TODO comment fixes and removals

- **`Request.StreamCommands.cs`** — Remove SER-specific `StreamTrimAsync` method; replace with GLIDE-style method accepting `StreamTrimOptions`. SER `Database.StreamCommands` now routes through the `BaseClient.StreamTrimAsync` GLIDE interface method.
- **`IDatabaseAsync.StreamCommands.cs`** — Fix `StreamTrimAsync` signature to use non-nullable `long maxLength` (matching StackExchange.Redis 2.8.58).
- **`IBaseClient.StreamCommands.cs` / `BaseClient.StreamCommands.cs`** — Uncomment and implement `StreamTrimAsync(ValkeyKey, StreamTrimOptions)`.
- **`ScanOptions.cs`** — Add `seealso` links for SCAN, SSCAN, and ZSCAN commands.
- **`Request.cs`** — Extract shared `ToValkeyKeyArray` and `ToValkeyValueArray` collection converters.
- **`Request.SetCommands.cs`** — Extract `ConvertSetScanResponse` for SSCAN response parsing.
- **`Request.SortedSetCommands.cs`** — Extract `ConvertSortedSetScanResponse` and `ToSortedSetEntriesFromFlatArray` for ZSCAN response parsing; remove old `ToValkeyValues` delegate.
- **`Request.GenericCommands.cs`** — Extract `ConvertClusterScanResponse` for top-level SCAN (string cursor for cluster compatibility).
- **`GlideClusterClient.cs`** — Refactor `ClusterScanCommand` to accept `ClusterScanCursor` instead of raw string; simplify `ScanAsync` loop.
- **`GlideString.cs`** — Replace round-trip UTF-8 encode/decode with `Utf8.IsValid` (net8.0 API, allocation-free).
- **`Helpers.cs`** — Remove dead code (`DownCastVals`, `ConvertKeyValuePairsToArray`).
- **`GlideClient.cs` / `GlideClusterClient.cs`** — Remove `// TODO add pubsub` comments; docs already updated in prior commits.
- **`ConnectionMultiplexer.cs`** — Remove TODO referencing closed upstream issue (valkey-io/valkey-glide#4293).
- **`FFI.structs.cs`** — Replace vague TODO with issue reference (`TODO #472`).
- **`Valkey.Glide.nuspec`** — Resolve all packaging TODOs: add `<releaseNotes>`, `commit` attribute, remove stale comments.
- **`cd.yml`** — Pass `commit=${{ github.sha }}` to nuspec via `NuspecProperties`.
- **Integration/unit tests** — Remove resolved TODO skip markers; update `CommandTests` for new `StreamTrimAsync` signature.

### Cleanup

- Move `dev/coverage/coverage-baseline.json` to `dev/conf/coverage-baseline.json`.
- Move `dev/coverage/.runsettings` to `dev/conf/coverage-runsettings.xml`.
- Remove empty `dev/coverage/` directory (output subdirectories are recreated at runtime).
- Update all references in CI workflows, scripts, and documentation.

## Implementation

The main design decisions:

1. **Stream trim refactoring**: SER methods now build `StreamTrimOptions` (MaxLen or MinId) and call through the GLIDE `BaseClient.StreamTrimAsync` method, eliminating direct `Request` calls from the SER layer.
2. **Scan response converters**: Each scan command (SCAN, SSCAN, ZSCAN) has its own dedicated converter method rather than a shared generic — simpler to read and avoids unnecessary indirection.
3. **TODO checker**: Uses `git grep` for discovery, `gh issue view` for validation, and fnmatch patterns for exclusions. Fails hard if the ignore file is missing.

## Limitations

:white_circle: None

## Testing

- All 547 unit tests pass.
- Integration tests not run locally (require a running Valkey instance).
- TODO checker validates all 25 remaining TODOs pass format and issue checks.

## Related Issues

- #472 — Auto-generate RequestType enum (created as part of this work)

## Checklist

- [x] ~~`CHANGELOG.md`, `README.md`, `DEVELOPER.md`, and other documentation files are updated.~~
- [x] All TODO comments referencing issue(s) closed by this PR have been resolved.


sources/Valkey.Glide/Abstract/ValkeyServer.cs
```diff
@@ -288,42 +288,17 @@ public async Task ScriptFlushAsync(CommandFlags flags = CommandFlags.None)
...
-            }
-
-            foreach (ValkeyKey key in keys)
-            {
-                yield return key;
-            }
-
-        } while (currentCursor != "0");
+        var options = new ScanOptions { MatchPattern = pattern, Count = pageSize };
+        return _conn.ScanAsync(options).SkipAsync(pageOffset);
```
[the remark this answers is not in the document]
currantw: Addressed in 6260c494

sources/Valkey.Glide/Abstract/Database.SetCommands.cs
```diff
@@ -138,47 +138,18 @@ public Task<bool> SetMoveAsync(ValkeyKey source, ValkeyKey destination, ValkeyVa
...
-            foreach (ValkeyValue element in elementsToYield)
-            {
-                yield return element;
-            }
-
-            currentCursor = nextCursor;
-            currentOffset = 0; // Only skip on first iteration
-        } while (currentCursor != 0);
+        var options = new ScanOptions { MatchPattern = pattern, Count = pageSize };
+        return SetScanAsync(key, options).SkipAsync(pageOffset);
```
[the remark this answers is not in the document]
currantw: Addressed in 6260c494

sources/Valkey.Glide/Abstract/Database.SortedSetCommands.cs
```diff
@@ -291,48 +291,18 @@ public Task<long> SortedSetRemoveRangeByScoreAsync(ValkeyKey key, double start,
...
-            foreach (SortedSetEntry element in elementsToYield)
-            {
-                yield return element;
-            }
-
-            currentCursor = nextCursor;
-            currentOffset = 0; // Only skip on first iteration
-        } while (currentCursor != 0);
+        var options = new ScanOptions { MatchPattern = pattern, Count = pageSize };
+        return SortedSetScanAsync(key, options).SkipAsync(pageOffset);
```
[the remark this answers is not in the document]
currantw: Addressed in 6260c494

sources/Valkey.Glide/Abstract/Database.StreamCommands.cs
```diff
@@ -200,25 +200,38 @@ public Task<StreamAutoClaimJustIdResult> StreamAutoClaimIdsOnlyAsync(ValkeyKey k
 
     /// <inheritdoc cref="IDatabaseAsync.StreamTrimAsync(ValkeyKey, int, bool, CommandFlags)"/>
     public Task<long> StreamTrimAsync(ValkeyKey key, int maxLength, bool useApproximateMaxLength = false, CommandFlags flags = CommandFlags.None)
-    {
-        GuardClauses.ThrowIfCommandFlags(flags);
-        return Command(Request.StreamTrimAsync(key, maxLength, default, useApproximateMaxLength, null));
-    }
+        => StreamTrimAsync(key, maxLength, useApproximateMaxLength);
```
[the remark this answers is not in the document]
currantw: Addressed in 276a7fdf

currantw:
Sample output from recent run:

```text
task: [check:todos] python3 dev/scripts/check_todos.py
Checking TODO comments...

  FAIL  sources/Valkey.Glide/Client/BaseClient.StreamCommands.cs:98
        missing description (expected format: TODO #<number>: <description>)

Checked 27 TODOs: 26 passed, 1 failed.
task: Failed to run task "check:todos": exit status 1
Error: exit status 1

Error: Process completed with exit code 201.
```

.github/workflows/cd.yml
```diff
@@ -158,8 +158,12 @@ jobs:
...
               working-directory: sources/Valkey.Glide
               run: |
-                  dotnet build Valkey.Glide.csproj --configuration Release "/property:Version=$RELEASE_VERSION"
-                  dotnet pack Valkey.Glide.csproj --configuration Release "-p:NuspecProperties=version=$RELEASE_VERSION"
+                  dotnet build Valkey.Glide.csproj \
+                    --configuration Release \
+                    "/property:Version=$RELEASE_VERSION"
+                  dotnet pack Valkey.Glide.csproj \
+                    --configuration Release \
+                    "-p:NuspecProperties=version=$RELEASE_VERSION;commit=${{ github.sha }}"
```
currantw, line 166: Added the GitHub commit SHA to the package info. Otherwise just multi-line for readability.

dev/conf/coverage-baseline.json
currantw, line 1: Moved to new `dev/conf` directly to keep these configuration files all together

dev/scripts/get_licenses_from_ort.py
```diff
@@ -10,7 +10,6 @@
 The script outputs a set of licenses identified by the analyzer. GLIDE maintainers should review the returned list to ensure that all licenses are approved.
 """
 
-# TODO: Modify to use logic operations instead of including AND and OR in strings
```
currantw, line 13: Don't think this is realistic or particularly desirable 🤷

sources/Valkey.Glide/Abstract/ConnectionMultiplexer.cs
```diff
@@ -121,8 +121,6 @@ public IServer GetServer(EndPoint endpoint)
         throw new ArgumentException("The specified endpoint is not defined", nameof(endpoint));
     }
 
-    // TODO currently this returns only primary node on standalone
-    // https://github.com/valkey-io/valkey-glide/issues/4293
```
currantw, line 125: This issue is closed as will not do.

sources/Valkey.Glide/Client/GlideClient.cs
```diff
@@ -18,42 +18,12 @@ public partial class GlideClient :
     BaseClient,
     IGlideClient
 {
-    internal GlideClient() { }
+    #region Public Methods
 
-    // TODO add pubsub and other params to example and remarks
     /// <summary>
```
currantw, line 24: All of this is documented elsewhere, and there are now plenty of these examples of the docs site. Don't need to duplicate that all here 🤷

sources/Valkey.Glide/Internals/Helpers.cs
```diff
@@ -10,11 +10,6 @@ internal static class Helpers
     public static Dictionary<string, T> DownCastKeys<T>(this Dictionary<GlideString, T> dict)
         => dict.Select(p => (Key: p.Key.ToString(), p.Value)).ToDictionary(p => p.Key, p => p.Value);
 
-    // TODO make recursive?
-    // Downcast dictionary values from `GlideString` to `string`
-    public static Dictionary<T, string> DownCastVals<T>(this Dictionary<T, GlideString> dict) where T : class
```
currantw, line 15: Unused. Same with `ConvertKeyValuePairsToArray` below.

tests/Valkey.Glide.IntegrationTests/StandaloneClientTests.cs
```diff
@@ -67,8 +67,6 @@ public async Task CanConnectWithDifferentParameters()
 
     [Theory(DisableDiscoveryEnumeration = true)]
     [MemberData(nameof(Config.TestStandaloneClients), MemberType = typeof(TestConfiguration))]
-    // Verify that client can handle complex return types, not just strings
-    // TODO: remove this test once we add tests with these commands
```
currantw, line 71: This test is still valuable for testing custom commands.

AGENTS.md
```diff
@@ -40,16 +40,6 @@ Common commands:
...
 
-## Lint and Format Rules (Agents)
-
-- Prefer `task` commands for linting and formatting.
-- Lint checks (read-only, fail on issues):
-  - `task lint` (all checks), `task lint:rust`, `task lint:csharp`, `task lint:yaml`, `task lint:actions`, `task lint:markdown`
-- Auto-fix formatting:
-  - `task format` (all languages), `task format:rust`, `task format:csharp`, `task format:yaml`, `task format:markdown`
-- Link checking (separate from lint, slower):
-  - `task check-links`
```
currantw, line 51: Consollidated to DEVELOPER.md and Quality Gates checklist below.

Taskfile.yml
```diff
@@ -126,15 +126,15 @@ tasks:
 
     lint:yaml:
         desc: "Run YAML linting"
-        cmd: npx prettier --check "**/*.yml"
+        cmd: npx --yes prettier --check "**/*.yml"
```
currantw, line 129: Allows auto install if package missing.

dev/scripts/check_todos.py
```diff
@@ -0,0 +1,161 @@
...
+    file: str
+    line: int
+    text: str
+
+
+# Matches TODO and captures GitHub issue ID and description.
+_TODO_PATTERN = re.compile(
+    r"TODO\b(\s+#(?P<github_id>\d+)(:\s+(?P<description>.+))?)?",
+    re.IGNORECASE,
+)
```
xShinnRyuu: [BLOCKING] F1: regex has no left-side word boundary, so it false-positives on identifiers ending in `todo`.

The pattern is `TODO\b(...)` with `re.IGNORECASE`, then piped to `git grep -i -P`. Reproduced in an isolated repo: `public class TodoList`, `public string SomeTodo;`, `public void HandleTodoAction()`, `var listTodo = 1;`, `wide_todo = 1`, and a plain `// a todo entry` all get flagged as "missing GitHub issue reference". The current tree escapes only because the two files that legitimately contain `_Todo` and `_TODO_PATTERN` are hard-coded in the ignore list and no such identifier lives in the source tree today.

The first future contributor who names a type or field with a trailing `Todo` will hit a confusing CI failure on a symbol that has nothing to do with a TODO.

Suggested fix: anchor both sides.

```python
_TODO_PATTERN = re.compile(
    r"\bTODO\b(\s+#(?P<github_id>\d+)(:\s+(?P<description>.+))?)?",
    re.IGNORECASE,
)
```

Verified locally: `\bTODO\b` still matches `// TODO #1: x` and `#pragma ... // TODO #475: ...`, and rejects `SomeTodo`, `aTodo`, `wide_todo`. Also worth considering: drop `re.IGNORECASE` if the enforced spelling is uppercase only, since the case-insensitive match is what makes the underscore-and-camelCase collision bite.

sources/Valkey.Glide/Errors.cs
```diff
@@ -164,8 +164,6 @@ public ConnectionException(string message, Exception innerException) : base(mess
     /// </summary>
     public sealed class ConfigurationError : GlideException
     {
-        // TODO set HelpLink with link to wiki
```
xShinnRyuu, line 167: [POLISH] F4: `// TODO set HelpLink with link to wiki` was removed here but no `HelpLink` is set anywhere in `Errors.cs` and no issue captures the concern. Same silent-drop shape as the rust config comment (F3), lower severity because it is cosmetic.

Suggested fix: either file a small issue and add `// TODO #<n>: Set HelpLink for each exception with a wiki URL.` here, or note in the PR body that HelpLink is deliberately dropped.

rust/src/ffi.rs
```diff
@@ -148,11 +148,6 @@ pub struct ConnectionConfig {
     pub node_discovery_mode: NodeDiscoveryMode,
     pub has_client_side_cache_config: bool,
     pub client_side_cache_config: ClientSideCacheConfig,
-    /*
-    TODO below
-    pub periodic_checks: Option<PeriodicCheck>,
-    pub inflight_requests_limit: Option<u32>
-    */
```
xShinnRyuu, line 155: [BLOCKING] F3: this deletion, together with the `// TODO more config params, see ffi.rs` deletion in `sources/Valkey.Glide/Internals/FFI.structs.cs`, silently drops a live reminder about missing config surface.

The `periodic_checks: Option<PeriodicCheck>` and `inflight_requests_limit: Option<u32>` fields still ship as hardcoded `None` at `rust/src/ffi.rs:373-374`:

```rust
periodic_checks: None,
inflight_requests_limit: None,
```

The C# `ConnectionConfig` has no surfacing of them. Neither the PR body's "Related Issues" nor the diff creates a GitHub issue for this gap. This is exactly the failure mode #325 sought to avoid: a real reminder deleted because it did not match the enforced format, leaving the concern uncaptured.

Suggested fix: file a `feat(config): expose periodic_checks and inflight_requests_limit through the C# ConnectionConfig` issue and replace the deleted C# comment with `// TODO #<n>: Expose periodic_checks and inflight_requests_limit config options.` on the `ConnectionConfig` struct in `FFI.structs.cs`. If the team decided these should not ship, please say so in the PR body.

sources/Valkey.Glide/Abstract/IDatabaseAsync.StreamCommands.cs
```diff
@@ -344,14 +344,24 @@ public partial interface IDatabaseAsync
...
     /// <param name="trimMode">The trim mode to use.</param>
     /// <returns>The number of entries removed.</returns>
-    Task<long> StreamTrimAsync(ValkeyKey key, long? maxLength = null, bool useApproximateMaxLength = false, long? limit = null, StreamTrimMode trimMode = StreamTrimMode.KeepReferences, CommandFlags flags = CommandFlags.None);
+    Task<long> StreamTrimAsync(
+        ValkeyKey key,
+        long maxLength,
+        bool useApproximateMaxLength = false,
+        long? limit = null,
+        StreamTrimMode trimMode = StreamTrimMode.KeepReferences,
+        CommandFlags flags = CommandFlags.None);
```
xShinnRyuu, line 375: [BLOCKING] F5: this signature changes `Task<long> StreamTrimAsync(ValkeyKey key, long? maxLength = null, ...)` to `Task<long> StreamTrimAsync(ValkeyKey key, long maxLength, ...)`. Two things change together: `long?` becomes `long`, and the default disappears.

Any caller compiling `db.StreamTrimAsync(key)` (no maxLength) will no longer compile, and callers using named arg `maxLength: null` are broken at compile time. This is a source-breaking change to `IDatabaseAsync` on the SER-compatibility surface. Defensible on SER-parity grounds (SER's `StreamTrimAsync` takes an `int maxLength` with no default), but the PR body only mentions it in passing.

Suggested fix: confirm on record that this is acceptable, add a CHANGELOG entry (see summary), and consider keeping a deprecated `long?` overload for one release if any downstream consumer was relying on the pre-1.2 shape.

sources/Valkey.Glide/Abstract/Database.SetCommands.cs
```diff
@@ -138,47 +138,18 @@ public Task<bool> SetMoveAsync(ValkeyKey source, ValkeyKey destination, ValkeyVa
...
-
-            foreach (ValkeyValue element in elementsToYield)
-            {
-                yield return element;
-            }
-
-            currentCursor = nextCursor;
-            currentOffset = 0; // Only skip on first iteration
-        } while (currentCursor != 0);
+        var options = new ScanOptions { MatchPattern = pattern, Count = pageSize };
```
xShinnRyuu, line 152: [POLISH] F6/F7: this line and the two parallel sites (`Database.SortedSetCommands.cs:304`, `ValkeyServer.cs:294`) unconditionally set `Count = pageSize`. `ScanOptions.ToArgs()` emits `COUNT <value>` whenever `Count.HasValue`, so passing `pageSize = 0` or a negative sends invalid `COUNT` to the server. The prior code had `if (pageSize > 0)` (KeysAsync) or `if (pageSize != 250)` (Set/SortedSet) guards.

Copilot's four review threads (3639425667/725/758/781) flagged exactly this, and the author replied "Addressed in 6260c494" (3639798460/635/797), but `6260c49` is not in the current branch history and the guard is still missing at HEAD `5e75e2d`. Those threads are also not marked resolved on GitHub, so downstream reviewers may believe this was fixed.

Suggested fix:

```csharp
var options = new ScanOptions { MatchPattern = pattern };
if (pageSize > 0) options.Count = pageSize;
```

Apply to all three call sites. If the intent is that only the default (250) still hits the server, restore `if (pageSize != 250)` at the Set/SortedSet paths.

Separately, this whole family of edits (SER Scan delegation, StreamTrim rewrite, `Request.ToScanArgs` -> `ScanOptions.ToArgs`, GlideString UTF-8 swap) is behavior-touching and does not really fit under a `chore: TODO cleanup` title. For future PRs, splitting into `chore: add TODO checker + format all TODOs` and `refactor: consolidate scan / stream-trim plumbing` would help reviewers know what to look for.

dev/scripts/check_todos.py
```diff
@@ -0,0 +1,161 @@
...
+        print(f"        {reason}\n")
+
+    passed = len(todos) - len(failures)
+    print(f"Checked {len(todos)} TODOs: {passed} passed, {len(failures)} failed.")
+
+    sys.exit(1 if failures else 0)
+
+
+if __name__ == "__main__":
+    main()
```
xShinnRyuu, line 162: [POLISH] F9: no unit test for this file. `dev/scripts/` has `check_examples.py`, `coverage.py`, `test.py`, `validate_examples.py`, `extract_examples.py`, but nothing for the new checker. The regex is subtle (see F1, F10) and the checker gates CI; a regression will silently change what the whole repo does.

Suggested fix: add a small pytest with sample TODO strings covering `// TODO #1: valid ten chars`, `// TODO fix later`, `// TODO #1`, `// TODO #1: <9 chars>`, `SomeTodo`, `wide_todo`, `URL /todo/1`, `<!-- TODO #1: html -->`, and `/* TODO ... */`. About 40 LOC, pins the regex behavior.

dev/scripts/check_todos.py
```diff
@@ -0,0 +1,161 @@
...
+    """A TODO found in the codebase."""
+
+    file: str
+    line: int
+    text: str
+
+
+# Matches TODO and captures GitHub issue ID and description.
+_TODO_PATTERN = re.compile(
+    r"TODO\b(\s+#(?P<github_id>\d+)(:\s+(?P<description>.+))?)?",
```
xShinnRyuu: [POLISH] F10: two format-drift issues with the current pattern.

1. `<!-- TODO #123: html body -->` matches with description captured as `html body -->` (trailing `-->` becomes part of the description). Same for `/* TODO #123: ... */` capturing trailing ` */`. Not in-tree today, but on the road the checker will accept HTML/block comments with the wrong description text.
2. `//TODO #414: Remove when ClientInfoAsync implemented.` at `tests/Valkey.Glide.IntegrationTests/ConnectionManagementCommandTests.cs:16` (no space after `//`) is accepted, but the DEVELOPER.md example shows `// TODO #<n>: ...` with a space. Style drift the sweep missed.

Suggested fix: tighten the description group to stop at closing `*/` or `-->`, e.g. `(?P<description>[^\n]+?)(?=\s*(?:-->|\*/)|$)`. Separately, normalize the one `//TODO` occurrence in the tree to `// TODO`.

dev/scripts/check_todos.py
```diff
@@ -0,0 +1,161 @@
...
+
+def _is_ignored(filepath: str, patterns: list[str]) -> bool:
+    """Check if a filepath matches any exclusion pattern."""
+    return any(fnmatch(filepath, p) for p in patterns)
+
+
+def _find_todos() -> list[_Todo]:
+    """Find all TODOs in tracked files using git grep."""
+    result = subprocess.run(
+        ["git", "grep", "-n", "-i", "-P", _TODO_PATTERN.pattern],
```
xShinnRyuu: [NIT] F11: `subprocess.run(["git", "grep", "-n", "-i", "-P", _TODO_PATTERN.pattern], ...)` passes the whole validation pattern (with the optional `#<num>` groups) to `git grep`. `git grep -P` interprets PCRE, so the current expression works, but discovery and validation would be safer as two separate regexes.

Suggested fix: split into `_TODO_DISCOVERY = r"\bTODO\b"` used by `git grep`, and `_TODO_PATTERN` used by `_validate_todos`. Any future edit to `_TODO_PATTERN` for validation semantics then does not silently mutate what `git grep` finds.

dev/scripts/check_todos.py
```diff
@@ -0,0 +1,161 @@
...
+def _load_ignore_patterns() -> list[str]:
+    """Load ignore patterns from the check-todos-ignore file."""
+    if not os.path.isfile(_IGNORE_FILE):
+        print(f"Error: ignore file not found: {_IGNORE_FILE}", file=sys.stderr)
+        sys.exit(1)
+
+    return [
+        ignore for line in Path(_IGNORE_FILE).read_text().splitlines()
+        if (ignore := line.strip()) and not ignore.startswith("#")
+    ]
```
xShinnRyuu, line 57: [NIT] F12: `not ignore.startswith("#")` filters full-line comments, but any inline `#` comment (`some/glob # my note`) becomes part of the glob and `fnmatch` will never match. Minor because no one uses this style today.

Suggested fix: either document "full-line comments only; no inline `#`" in the ignore file header, or strip inline comments before returning.

sources/Valkey.Glide/Client/BaseClient.SetCommands.cs
```diff
@@ -88,22 +90,26 @@ public async Task<ValkeyValue[]> SetRandomMembersAsync(ValkeyKey key, long count
...
+            (cursor, var elements) = await Command(Request.SetScanAsync(key, cursor, options));
+            foreach (var element in elements)
             {
                 yield return element;
             }
 
-            currentCursor = nextCursor;
-        } while (currentCursor != 0);
+        } while (cursor != 0);
     }
```
xShinnRyuu, line 115: [QUESTION] F14: this refactor exposes a new `protected async IAsyncEnumerable<ValkeyValue> SetScanAsync(ValkeyKey, long, ScanOptions?)` (and the parallel `SortedSetScanAsync(ValkeyKey, long, ScanOptions?)` in `BaseClient.SortedSetCommands.cs:257-269`). Since `BaseClient` is `public abstract`, `protected` members are part of the public API subclassers see. The prior code had these as `private`.

Any user who subclasses `BaseClient` now has an accessible-from-subclass API named identically to public overloads but with a `long cursor` parameter. If future SER-parity forces the cursor to become `object` or `ClusterScanCursor`, the protected shape leaks.

Question: is the `protected` intentional here (for cross-partial reuse), or should this be `private` / `private protected`?

currantw: Made `private protected` in 061a7f8c. Can't be `private` because they are used by `Database` sub-class.

dev/scripts/check_todos.py
currantw: (1) This seems like extra complexity without much benefit to me. The only purpose of capturing the description in the regex is to ensure that it is longer than some trivial length (currently ten chars), so that it can't just be something like "TODO: fix this". Including a few extra chars ins't a big deal – in addition to the fact that we don't even have any HTML docs in this repo anyway!

currantw: (2) Updated in ec9355d2

dev/scripts/check_todos.py
currantw: Split into `_TODO_VALIDATION_PATTERN` and `_TODO_VALIDATION_PATTERN` in 7ff432af.

dev/scripts/check_todos.py
currantw: I think this is overkill for now?

1. This is an internal dev tool, not part of production code. There are no existing tests for other dev scripts either.
2. We would need to do a bunch of work to integrate pytest into existing test commands, workflows, documentation, etc.

If we do ever end up using something like this in multiple repos, I certainly think this would make sense.

sources/Valkey.Glide/Abstract/Database.SetCommands.cs
currantw: Update in 43e18ec2 to add `GuardClauses.ThrowIfNotPositive`, which throws if pageOffset is non-positive. This matches the StackExchange.Redis behaviour. Previously, we silently ignored negative values.

sources/Valkey.Glide/Errors.cs
currantw: The wiki has been superseded by the docs site, and I don't think there is anything that we need to link here.

dev/scripts/check_todos.py
currantw: Fixed by previous changes. The discovery pattern (git grep) now uses `\bTODO\b` with word boundaries on both sides, so identifiers like `SomeTodo` or `wide_todo` are never matched. 

With regards to case, TODO comments are matched without regard to case (so that we catch "todo: xyz" for example), but they fail _validation_ if it isn't uppercase (so we should always eventually get comments in the "canonical" form "TODO: xyz"). Updated in 2361673a.

rust/src/ffi.rs
currantw: Raised issues for these and added TODO comments in b2e6bf13.

dev/scripts/check_todos.py
currantw: Added comment in. 7ce7aa3e

sources/Valkey.Glide/Abstract/IDatabaseAsync.StreamCommands.cs
currantw: The `long? maxLength = null` signature has been restored to avoid a source-breaking change in this PR.
 
The previous behavior when null was passed would have sent an invalid `XTRIM` key command to the server, resulting in a `RequestException`. We now throw `ArgumentNullException` earlier with a clearer message.

More importantly, the current signature does not match StackExchange.Redis! I raised #486 to fix this in release 2.0, and added TODO comments as appropriate.

dev/scripts/check_todos.py
```diff
@@ -0,0 +1,162 @@
...
+        if github_id not in checked_issues:
+            checked_issues[github_id] = _check_issue(github_id)
+        if checked_issues[github_id]:
+            failures[todo] = checked_issues[github_id]
+
+        # Check description length
+        description = match.group("description")
+        if len(description.strip()) < _MIN_DESCRIPTION_LENGTH:
+            failures[todo] = f"description too short (must be at least {_MIN_DESCRIPTION_LENGTH} characters)"
+            continue
```
xShinnRyuu: [NIT] Validation order lets a short-description failure overwrite an issue-state failure, and calls `gh issue view` for TODOs that will fail on description length anyway.

If a TODO has both a closed issue and a `<10`-char description, the loop first sets `failures[todo] = "#123 is not open ..."` at line 134, then reaches line 138-140 and overwrites it with `"description too short ..."`. The `continue` at 127 for format failure is correct; the two later checks lack the same guard.

Suggested fix: check description length before the issue call, and `continue` after each failing check.

```python
description = match.group("description")
if len(description.strip()) < _MIN_DESCRIPTION_LENGTH:
    failures[todo] = f"description too short (must be at least {_MIN_DESCRIPTION_LENGTH} characters)"
    continue

github_id = int(match.group("github_id"))
if github_id not in checked_issues:
    checked_issues[github_id] = _check_issue(github_id)
if checked_issues[github_id]:
    failures[todo] = checked_issues[github_id]
    continue
```

As a side benefit, this avoids a network call per TODO that already fails locally.

xShinnRyuu (approved):

dev/scripts/check_todos.py
currantw: Added `continue` in 5545a969. Thought order was fine – more readable to match actual reading order 🤷

jamesx-improving (commented):
Nice cleanup — check_todos + CI wiring is a good forcing function, and the SER→GLIDE stream-trim refactor reads well. One inline note on the SkipAsync change.

Happy to re-review once you've had a look.

sources/Valkey.Glide/Internals/Helpers.cs
```diff
@@ -49,19 +44,20 @@ public static string GetRealTypeName(this Type t)
     }
 
     /// <summary>
-    /// Converts a KeyValuePair array to an array of key-value pairs for use with MSet command.
+    /// Skips the first <paramref name="skip"/> elements from an async sequence.
     /// </summary>
-    /// <param name="values">The key-value pairs to convert.</param>
-    /// <returns>An array where keys and values are interleaved: [key1, value1, key2, value2, ...]</returns>
-    public static GlideString[] ConvertKeyValuePairsToArray(KeyValuePair<GlideString, GlideString>[] values)
+    public static async IAsyncEnumerable<T> SkipAsync<T>(this IAsyncEnumerable<T> source, int skip)
```
jamesx-improving, line 49: `SkipAsync` skips across all SCAN batches, but the interface docs for `SetScanAsync` / `SortedSetScanAsync` / `KeysAsync` say `pageOffset` is "the number of elements to skip from the **first** page" — the old code enforced that by resetting the local offset to 0 after the first iteration. When `pageOffset > firstBatchSize`, the two return different subsets of the key/set.

**Suggested fix:** either update the XML docs to document the new "skip across pages" semantics, or restore the first-page-only skip inside each scan method's iterator (apply `Skip(pageOffset)` only on iteration 1, then reset). Add a regression test where `pageOffset > pageSize` — the existing tests use `pageSize: 1000, pageOffset: 2` against ≤10 elements, so both implementations happen to return the same result.

currantw: I think the  current `SkipAsync` implementation is correct. `IAsyncEnumerable<T>` is a flat stream of elements — there's no concept of "pages" at the client level. `pageSize` is just a hint to the Valkey 	 command about internal batching, while `pageOffset` means "skip the first N elements from the overall result stream." SkipAsync does that.

currantw: Discussed offline. Addressed in e41257bf.

jamesx-improving: The updated docs + `pageOffset > pageSize` tests resolve this. Agreed the flat-stream semantics are the better long-term choice, and calling out the SER divergence in `<remarks>` is exactly right. Thanks for the follow-up.

jamesx-improving (approved):
Docs match the implementation now and the pageOffset > pageSize tests pin the choice — LGTM.

PR #483, opened 2026-07-27
Add circuit breaker configuration support

### LINKED ISSUE #474: task(config): Add circuit breaker configuration support
### Description

Add `ClientCircuitBreakerConfiguration` to the client configuration, matching the existing implementation in Java, Python, Node, and Go (valkey-io/valkey-glide#6050). The circuit breaker logic lives in the Rust core, this task only requires adding the config class, serializing it to protobuf, and surfacing the `CircuitBreakerException` type to callers.

### Checklist

- [ ] Add `ClientCircuitBreakerConfiguration` class with builder/options pattern
- [ ] Serialize config to protobuf `ConnectionRequest`
- [ ] Add `CircuitBreakerException` type
- [ ] Add unit test for config serialization

### Related

- valkey-io/valkey-glide#6050
- valkey-io/valkey-glide#5996

### Additional Information

:white_circle: None

currantw:
## Summary

Add `CircuitBreakerConfig` to the client configuration, matching the existing implementation in Java, Python, Node, and Go (valkey-io/valkey-glide#6050). The circuit breaker logic lives in the Rust core — this task adds the config class, serializes it via FFI, and surfaces the `CircuitBreakerException` type to callers.

## Issue Link

Closes #474.

- New `CircuitBreakerConfig` class with fluent `WithXxx()` API for configuring the client-wide circuit breaker
- New `Errors.CircuitBreakerException` thrown when the breaker is open and requests are rejected
- `ConnectionConfiguration` builder gains `WithCircuitBreaker(CircuitBreakerConfig)` method
- All config properties are optional — unset properties use Rust core defaults

## Implementation

- **`CircuitBreakerConfig.cs`**: Sealed class with nullable properties, fluent setters with validation, and `ToFfi()` conversion
- **`Errors.cs`**: Added `CircuitBreakerOpen = 4` to `RequestErrorType` enum and `CircuitBreakerException` class
- **`FFI.structs.cs`**: Added `CircuitBreakerConfig` readonly struct (primary constructor pattern) and wired through `ConnectionRequest`
- **`ConnectionConfiguration.cs`**: Added field to `ConnectionConfig` record and builder method
- **`rust/src/ffi.rs`**: Added `CircuitBreakerConfig` FFI struct and conversion in `create_connection_request()`. Also simplified existing code with `then_some`/`then`/`transpose` patterns.

## Limitations

- Only standalone server positive test (trip + assert exception) — cluster positive test would require pausing individual cluster nodes
- Documentation page at https://glide.valkey.io/how-to/connections/circuit-breaker/ does not yet have C# examples (tracked in valkey-io/valkey-glide-docs#273)

## Testing

- 23 unit tests (`CircuitBreakerConfigTests.cs`): defaults, validation, fluent chaining, builder integration
- 9 integration tests (`CircuitBreakerTests.cs`):
  - Default/custom config connects successfully (standalone + cluster)
  - Does not interfere with normal operation (aggressive thresholds)
  - No false positives under load (200 sequential commands)
  - **Positive test**: trips on timeouts and rejects with `CircuitBreakerException`
- All tests pass locally

## Related Issues

- valkey-io/valkey-glide#6050 (upstream circuit breaker feature)
- valkey-io/valkey-glide#5996 (circuit breaker design)
- valkey-io/valkey-glide-docs#273 (C# docs for circuit breaker page)

## Checklist


rust/src/ffi.rs
```diff
@@ -274,14 +286,10 @@ pub(crate) unsafe fn create_connection_request(
...
                     },
-                    refresh_interval_seconds: if auth_info
+                    refresh_interval_seconds: auth_info
                         .iam_credentials
                         .has_refresh_interval_seconds
-                    {
-                        Some(auth_info.iam_credentials.refresh_interval_seconds)
-                    } else {
-                        None
-                    },
```
currantw, line 279: Just simplified slightly using `then_some`.

CHANGELOG.md
```diff
@@ -34,7 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `RESET` (#435)
   - `SAVE` (#440)
 - Custom socket address resolution support via callback (#392)
-- `NodeDiscoveryMode` configuration option for standalone clients, with `Standard`, `Static` (proxy-compatible, e.g. Envoy), and `DiscoverAll` modes (#131)
+- `NodeDiscoveryMode` configuration option for standalone clients (#131)
```
currantw, line 38: Just simplified a bit. More details in issue. 🤷

sources/Valkey.Glide/CircuitBreakerConfig.cs
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    #endregion
+    #region Internal Methods
+
+    /// <summary>
+    /// Converts to the FFI representation for marshalling to Rust core.
+    /// </summary>
+    internal FFI.CircuitBreakerConfig ToFfi() => new(
+        (uint)(WindowSize?.TotalMilliseconds ?? 0),
```
currantw: This is a common problem I've noticed. Many of the configuration values only accept `uint` (C#) / `u32` (Rust), including this one – though this actually isn't consistent: `ClientSideCacheConfig.EntryTtlMs`, for example, us a `ulong` / `u64`. 

For `CircuitBreakerConfig`, we now evaluate this dynamically. But I think if might be nicer to instead update the core to use `u64` for these values? Even though this isn't a realistic timeout value, it would mean that we wouldn't need to do any validation on the client side, and wouldn't need to think about all these exceptions and error handling.

Thoughts?

sources/Valkey.Glide/ConnectionConfiguration.cs
```diff
@@ -895,6 +897,28 @@ public T WithClientSideCache(ClientSideCacheConfig clientSideCacheConfig)
...
         }
 
+        #endregion
+        #region Circuit Breaker
+
+        /// <summary>
+        /// Circuit breaker configuration for the client. When set, enables a circuit breaker
+        /// that detects unhealthy core state and rejects requests at the client boundary.
+        /// </summary>
+        /// <seealso cref="Glide.CircuitBreakerConfig"/>
```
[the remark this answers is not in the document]
currantw: Addressed in 4345ed63.

sources/Valkey.Glide/CircuitBreakerConfig.cs
```diff
@@ -0,0 +1,144 @@
...
+    /// </summary>
+    /// <param name="windowSize">The window size. Must be positive.</param>
+    /// <returns>This instance for method chaining.</returns>
+    /// <exception cref="ArgumentOutOfRangeException">Thrown when <paramref name="windowSize"/> is zero or negative.</exception>
+    public CircuitBreakerConfig WithWindowSize(TimeSpan windowSize)
+    {
+        ArgumentOutOfRangeException.ThrowIfLessThanOrEqual(windowSize, TimeSpan.Zero, nameof(windowSize));
+        WindowSize = windowSize;
+        return this;
+    }
```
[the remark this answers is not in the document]
currantw: Addressed in 6303792e.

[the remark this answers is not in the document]
currantw: Addressed in https://github.com/valkey-io/valkey-glide-csharp/commit/4345ed637ce5ff2602aec99db170aecea60c8736.

jeremyprime:
As mentioned in comments on the PHP PR I think we should ensure consistency between all the clients, the key pieces being (A) having the CB defaults in the client, and (B) not worrying about config edge case validation (until https://github.com/valkey-io/valkey-glide/issues/5937, although you already have it here so that's fine).

sources/Valkey.Glide/CircuitBreakerConfig.cs
jeremyprime: In this case a window size of u32 makes sense (u32 = ~50 days, u64 = ~580 million years...).  Seems only cache related ones are u64 since they can cover large intervals (many months).  So I think the current u32/u64 split is fine.

currantw: Since we use `TimeSpan` for intervals in C#, do you think the current behaviour – throw if value is too big – works then?

jeremyprime: Yes, that catches any overflow issues (values too big to fit in u32), so that's good for WindowSize and OpenTimeout here.

currantw:
Made the following changes for consistency with other clients:

- Added default values in the client
- Remove unnecessary edge case validation
- Removed integration tests

However I didn't retain some edge case validation for these two cases:

- `TimeSpan` (window size and open timeout). The client validates that these can fit in u32 to avoid overflow.
- Zero values. Since the Rust core uses zero as a default value for many of these properties, I also validate that these values cannot be set to zero by the user. These aren't valid values, and if we allowed them, the user would get some unexpected results!

jeremyprime (approved):

prateek-kumar-improving (approved):

Issue #499, opened 2026-07-31
feat(tests): Add mTLS integration tests with server requiring client certificates


currantw:
### Describe the feature

Add integration tests that verify mutual TLS (mTLS) works end-to-end against a server that **requires** client certificate authentication (`--tls-auth-clients yes`).

There are currently no mTLS integration tests. The mTLS implementation (#488) includes unit tests for builder validation and FFI marshaling, but no end-to-end tests that verify the client certificate is actually presented during the TLS handshake.

### Use Case

Without these tests, we cannot confirm that the mTLS implementation actually works for its intended purpose: authenticating the client to a server that requires it. A regression in certificate presentation during the TLS handshake would go undetected.

### Proposed Solution

1. **Test infrastructure**: Extend `ServerManager` to support starting a Valkey server with `--tls-auth-clients yes` and generate a client certificate/key signed by the test CA.
2. **Upstream dependency**: The `cluster_manager.py` script in the `valkey-glide` submodule hardcodes `--tls-auth-clients no`. Either:
   - Submit a PR to `valkey-io/valkey-glide` to make this configurable (e.g., `--tls-auth-clients` CLI arg).
   - Or bypass `cluster_manager.py` and start the server directly with custom TLS args.
3. **New test fixture**: `MutualTlsServerFixture` that provides client cert/key material and a server requiring mTLS.
4. **Tests to add**:
   - Connection succeeds with valid client cert/key (standalone + cluster).
   - Connection fails when no client cert is provided but server requires one.
   - Connection fails with a cert not signed by the server's trusted CA.

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might cause a breaking change

### Related

- #488 (mTLS implementation PR)
- `valkey-glide` submodule `cluster_manager.py` — hardcodes `--tls-auth-clients no` at [line 513](https://github.com/valkey-io/valkey-glide/blob/main/utils/cluster_manager.py#L513)
- Go client has the same gap: `TODO #5509: TLS tests do not currently run as part of CI`

### Additional Information

- No other GLIDE client (Java, Python, Go, Node) currently tests against a server requiring client authentication.
- The Python and Go clients use the server's own cert/key as client material against a non-requiring server (same approach as our current tests).

PR #500, opened 2026-07-31
feat(config): Add mutual TLS (mTLS) client certificate support

### LINKED ISSUE #488: feat(config): add mTLS (mutual TLS) support
### Describe the feature

Add mutual TLS (mTLS) support to the Valkey GLIDE C# client. mTLS enables two-way certificate authentication between the client and the Valkey server, where both parties present and verify certificates. This is already supported in other GLIDE language clients (Java, Python, Node.js) but is currently missing from the C# implementation.

### Use Case

Secure enterprise deployments often require mTLS to ensure that only authorized clients can connect to database servers. Without mTLS support, C# users cannot:

- Meet zero-trust network security requirements that mandate mutual authentication.
- Deploy Valkey GLIDE C# in environments where server-only TLS is insufficient (e.g., regulated industries, multi-tenant cloud environments).
- Achieve feature parity with other GLIDE language clients that already support mTLS.

### Proposed Solution

Extend `ConnectionConfiguration` (or `TlsOptions`) to accept client certificate and key parameters for mTLS authentication. The implementation should:

1. Add configuration options for client certificate (PEM or PFX) and private key.
2. Pass the client credentials through to `glide-core` via the existing FFI layer.
3. Support both file-path-based and in-memory certificate loading.
4. Follow the same API patterns used by other GLIDE clients for consistency.

Reference implementations:

- [Java GLIDE mTLS](https://github.com/valkey-io/valkey-glide/tree/main/java)
- [Python GLIDE mTLS](https://github.com/valkey-io/valkey-glide/tree/main/python)

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might cause a breaking change

### Related

- Other GLIDE clients already support mTLS via `glide-core`
- The Rust core (`glide-core`) already has mTLS plumbing available

### Additional Information

- The `glide-core` Rust layer already supports mTLS configuration, so the primary work is exposing this capability through the C# FFI boundary and public API.
- This should be a non-breaking addition (new optional configuration parameters).
- Integration tests will require a Valkey server configured with mTLS (client certificate verification enabled).

currantw:
## Summary

Add mutual TLS (mTLS) client certificate support to the Valkey GLIDE C# client.

## Issue Link

Closes #488.

### GLIDE

- **`WithClientCertificate(byte[] cert, byte[] key)`** — configures static mTLS with client certificate and key data.
- **`WithClientCertificate(string certPath, string keyPath)`** — configures dynamic mTLS from client certificate and key paths.
- **`WithClientCertificate(string certPath, string keyPath, TimeSpan reloadInterval)`** — same as above with a custom reload interval.
- **`GuardClauses`** extended with reusable validation methods:
  - `ThrowIfDataNotSupported`
  - `ThrowIfFileNotSupported`
  - `ThrowIfNotUintMilliseconds`
  - `ThrowIfNotPositiveUintSeconds`
- Timeout/interval fields in `ConnectionConfig` changed from `TimeSpan?` to `uint?` (milliseconds/seconds) to match the FFI wire format directly.

### StackExchange.Redis Compatibility

Matches StackExchange.Redis API:

- **`SetUserPemCertificate(string certPath, string? keyPath)`** — supply a user certificate from a PEM file pair and enable TLS.

## Implementation

- mTLS fields are inlined directly on both the C# `ConnectionRequest` FFI struct and the Rust `ConnectionConfig` struct — no intermediate wrapper types.
- The Rust FFI extracts cert/key bytes via `ptr_to_bytes` helper, paths via `ptr_to_opt_str`, and cert reload config via `bool`/`u32` fields.
- `ClearMutualTls()` resets all mTLS fields when switching between byte-based and path-based modes (last call wins).
- SER compat: when `userKeyPath` is null, the same file bytes are passed for both cert and key — glide-core's PEM parser (rustls-pki-types) extracts each by scanning for respective PEM headers.

## Limitations

- **No integration tests** — the test infrastructure (`cluster_manager.py`) doesn't support `--tls-auth-clients yes`. Filed as #499.
- **PFX support deferred (SER only)** — `SetUserPfxCertificate` requires algorithm-specific key export. Not implemented for GLIDE at this time.

## Testing

- Unit tests for `WithClientCertificate`
- Unit tests for `SetUserPemCertificate`
- Unit tests for `RequestTimeout`, `ConnectionTimeout`, `PubSubReconciliationInterval` validation.

Integration tests will be added in #499.

## Related Issues

- #499 — Add mTLS integration tests with server requiring client certificates.

## Checklist

- [x] All TODOs referencing issue(s) closed by this PR have been resolved.


sources/Valkey.Glide/ConnectionConfiguration.cs
```diff
@@ -834,15 +911,20 @@ public T WithLazyConnect(bool lazyConnect)
...
+                ? TimeSpan.FromMilliseconds(ms)
+                : null;
             set
             {
-                if (value <= TimeSpan.Zero)
+                if (value.HasValue)
                 {
-                    throw new ArgumentException("PubSubReconciliationInterval must be positive", nameof(value));
+                    GuardClauses.ThrowIfNotUintMilliseconds(value.Value, nameof(PubSubReconciliationInterval));
+                    Config.PubSubReconciliationIntervalMilliseconds = (uint)value.Value.TotalMilliseconds;
```
currantw: Added fail fast validation to some of these existing `TimeSpan` instances. Previously, a value that was too large would have thrown on client creation when trying to cast to `uint`.

sources/Valkey.Glide/ConnectionConfiguration.cs
```diff
@@ -535,64 +549,119 @@ public T WithInsecureTls(bool useInsecure = true)
...
+        /// <summary>
+        /// Configures mutual TLS for the given client certificate and key path.<p/>
+        /// The client periodically reloads the client certificate and key to support certificate rotation.
+        /// </summary>
+        /// <param name="certificatePath">Client certificate file path</param>
+        /// <param name="keyPath">Client key file path</param>
+        /// <returns>This builder for method chaining.</returns>
+        /// <exception cref="ArgumentException">If <paramref name="certificatePath"/> or <paramref name="keyPath"/> is null, empty, or too large.</exception>
+        /// <seealso href="https://glide.valkey.io/how-to/security/tls/">Valkey GLIDE – Configure TLS</seealso>
+        public T WithClientCertificate(string certificatePath, string keyPath)
```
currantw, line 618: **Design Decisions**:

- These methods name intentionally differ from other GLIDE clients to match the existing `WithTrustedCertificate` methods (which are themselves intended to be similar to SER).
- Similarly, we do not have a separate configuration object for mTLS. Again, this is to stay consistent with existing pattern, where these are exposed directly on the GLIDE configuration object.
- Did not expose a "read bytes from path" method, unlike other GLIDE clients. This follows the current patten for `WithTrustedCertificate` – and it is trivial to call `ReadFile`.

sources/Valkey.Glide/ConnectionConfiguration.cs
```diff
@@ -535,64 +549,119 @@ public T WithInsecureTls(bool useInsecure = true)
...
+            ArgumentException.ThrowIfNullOrEmpty(keyPath, nameof(keyPath));
+
+            ClearMutualTls();
+
+            Config.ClientCertificatePath = certificatePath;
+            Config.ClientKeyPath = keyPath;
+            Config.CertReloadEnabled = true;
+
+            return (T)this;
+        }
```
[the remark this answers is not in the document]
currantw: For path-based mTLS, glide-core reads the files directly from disk — the C# client never reads them. We only pass the paths through to the Rust core, which handles I/O and periodic reload. We cannot validate file size without reading the file, which would defeat the purpose of path-based mode. Core will reject invalid files with its own error during connection.

sources/Valkey.Glide/Internals/GuardClauses.cs
```diff
@@ -33,6 +41,42 @@ public static void ThrowIfCommandFlags(CommandFlags flags)
...
+    /// Throws if the file at the given path is empty or exceeds <see cref="MaxDataSize"/>.
+    /// </summary>
+    /// <param name="path">The file path to check.</param>
+    /// <param name="paramName">The parameter name for the exception.</param>
+    internal static void ThrowIfFileNotSupported(string path, string paramName)
+    {
+        var fileLength = new FileInfo(path).Length;
+        ArgumentOutOfRangeException.ThrowIfZero(fileLength, paramName);
+        ArgumentOutOfRangeException.ThrowIfGreaterThan(fileLength, MaxDataSize, paramName);
+    }
```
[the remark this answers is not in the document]
currantw: FileInfo already throws ArgumentNullException for null and ArgumentException for empty string with clear messages. The param name will differ ("fileName" vs the caller's param name), but this is consistent with how .NET BCL methods work — they use their own parameter names. Adding an explicit check would be redundant.

sources/Valkey.Glide/Abstract/ConfigurationOptions.cs
```diff
@@ -318,13 +296,26 @@ public void TrustIssuer(string certificatePath)
...
+    /// <param name="userKeyPath">The path for the user key (commonly a .key file).</param>
+    /// <exception cref="ArgumentNullException">If <paramref name="userCertificatePath"/> is <see langword="null"/>.</exception>
+    /// <exception cref="FileNotFoundException">If a specified file does not exist.</exception>
+    public void SetUserPemCertificate(string userCertificatePath, string? userKeyPath = null)
+    {
+        // When keyPath is null, pass the same blob for both cert and key.
+        // glide-core extracts certificates and private keys independently by scanning for
+        // their respective PEM headers, ignoring irrelevant sections in the same blob.
+        _clientCertificate = File.ReadAllBytes(userCertificatePath);
+        _clientKey = userKeyPath is not null ? File.ReadAllBytes(userKeyPath) : _clientCertificate;
```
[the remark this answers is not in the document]
currantw: Addressed in 491b6976.

sources/Valkey.Glide/Internals/FFI.structs.cs
```diff
@@ -1144,13 +1183,50 @@ private struct ConnectionRequest
...
+        public IntPtr ClientCertPtr;
+
+        public nuint ClientKeyLen;
+        public IntPtr ClientKeyPtr;
+
+        [MarshalAs(UnmanagedType.LPStr)]
+        public string? ClientCertPath;
+
+        [MarshalAs(UnmanagedType.LPStr)]
+        public string? ClientKeyPath;
```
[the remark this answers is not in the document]
currantw: Raised #501

dev/conf/coverage-baseline.json
```diff
@@ -1,6 +1,6 @@
 {
   "server_type": "valkey",
   "server_version": "9.0",
-  "line_coverage": 71.4,
-  "branch_coverage": 51.9
+  "line_coverage": 71.7,
+  "branch_coverage": 51.7
```
currantw: Branch coverage decrease because we consolidated logic in `GuardClauses` – so there are actually fewer branches!

tests/Valkey.Glide.IntegrationTests/ConnectionManagementCommandTests.cs
```diff
@@ -155,14 +155,24 @@ private static void AssertTrackingInfoOn(ClientTrackingInfo info)
...
+
+        await using BaseClient client = clusterMode
+            ? await GlideClusterClient.CreateClient(
+                fixture.ClusterServer.CreateConfigBuilder()
+                    .WithRequestTimeout(requestTimeout)
+                    .Build())
+            : await GlideClient.CreateClient(
+                fixture.StandaloneServer.CreateConfigBuilder()
+                    .WithRequestTimeout(requestTimeout)
+                    .Build());
```
currantw, line 170: Unrelated attempt to fix a flaky test.

sources/Valkey.Glide/Internals/GuardClauses.cs
```diff
@@ -7,18 +7,26 @@ namespace Valkey.Glide.Internals;
...
-    /// <exception cref="NotImplementedException">Thrown if <paramref name="asyncState"/> is not null.</exception>
-    public static void ThrowIfAsyncState(object? asyncState)
-    {
-        if (asyncState is not null)
-        {
-            throw new NotImplementedException("Async state is not supported by Valkey GLIDE");
-        }
-    }
+    /// <seealso href="https://github.com/valkey-io/valkey-glide-csharp/issues/226">#226</seealso>
+    internal static readonly long MaxDataSize = 10 * 1024 * 1024;
```
xShinnRyuu: nit: `MaxDataSize` can be a int

sources/Valkey.Glide/ConnectionConfiguration.cs
```diff
@@ -535,64 +549,118 @@ public T WithInsecureTls(bool useInsecure = true)
...
+            ArgumentException.ThrowIfNullOrEmpty(certificatePath, nameof(certificatePath));
+            ArgumentException.ThrowIfNullOrEmpty(keyPath, nameof(keyPath));
+            GuardClauses.ThrowIfNotPositiveUintSeconds(reloadInterval, nameof(reloadInterval));
+
+            ClearMutualTls();
+
+            Config.ClientCertificatePath = certificatePath;
+            Config.ClientKeyPath = keyPath;
+            Config.CertReloadEnabled = true;
+            Config.CertReloadIntervalSeconds = (uint)reloadInterval.TotalSeconds;
```
xShinnRyuu: `(uint)reloadInterval.TotalSeconds` silently truncates fractional seconds, and the guard on line 639 only rejects non-positive or overflowing values. `TimeSpan.FromMilliseconds(500)` passes the guard, becomes `0u` here, and glide-core then treats `Some(0)` as "use default" (see `interval_seconds.filter(|&s| s != 0)` in `client/types.rs`), so the reload interval quietly jumps to the 5-minute core default. `TimeSpan.FromSeconds(1.5)` truncates to `1u`. The same shape exists in `RequestTimeout` (line 680), `ConnectionTimeout` (line 706), and `PubSubReconciliationInterval` (line 921) where `TotalMilliseconds` is cast to `uint`. Either reject sub-unit `TimeSpan` values in the guard (Copilot flagged this too), or round/ceil explicitly and document the rounding rule.

sources/Valkey.Glide/ConnectionConfiguration.cs
```diff
@@ -535,64 +549,118 @@ public T WithInsecureTls(bool useInsecure = true)
...
+        {
+            GuardClauses.ThrowIfBytesNotSupported(certificateData, nameof(certificateData));
+            GuardClauses.ThrowIfBytesNotSupported(keyData, nameof(keyData));
+
+            ClearMutualTls();
+
+            Config.ClientCertificate = certificateData;
+            Config.ClientKey = keyData;
+
+            return (T)this;
```
xShinnRyuu, line 605: Blocking: none of the three `WithClientCertificate` overloads sets `Config.TlsMode` to `SecureTls`. If a caller writes `builder.WithClientCertificate(certPath, keyPath)` without also calling `WithTls()`, the request lands in glide-core with `TlsMode::NoTls` and cert material set, and connect fails with `TLS certificates provided but TLS is disabled` (see `client/mod.rs` create_cluster_client / build_standalone_client). That is a confusing runtime error for what looks like a valid configuration. Either auto-enable TLS the way `SetUserPemCertificate` does (`Ssl = true`) or throw a clear `InvalidOperationException` at build time when mTLS is configured without TLS.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Mutual TLS (mTLS) support (#488)
```
xShinnRyuu, line 24: Blocking: this PR silently deletes two publicly-exposed members that shipped in prior releases: `ConnectionConfiguration.CertificateMaxSize` (was `public static readonly long`) and `CircuitBreakerConfig.MaxTimeSpan` (was `public static readonly TimeSpan`). Both were documented in exception messages and referenced by user code that pinned to these constants. Removing them is a source-breaking change and should either (a) be listed under a `### Changed` / `### Removed` heading in this CHANGELOG entry, or (b) be preserved as `[Obsolete]` aliases that forward to `GuardClauses.MaxDataSize` / `GuardClauses.MaxUintMilliseconds`. The project follows semver, and the PR body claims only additive API changes.

sources/Valkey.Glide/Abstract/ConfigurationOptions.cs
```diff
@@ -318,13 +296,36 @@ public void TrustIssuer(string certificatePath)
...
+        // their respective PEM headers, ignoring irrelevant sections in the same blob.
+        if (userKeyPath is not null)
+        {
+            GuardClauses.ThrowIfFileNotSupported(userKeyPath, nameof(userKeyPath));
+            _clientKey = File.ReadAllBytes(userKeyPath);
+        }
+        else
+        {
+            _clientKey = _clientCertificate;
+        }
```
xShinnRyuu, line 326: The behavior of passing the same PEM bytes for both cert and key when `userKeyPath` is null depends on the input containing BOTH a `CERTIFICATE` and a `PRIVATE KEY` section. A file with only `CERTIFICATE` (the common .crt shape) will pass this method, then fail at connect time inside rustls-pki-types with an opaque "no private key found" error. Consider documenting the requirement on the XML doc ("the file must contain both PEM sections"), or scanning for a `BEGIN .* PRIVATE KEY` marker eagerly so the failure surfaces at configuration time.

sources/Valkey.Glide/Abstract/ConnectionMultiplexer.cs
```diff
@@ -250,6 +250,9 @@ internal static T CreateClientConfigBuilder<T>(ConfigurationOptions configuratio
         _ = configuration.ReconnectRetryPolicy.HasValue ? config.ConnectionRetryStrategy = configuration.ReconnectRetryPolicy.Value : new();
         _ = configuration.ReadFrom.HasValue ? config.ReadFrom = configuration.ReadFrom.Value : new();
 
+        config.Config.ClientCertificate = configuration._clientCertificate;
+        config.Config.ClientKey = configuration._clientKey;
```
xShinnRyuu: These two lines poke `Config` fields directly instead of going through `WithClientCertificate(byte[], byte[])`. That is safe today because the builder is freshly constructed on line 221, but it bypasses `ThrowIfBytesNotSupported` (already run in `SetUserPemCertificate` for the file path, but not for `X509Certificate2` sources or any future SER cert entry point), skips `ClearMutualTls`, and drifts if new mTLS fields are added to `ConnectionConfig`. Prefer `config.WithClientCertificate(configuration._clientCertificate, configuration._clientKey)` when both are non-null.

rust/src/ffi.rs
```diff
@@ -53,6 +53,20 @@ unsafe fn ptr_to_opt_str(ptr: *const c_char) -> Result<Option<String>, String> {
     }
 }
 
+/// Convert a raw byte pointer and length to an owned `Vec<u8>`.
+/// Returns an empty Vec if the pointer is null.
+///
+/// # Safety
+///
+/// * If non-null, `ptr` must point to `len` valid bytes.
+unsafe fn ptr_to_bytes(ptr: *const u8, len: usize) -> Vec<u8> {
```
xShinnRyuu, line 62: `ptr_to_bytes` silently returns `Vec::new()` when `ptr.is_null()` even if `len > 0`. That hides a marshalling bug on the C# side (non-zero `ClientCertLen` paired with `IntPtr.Zero`) as an empty client cert, which then fails glide-core's `validate_client_cert_config` with a generic "both or neither" error. A `debug_assert!(len == 0)` before the `Vec::new()` return, or an explicit `Err(..)` when `ptr.is_null() && len > 0`, would make FFI misuse fail loud during development.

sources/Valkey.Glide/ConnectionConfiguration.cs
```diff
@@ -535,64 +549,118 @@ public T WithInsecureTls(bool useInsecure = true)
...
+
+        /// <summary>
+        /// Configures mutual TLS for the given client certificate and key path.<p/>
+        /// The client periodically reloads the client certificate and key to support certificate rotation.
+        /// </summary>
+        /// <param name="certificatePath">Client certificate file path</param>
+        /// <param name="keyPath">Client key file path</param>
+        /// <returns>This builder for method chaining.</returns>
+        /// <exception cref="ArgumentException">If <paramref name="certificatePath"/> or <paramref name="keyPath"/> is null, empty, or too large.</exception>
+        /// <seealso href="https://glide.valkey.io/how-to/security/tls/">Valkey GLIDE – Configure TLS</seealso>
```
xShinnRyuu, line 617: Two doc nits on the path-based overloads: 

- (1) the summary says "periodically reloads" but does not name the interval; the actual default comes from glide-core (`DEFAULT_RELOAD_INTERVAL_SECONDS = 300`, i.e. 5 minutes). Please mention it so the two-arg vs three-arg overloads have obvious semantics.

- (2) The `<exception cref="ArgumentException">` line lists "null, empty, or too large" but `ArgumentException.ThrowIfNullOrEmpty` throws `ArgumentNullException` for null, and there is no "too large" check at all in the path overloads (currantw's response to Copilot confirms path validation is left to core). Reword to "null or empty" and add `<exception cref="ArgumentNullException">`.

sources/Valkey.Glide/ConnectionConfiguration.cs
```diff
@@ -535,64 +549,118 @@ public T WithInsecureTls(bool useInsecure = true)
...
+        #endregion
+        #region Mutual TLS
+
+        /// <summary>
+        /// Configures mutual TLS for the given client certificate and key data.
+        /// </summary>
+        /// <param name="certificateData">Client certificate data</param>
+        /// <param name="keyData">Client key data</param>
+        /// <returns>This builder for method chaining.</returns>
+        /// <exception cref="ArgumentException">If <paramref name="certificateData"/> or <paramref name="keyData"/> is empty or too large.</exception>
```
xShinnRyuu: Nit: `ThrowIfBytesNotSupported` throws `ArgumentNullException` when either byte[] is null, but the XML doc only advertises `ArgumentException`. Add `<exception cref="ArgumentNullException">` so callers know which catch to write.

sources/Valkey.Glide/ConnectionConfiguration.cs
```diff
@@ -15,11 +15,6 @@ namespace Valkey.Glide;
 /// <seealso href="https://glide.valkey.io/how-to/connections/timeouts-and-reconnect-strategy/">Valkey GLIDE – Timeouts and Reconnect Strategy</seealso>
 public abstract class ConnectionConfiguration
 {
-    /// <summary>
-    /// Maximum root certificate size for TLS connections.
-    /// </summary>
-    public static readonly long CertificateMaxSize = 10 * 1024 * 1024; // 10 MB
```
prateek-kumar-improving, line 21: This was a public variable in a public class. Is this a breaking change?

sources/Valkey.Glide/ConnectionConfiguration.cs
currantw: Addressed in f8acef09:

- Replaced all `(uint)value.TotalSeconds` / `(uint)value.TotalMilliseconds` truncating casts with `TimeUtils` methods that use `Math.Round` (for `ToPositiveUintSecs`) or tick-based arithmetic (for `ToPositiveUintMs`).
- `ToPositiveUintMs` and `ToPositiveUintSecs` reject zero and overflow — used for configuration values like `RequestTimeout`, `ConnectionTimeout`, `CertReloadInterval`, `PubSubReconciliationInterval`.
- `ToULongMs` and `ToNonNegativeDoubleSecs` allow zero — used for command arguments where 0 means "block indefinitely" (BLPOP, BZMPOP, etc.).
- Opened #512 to address the `ConnectionTimeout`/`RequestTimeout` defaults diverging from glide-core.

sources/Valkey.Glide/ConnectionConfiguration.cs
currantw: No change — this is consistent with all other GLIDE clients (Java, Python, Node) which all treat TLS enablement as a separate explicit flag and do not auto-enable TLS when client certificates are configured. The C# client's existing `WithTrustedCertificate` also does not touch `TlsMode`.

glide-core already validates this and returns a clear error if mTLS is configured without TLS enabled, so we don't need to duplicate that check on the C# side.

sources/Valkey.Glide/Internals/GuardClauses.cs
currantw: Overtaken by changes. Reverted to CertificateMaxSize, which was already an `int`.

CHANGELOG.md
currantw: 1. Restored `ConnectionConfiguration.CertificateMaxSize` as a public constant (ebd30313). 
2. `CircuitBreakerConfig.MaxTimeSpan` remains removed, since this has not been released yet.

sources/Valkey.Glide/ConnectionConfiguration.cs
currantw: `CertificateMaxSize` has been restored as a public constant. No breaking change for this member. Thanks for flagging. 💪

sources/Valkey.Glide/Abstract/ConfigurationOptions.cs
currantw: This matches StackExchange.Redis behaviour (since this is the SER compatibility API): https://github.com/StackExchange/StackExchange.Redis/blob/09d80656fadf9574dde1a477d41f401f962591ac/src/StackExchan
  ge.Redis/ConfigurationOptions.cs

sources/Valkey.Glide/Abstract/ConnectionMultiplexer.cs
currantw: Addressed in 88864ed5.

rust/src/ffi.rs
currantw: Don't think this is necessary? This is an internal only method, and length will always be zero when pointer is nuil. 🤷

sources/Valkey.Glide/ConnectionConfiguration.cs
currantw: Addressed in afb48a79.

sources/Valkey.Glide/ConnectionConfiguration.cs
currantw: (2) Addressed in afb48a79 — fixed exception docs on all three overloads.

currantw: Regarding the default reload interval: the glide-core default (300s) is already documented at the linked TLS guide (https://glide.valkey.io/how-to/security/tls/) — hardcoding it in the XML doc would be another value that can drift. The `<seealso>` link already points users there. This is similar to the discussion I had with @jeremyprime around this PR https://github.com/valkey-io/valkey-glide-csharp/pull/509/changes: let's let glide-core do the default, and put default values just there and on the docs site if possible.

prateek-kumar-improving (approved):

xShinnRyuu (approved):

PR #510, opened 2026-08-04
Add `CLIENT KILL` command support

### LINKED ISSUE #276: feat: Add `CLIENT KILL` command support
### Summary

Add GLIDE-native `CLIENT KILL` command support to the Valkey GLIDE C# client.

### Description

The [`CLIENT KILL`](https://valkey.io/commands/client-kill/) command is supported by Valkey GLIDE core (request types `ClientKillSimple` / `ClientKill`) but is not yet implemented in the C# client.

StackExchange.Redis supports `CLIENT KILL` on `IServer`, but its signatures rely on SER-specific types (`EndPoint`, `ClientKillFilter`, `ClientType`) that are tightly coupled to SER's direct server connection model. SER compatibility is not feasible for this command — implement as GLIDE-native only.

### Proposed Solution

Add `CLIENT KILL` methods to the GLIDE-native client interfaces and implement in `GlideClient` and `GlideClusterClient`. Refer to the other Valkey GLIDE clients (Java, Python, Go, Node) for signature conventions, option types, and return types.

### Effort & Severity Estimates

- Effort: medium
- Severity: low

### Related

- #275 — Connection management command API cleanup.
- #270 — Parent issue: command API cleanup.

currantw:
## Summary

Add `CLIENT KILL` command support for both the GLIDE native API and the StackExchange.Redis compatibility layer, with comprehensive filter support for Valkey 7.2 through 9.0.

## Issue Link

Closes #276.

### GLIDE API

| Interface | Method | Returns |
|-----------|--------|---------|
| `IBaseClient` | `ClientKillAsync(ClientFilterOptions options)` | `Task<long>` |

#### `ClientFilterOptions`

Fluent builder for CLIENT KILL/LIST filters.

| Method | Filter | Since |
|--------|--------|-------|
| `WithType` / `WithoutType` | TYPE / NOT-TYPE | 7.2 / 9.0 |
| `WithId` / `WithIds` / `WithoutId` / `WithoutIds` | ID / NOT-ID | 7.2 / 9.0 |
| `WithUser` / `WithoutUser` | USER / NOT-USER | 7.2 / 9.0 |
| `WithAddress` / `WithoutAddress` | ADDR / NOT-ADDR | 7.2 / 9.0 |
| `WithLocalAddress` / `WithoutLocalAddress` | LADDR / NOT-LADDR | 7.2 / 9.0 |
| `WithSkipMe` | SKIPME | 7.2 |
| `WithMaxAge` | MAXAGE | 8.0 |
| `WithName` / `WithoutName` | NAME / NOT-NAME | 9.0 |
| `WithIdle` | IDLE | 9.0 |
| `WithFlag` / `WithFlags` / `WithoutFlag` / `WithoutFlags` | FLAGS / NOT-FLAGS | 9.0 |
| `WithLibraryName` / `WithoutLibraryName` | LIB-NAME / NOT-LIB-NAME | 9.0 |
| `WithLibraryVersion` / `WithoutLibraryVersion` | LIB-VER / NOT-LIB-VER | 9.0 |
| `WithDatabaseId` / `WithoutDatabaseId` | DB / NOT-DB | 9.0 |
| `WithCapability` / `WithoutCapability` | CAPA / NOT-CAPA | 9.0 |
| `WithIpAddress` / `WithoutIpAddress` | IP / NOT-IP | 9.0 |

#### Enums

| Enum | Description |
|------|-------------|
| `ClientType` | Connection type (`Normal`, `Primary`, `Replica`, `PubSub`) |
| `ClientFlag` | Char-backed client flags (`None`, `Primary`, `Replica`, `PubSub`, `Blocked`, `Multi`, etc.) |
| `ClientCapability` | Char-backed client capabilities (`Redirect`) |

### SER API

| Interface | Method | Returns |
|-----------|--------|---------|
| `IServer` | `ClientKillAsync(EndPoint endpoint, CommandFlags flags)` | `Task` |
| `IServer` | `ClientKillAsync(long? id, ClientType? clientType, EndPoint? endpoint, bool skipMe, CommandFlags flags)` | `Task<long>` |
| `IServer` | `ClientKillAsync(ClientKillFilter filter, CommandFlags flags)` | `Task<long>` |

### Supporting Types

- `ClientFilterOptions` — fluent builder for CLIENT KILL/LIST filters with `With*`/`Without*` methods
  - Base filters: ID, TYPE, USER, ADDR, LADDR, SKIPME, MAXAGE
  - Valkey 9.0 filters: NAME, IDLE, FLAGS, LIB-NAME, LIB-VER, DB, IP, CAPA, and all NOT-* variants
- `ClientType` — connection type enum with `Primary` value (serializes as `"master"` on wire)
- `ClientFlag` — char-backed enum for client connection flags (`'N'`, `'M'`, `'S'`, `'P'`, etc.)
- `ClientCapability` — char-backed enum for client capabilities (`'r'` for redirect) with `ToString()` extension for CLIENT CAPA full names
- `ClientKillFilter` — SER compatibility type, delegates to GLIDE API via `ValkeyServer`

## Implementation

- glide-core routes CLIENT KILL to all nodes with Sum response aggregation (submodule bump to [valkey-io/valkey-glide#6709](https://github.com/valkey-io/valkey-glide/pull/6709))
- Only the new filter-based format is used (glide-core does not support the legacy `CLIENT KILL addr:port` form)
- `ClientFilterOptions` is designed for reuse with CLIENT LIST (TODO #414)
- SER `ValkeyServer` converts SER types to `ClientFilterOptions` and delegates to the GLIDE client API
- `ClientType.Primary` serializes as `"master"` on the wire for backwards compatibility
- Wire format details verified against Valkey source:
  - `MAXAGE` and `IDLE` both use seconds (as `double` via `TotalSeconds`)
  - `FLAGS` and `CAPA` use combined char strings (e.g. `FLAGS MP`, `CAPA r`)
  - IPv6 addresses are bracket-wrapped (e.g. `[::1]:6379`)

## Limitations

- No batch/transaction support (consistent with SER and all other GLIDE language clients)

## Testing

- Unit tests for `ClientFilterOptions` serialization (success and failure cases, including all Valkey 9.0 filters)
- Unit test for `TimeUtils` helpers
- Integration tests for GLIDE API (kill by ID, by address, non-existent cases)
- Integration tests for SER API (kill by endpoint, by ID, with filter)
- Uses `Polling.WaitForAsync` for reconnection assertions to avoid flakiness

## Related Issues

- [valkey-io/valkey-glide#6709](https://github.com/valkey-io/valkey-glide/pull/6709) — glide-core routing fix (merged)
- #414 — CLIENT INFO / CLIENT LIST (future, `ClientFilterOptions` reuse)

## Checklist


sources/Valkey.Glide/abstract_Enums/ClientType.cs
```diff
@@ -1,33 +1,29 @@
...
-    Replica = 1, // as an implementation detail, note that enum.ToString without [Flags] prefers *earlier* values
+    Primary,
 
     /// <summary>
-    /// Replication connections.
+    /// Replica node connection.
     /// </summary>
-    [Obsolete("Starting with Valkey version 5, Valkey has moved to 'replica' terminology. Please use " + nameof(Replica) + " instead, this will be removed in 3.0.")]
-    [Browsable(false), EditorBrowsable(EditorBrowsableState.Never)]
-    Slave = 1,
```
currantw, line 27: Obselete SER values not supported by GLIDE.

sources/Valkey.Glide/abstract_Enums/ClientType.cs
```diff
@@ -1,33 +1,29 @@
...
     /// </summary>
-    Normal = 0,
+    Normal,
 
     /// <summary>
-    /// Replication connections.
+    /// Primary connection.
     /// </summary>
-    Replica = 1, // as an implementation detail, note that enum.ToString without [Flags] prefers *earlier* values
+    Primary,
```
currantw, line 18: Primary is actually not supported by SER, so in this case we are _extending_ the SER interface.

tests/Valkey.Glide.IntegrationTests/ConnectionManagementCommandTests.cs
currantw, line 1: Just added regions and reordered alphabetically, as it was growing in size. Only client kill tests are new or modified.

tests/Valkey.Glide.UnitTests/Options/FailoverOptionsTests.cs
```diff
@@ -7,20 +7,24 @@ namespace Valkey.Glide.UnitTests;
 public class FailoverOptionsTests
 {
     [Fact]
-    public void ToArgs_Success() => Assert.Multiple(
```
currantw, line 10: Don't like `Assert.Multiple` for these cases. It smooshes a bunch of separate tests into one.

sources/Valkey.Glide/abstract_Enums/ClientType.cs
currantw: While `ClientType` was previously released, it wasn't actually used by any commands — `CLIENT KILL` wasn't implemented until this PR. So it is functionally unreleased, and we are free to clean it up. Additionally, not supporting obsolete StackExchange.Redis features is consistent with the existing pattern in this project.

sources/Valkey.Glide/abstract_Enums/ClientType.cs
```diff
@@ -1,33 +1,29 @@
...
 /// </summary>
-[System.Diagnostics.CodeAnalysis.SuppressMessage("Design", "CA1069:Enums values should not be duplicated", Justification = "Compatibility")]
+/// <seealso href="https://valkey.io/commands/client-kill/" />
+/// <seealso href="https://valkey.io/commands/client-list/" />
 public enum ClientType
 {
     /// <summary>
-    /// Regular connections, including MONITOR connections.
+    /// Normal client connection.
     /// </summary>
```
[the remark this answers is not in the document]
currantw: While `ClientType` was previously released, it wasn't actually used by any commands — `CLIENT KILL` wasn't implemented until this PR. So it is functionally unreleased, and we are free to clean it up. Additionally, not supporting obsolete StackExchange.Redis features is consistent with the existing pattern in this project.

sources/Valkey.Glide/Commands/Options/ClientFilterOptions.cs
```diff
@@ -0,0 +1,235 @@
...
+            args.Add(SkipMe.Value ? ValkeyLiterals.yes : ValkeyLiterals.no);
+        }
+
+        return args;
+    }
+
+    #endregion
+    #region Private Fields
+
+    private readonly HashSet<long> _ids = [];
```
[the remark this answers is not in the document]
currantw: Switched to `SortedSet` in d731477f.

sources/Valkey.Glide/Commands/Options/ClientFilterOptions.cs
```diff
@@ -0,0 +1,235 @@
...
+
+    /// <summary>
+    /// Whether to skip the current connection.
+    /// </summary>
+    public bool? SkipMe { get; private set; }
+
+    /// <summary>
+    /// The maximum connection age to filter by.
+    /// </summary>
+    public TimeSpan? MaxAge => _maxAgeSecs.HasValue ? TimeSpan.FromSeconds(_maxAgeSecs.Value) : null;
```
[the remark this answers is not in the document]
currantw: I think the existing is clear and match valkey command itself.

sources/Valkey.Glide/Internals/TimeUtils.cs
```diff
@@ -65,6 +65,16 @@ public static ulong ToULongMs(TimeSpan timeSpan, string paramName)
...
 
+    /// <summary>
+    /// Converts a <see cref="TimeSpan"/> to <see cref="ulong"/> seconds, rounded to the nearest second.
+    /// </summary>
+    /// <exception cref="ArgumentOutOfRangeException">Thrown if <paramref name="timeSpan"/> is negative.</exception>
+    public static ulong ToULongSecs(TimeSpan timeSpan, string paramName)
+    {
+        ArgumentOutOfRangeException.ThrowIfLessThan(timeSpan, TimeSpan.Zero, paramName);
+        return (ulong)Math.Round(timeSpan.TotalSeconds);
+    }
```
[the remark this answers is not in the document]
currantw: Floating-point precision at seconds granularity is not a practical concern.

tests/Valkey.Glide.IntegrationTests/ServerTests.cs
```diff
@@ -13,9 +13,118 @@ public class ServerTests(TestConfiguration config)
...
+    {
+        var server = conn.GetServers().First();
+
+        // TODO #414: Update to use ClientInfoAsync() once available on IServer.
+        var info = (await server.ExecuteAsync("CLIENT", ["INFO"])).AsString()!;
+        var addr = info.Split(' ').First(f => f.StartsWith("addr=")).Split('=')[1];
+        var host = addr[..addr.LastIndexOf(':')];
+        var port = int.Parse(addr[(addr.LastIndexOf(':') + 1)..]);
+
+        await server.ClientKillAsync(new IPEndPoint(IPAddress.Parse(host), port));
```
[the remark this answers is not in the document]
currantw: Addressed in b9ad3c1a

AGENTS.md
```diff
@@ -31,6 +31,7 @@ Common commands:
   - `task test` (all tests)
   - `task test:unit` (unit tests only)
   - `task test:integration` (integration tests only)
+  - Integration tests automatically starts and stops Valkey servers. No external server is required.
```
currantw, line 34: Added a note here because this keep confusing Claude.

sources/Valkey.Glide/Commands/Options/ClientFilterOptions.cs
```diff
@@ -0,0 +1,237 @@
...
+        if (AddressHost is not null)
+        {
+            args.Add(ValkeyLiterals.ADDR);
+            args.Add($"{AddressHost}:{AddressPort}");
+        }
+
+        if (LocalAddressHost is not null)
+        {
+            args.Add(ValkeyLiterals.LADDR);
+            args.Add($"{LocalAddressHost}:{LocalAddressPort}");
```
affonsov: is this working with ipv6? We probably need the brackets formatting this. Can you double-check

tests/Valkey.Glide.IntegrationTests/ServerTests.cs
```diff
@@ -13,9 +13,117 @@ public class ServerTests(TestConfiguration config)
...
+    public async Task ClientKillAsync_ByAddress_KillsClient(ConnectionMultiplexer conn, bool _)
+    {
+        var server = conn.GetServers().First();
+
+        // TODO #414: Update to use ClientInfoAsync() once available on IServer.
+        var info = (await server.ExecuteAsync("CLIENT", ["INFO"])).AsString()!;
+        var addr = info.Split(' ').First(f => f.StartsWith("addr=")).Split('=')[1];
+        var endpoint = IPEndPoint.Parse(addr);
+
+        await server.ClientKillAsync(endpoint);
```
affonsov, line 79: there is no assertion on this test

affonsov (commented):
The other glide clients at the moment don't expose client kill. Do you know if we have a specific reason to not have this command exposed?

sources/Valkey.Glide/Commands/Options/ClientFilterOptions.cs
currantw: Good catch! Fixed in eb8e2770. Added `Utils.FormatAddress` method to convert. Added unit tests as well.

currantw:
> The other glide clients at the moment don't expose client kill. Do you know if we have a specific reason to not have this command exposed?

I don't believe that there is a specific reason. It is available on glide-core? 🤷 I just wanted to add it because it is a useful command for various tests.

Once this is merged, I was planning to raise issues for the other clients.

tests/Valkey.Glide.IntegrationTests/ServerTests.cs
currantw: Thanks for catching. Updated in 99c1fa58 to kill the connection and then verify it reconnects with a new ID.

sources/Valkey.Glide/Commands/Options/ClientFilterOptions.cs
```diff
@@ -0,0 +1,237 @@
...
+        return this;
+    }
+
+    #endregion
+    #region Internal Methods
+
+    /// <summary>
+    /// Converts to <c>CLIENT KILL</c> command arguments.
+    /// </summary>
+    internal GlideString[] ToClientKillArgs()
```
xShinnRyuu: An empty `ClientFilterOptions` produces an empty argument array from `ToClientKillArgs()`, so the request goes out as `CLIENT KILL` with nothing after it and the server returns a syntax error. Callers see that as an opaque error string. Throwing an `ArgumentException` here (or in `Request.ClientKill`) when no filter has been set would give a clearer signal at the call site.

sources/Valkey.Glide/Commands/Options/ClientFilterOptions.cs
```diff
@@ -0,0 +1,237 @@
...
+/// <seealso href="https://valkey.io/commands/client-kill/" />
+/// <seealso href="https://valkey.io/commands/client-list/" />
+public class ClientFilterOptions
+{
+    #region Public Properties
+
+    /// <summary>
+    /// The client IDs to filter by.
+    /// </summary>
+    public IReadOnlySet<long> Ids => _ids;
```
xShinnRyuu, line 32: Minor: `Ids` is exposed as `IReadOnlySet<long>` but the backing field is `SortedSet<long>`. Switching to `SortedSet` in d731477f fixed the deterministic serialization concern, but the return type doesn't tell callers that iteration is ascending. Either add a remark on the property spelling that out, or expose it as `IReadOnlyList<long>` so the ordering is part of the type contract.

xShinnRyuu (approved):
Minor issues

tests/Valkey.Glide.IntegrationTests/ServerTests.cs
affonsov: Only small concern in the core we we are using kill we usually have a loop waiting for reconnection. I'm just concerned that theses kill test may be flaky. Is something that we need to worry about it?

sources/Valkey.Glide/Commands/Options/ClientFilterOptions.cs
currantw: Hmmm. I'm inclined to leave as is? I would rather make no guarantee of ordering to the caller: gives us flexibility to change the internal implementation in the future if we like. 🤷

sources/Valkey.Glide/Commands/Options/ClientFilterOptions.cs
currantw: Bare `CLIENT KILL` with no filters is valid per the docs (all filters are optional within the new format) and works in practice (it just kills the current client). In fact, the main valkey-glide repo already uses it in `pubsub_reconnection_test.go` via a custom command.

tests/Valkey.Glide.IntegrationTests/ServerTests.cs
currantw: Updated to check reconnection with a different client ID, i.e.

```csharp
// Verify that client reconnected with new connection ID.
await AssertReconnected(server);
Assert.NotEqual(id, await server.ClientIdAsync());
```

sources/Valkey.Glide/Commands/Options/ClientFilterOptions.cs
```diff
@@ -0,0 +1,722 @@
...
+    /// Converts to <c>CLIENT KILL</c> command arguments.
+    /// </summary>
+    internal GlideString[] ToClientKillArgs()
+    {
+        List<GlideString> args = BuildCommonArgs();
+
+        if (MaxAge.HasValue)
+        {
+            args.Add(ValkeyLiterals.MAXAGE);
+            args.Add(MaxAge.Value.TotalSeconds.ToGlideString());
```
xShinnRyuu: **MAXAGE is integer-only on the wire; sending a fractional value causes a server error.**

Valkey's `CLIENT KILL` parser at `src/networking.c` (see `clientKillCommand` -> `parseClientFiltersOrReply`) parses MAXAGE via `getLongLongFromObjectOrReply(..., "maxage is not an integer or out of range")`. The command JSON confirms this: `MAXAGE` is `"type": "integer"` in `src/commands/client-kill.json`.

Here the code emits `MaxAge.Value.TotalSeconds.ToGlideString()`, which uses the `double` overload of `ToGlideString` and produces `"0.5"` for `TimeSpan.FromMilliseconds(500)`. Concrete failure:

```csharp
await client.ClientKillAsync(new ClientFilterOptions().WithMaxAge(TimeSpan.FromMilliseconds(500)));
// Sends: CLIENT KILL MAXAGE 0.5
// Server replies with an error, GLIDE surfaces it as an exception.
```

`ClientFilterOptionsTests.ToClientKillArgs_Success` currently asserts `["MAXAGE", "0.5"]` on line 29, so the unit tests bless the broken wire format and there is no integration test for MAXAGE to catch it.

Emit an integer instead (e.g. `TimeUtils.ToPositiveULongSecs(MaxAge.Value, nameof(MaxAge)).ToGlideString()`), and update `WithMaxAge` to fail eagerly when the caller passes a sub-second `TimeSpan` (or round, but say so in the doc). See also line 636 for the same bug on IDLE.

sources/Valkey.Glide/Commands/Options/ClientFilterOptions.cs
```diff
@@ -0,0 +1,722 @@
...
+        if (NotName is not null)
+        {
+            args.Add(ValkeyLiterals.NOT_NAME);
+            args.Add(NotName);
+        }
+
+        if (Idle.HasValue)
+        {
+            args.Add(ValkeyLiterals.IDLE);
+            args.Add(Idle.Value.TotalSeconds.ToGlideString());
```
xShinnRyuu: **IDLE is integer-only on the wire; same bug as MAXAGE.**

`src/networking.c` parses IDLE via `getLongLongFromObjectOrReply(..., "idle is not an integer or out of range")` and the command JSON declares IDLE as `"type": "integer"`. Sending `IDLE 0.5` yields the quoted error.

Concrete failure:

```csharp
await client.ClientKillAsync(new ClientFilterOptions().WithIdle(TimeSpan.FromMilliseconds(500)));
// Sends: CLIENT KILL IDLE 0.5 -> server error.
```

`ClientFilterOptionsTests.ToClientKillArgs_Success` currently asserts `["IDLE", "0.5"]` on line 35, so the wire format is baked into the unit tests.

Additionally, the `WithIdle` doc-comment (line 355) says "minimum idle time of connections to match" which reads as sub-second-capable, but the server's matching logic in `clientMatchesFilter` uses `commandTimeSnapshot()/1000 - client->last_interaction` (seconds granularity). Users who read the .NET signature and pass a `TimeSpan.FromMilliseconds(500)` will get either a server error today or, if the wire format is fixed to integer, a silently rounded value. Please pick one behavior and document it.

sources/Valkey.Glide/Commands/Options/ClientFilterOptions.cs
```diff
@@ -0,0 +1,722 @@
...
+        SkipMe = skipMe;
+        return this;
+    }
+
+    /// <inheritdoc cref="MaxAge" />
+    /// <param name="maxAge">The maximum connection age.</param>
+    /// <returns>This instance for method chaining.</returns>
+    public ClientFilterOptions WithMaxAge(TimeSpan maxAge)
+    {
+        ArgumentOutOfRangeException.ThrowIfLessThan(maxAge, TimeSpan.Zero, nameof(maxAge));
```
xShinnRyuu: **`WithMaxAge(TimeSpan.Zero)` is permitted client-side but rejected server-side.**

`ArgumentOutOfRangeException.ThrowIfLessThan(maxAge, TimeSpan.Zero, ...)` allows `TimeSpan.Zero`. Valkey's parser explicitly rejects zero:

```c
if (maxage <= 0) {
    addReplyError(c, "maxage should be greater than 0");
    return C_ERR;
}
```

(`src/networking.c`, in `parseClientFiltersOrReply`). So `WithMaxAge(TimeSpan.Zero)` currently reaches the server as `MAXAGE 0` and the server returns the error. The unit tests reinforce this by asserting `["MAXAGE", "0"]` at `ClientFilterOptionsTests.cs:28`, but there is no integration test to reveal the runtime failure.

Switch the guard to `ArgumentOutOfRangeException.ThrowIfLessThanOrEqual(maxAge, TimeSpan.Zero, ...)`. Same fix needed for `WithIdle` on line 359.

sources/Valkey.Glide/Commands/Options/ClientFilterOptions.cs
```diff
@@ -0,0 +1,722 @@
...
+        NotName = name;
+        return this;
+    }
+
+    /// <inheritdoc cref="Idle" />
+    /// <param name="idle">The minimum idle time of connections to match.</param>
+    /// <returns>This instance for method chaining.</returns>
+    public ClientFilterOptions WithIdle(TimeSpan idle)
+    {
+        ArgumentOutOfRangeException.ThrowIfLessThan(idle, TimeSpan.Zero, nameof(idle));
```
xShinnRyuu: **`WithIdle(TimeSpan.Zero)` accepted client-side, rejected server-side.**

`src/networking.c`:

```c
if (idle_time <= 0) {
    addReplyError(c, "idle should be greater than 0");
    return C_ERR;
}
```

Failure: `new ClientFilterOptions().WithIdle(TimeSpan.Zero)` -> `IDLE 0` -> `"idle should be greater than 0"` from the server, surfaced as an exception. `ClientFilterOptionsTests.cs:34` bakes in the broken wire arg `["IDLE", "0"]`.

Use `ArgumentOutOfRangeException.ThrowIfLessThanOrEqual(idle, TimeSpan.Zero, ...)`.

sources/Valkey.Glide/Commands/Options/ClientFilterOptions.cs
```diff
@@ -0,0 +1,722 @@
...
+        }
+        if (NotType is not null)
+        {
+            args.Add(ValkeyLiterals.NOT_TYPE);
+            args.Add(ValkeyLiterals.Get(NotType.Value));
+        }
+
+        foreach (long id in Ids)
+        {
+            args.Add(ValkeyLiterals.ID);
```
xShinnRyuu, line 553: **Emitting repeated `ID` tokens silently narrows a multi-id kill to the last id on Valkey 7.2 and 8.0.**

The multi-ID form of `CLIENT KILL ID id1 id2 ...` is `"since": "8.1.0"` per the Valkey command JSON. Before 8.1 the parser at `src/networking.c` handles the `id` filter with a single scalar assignment:

```c
if (!strcasecmp(c->argv[i]->ptr, "id") && moreargs) {
    long tmp;
    if (getRangeLongFromObjectOrReply(c, c->argv[i+1], 1, LONG_MAX, &tmp, ...) != C_OK)
        return;
    id = tmp;
}
```

So when this code emits `ID 1 ID 2 ID 3` against a 7.2 or 8.0 server, the parser sees three `id` filters in sequence and each overwrites the previous, leaving `id = 3` and killing only that one client. The caller who wrote `WithIds([1, 2, 3])` gets a silent, partial kill and cannot tell from the returned count what happened. The `WithIds` doc at line 240 says `Since Valkey 8.1.0`, but the singular `WithId` (line 233) has no version note yet is affected the moment a caller chains it twice.

Either: (a) coalesce multiple ids into a single `ID id1 id2 ...` clause and add a client-side gate that errors on pre-8.1 servers when more than one id is present; or (b) at minimum document loudly on `WithId` that chained calls behave as "last id wins" on Valkey < 8.1.

sources/Valkey.Glide/Commands/Options/ClientFilterOptions.cs
```diff
@@ -0,0 +1,722 @@
...
+    /// <returns>This instance for method chaining.</returns>
+    public ClientFilterOptions WithoutType(ClientType type)
+    {
+        NotType = type;
+        return this;
+    }
+
+    /// <inheritdoc cref="Ids" />
+    /// <returns>This instance for method chaining.</returns>
+    public ClientFilterOptions WithId(long id)
```
xShinnRyuu, line 235: **Non-positive ids (0, negative) are accepted client-side but rejected server-side.**

Valkey's parser at `src/networking.c`:

```c
if (id < 1) {
    addReplyError(c, "client-id should be greater than 0");
    return C_ERR;
}
```

Failure:

```csharp
await client.ClientKillAsync(new ClientFilterOptions().WithId(0));
// Sends: CLIENT KILL ID 0 -> server error "client-id should be greater than 0".
```

Same issue on `WithoutId` (line 250), `WithIds` (line 242), and `WithoutIds` (line 258). Guard with `ArgumentOutOfRangeException.ThrowIfLessThan(id, 1L, nameof(id))` (or equivalent on the plural overloads).

Related: `WithDatabaseId(ushort)` on line 482 accepts any value up to 65535, but Valkey rejects `db_id >= server.dbnum` (usually 16). That would surface as `"DB number should be between 0 and 15"`. Consider documenting the practical range or making the type smaller.

sources/Valkey.Glide/Abstract/ValkeyServer.cs
```diff
@@ -117,6 +117,48 @@ public async Task<long> ClientIdAsync(CommandFlags flags = CommandFlags.None)
...
+            _ = options.WithType(clientType.Value);
+        }
+
+        if (endpoint is not null)
+        {
+            (string host, ushort port) = Utils.SplitEndpoint(endpoint);
+            _ = options.WithAddress(host, port);
+        }
+
+        return await _conn.ClientKillAsync(options);
```
xShinnRyuu: **SER `IServer.ClientKillAsync` fans out to all cluster nodes instead of targeting the endpoint this `IServer` represents.**

Every other method on `ValkeyServer` passes `MakeRoute()` to scope the call to the endpoint the SER caller looked up (for example `ClientGetNameAsync` at line 110, `ClientIdAsync` at line 117, `PingAsync` at line 81, `InfoRawAsync` at line 66, and so on). This method does not:

```csharp
return await _conn.ClientKillAsync(options);
```

`_conn.ClientKillAsync` uses GLIDE's default routing for `CLIENT KILL`, which was just changed in `valkey-io/valkey-glide` PR #6709 to `AllNodes` with `ResponsePolicy::Aggregate(Sum)`. So an SER caller that does:

```csharp
var server = conn.GetServer(nodeAEndpoint);
await server.ClientKillAsync(nodeBEndpoint);  // expected: fire CLIENT KILL only against node A
```

actually fires the kill across every primary in the cluster, and if `nodeBEndpoint` happens to match a client on any other node, that client is killed. The returned count is a cluster-wide sum, not a per-node count.

This diverges from `StackExchange.Redis`' own `RedisServer.ClientKill` which builds a message with `RedisCommand.CLIENT` and executes it against the specific server (`Message.Create(-1, flags.WithCategory(NodeLocalAdmin), ...)` -> `ExecuteSync`/`ExecuteAsync` on that server), preserving the SER contract that `IServer` methods act on the specific server. The fix is to plumb `MakeRoute()` into `BaseClient.ClientKillAsync` (or expose an overload that takes a route).

sources/Valkey.Glide/Abstract/ValkeyServer.cs
```diff
@@ -117,6 +117,48 @@ public async Task<long> ClientIdAsync(CommandFlags flags = CommandFlags.None)
...
+    }
+
+    /// <inheritdoc/>
+    public async Task<long> ClientKillAsync(ClientKillFilter filter, CommandFlags flags = CommandFlags.None)
+    {
+        GuardClauses.ThrowIfCommandFlags(flags);
+        return await _conn.ClientKillAsync(ToClientFilterOptions(filter));
```
xShinnRyuu: **Same routing defect on the SER `ClientKillAsync(ClientKillFilter, CommandFlags)` overload.**

`_conn.ClientKillAsync(ToClientFilterOptions(filter))` does not carry `MakeRoute()`, so the filter-based kill fans out across the whole cluster instead of the endpoint associated with this `IServer`. See the neighboring comment on line 152 for the SER-parity reference and the concrete failure shape.

tests/Valkey.Glide.IntegrationTests/ConnectionManagementCommandTests.cs
```diff
@@ -70,82 +72,51 @@ public async Task TestClientInfo_WithClientName_ReportsName(bool clusterMode)
...
+        var targetId = await target.ClientIdAsync();
+        var killed = await client.ClientKillAsync(new ClientFilterOptions().WithId(targetId));
 
-    private static void AssertTrackingInfoOn(ClientTrackingInfo info)
-    {
-        Assert.Equivalent(new HashSet<string> { "on", "bcast" }, info.Flags);
-        Assert.Equal(0, info.Redirect);
-        Assert.Equivalent(new HashSet<string> { "" }, info.Prefixes);
+        Assert.Equal(1, killed);
     }
```
xShinnRyuu, line 124: **Integration test coverage stops at `ID` and `ADDR`; the risky new filters have no end-to-end signal.**

All four `ClientKillAsync_*` tests in this file exercise only `WithId` or `WithAddress` (which are 7.2 features). There is no integration test that:

- kills a client via `WithMaxAge` (which today sends a fractional wire arg and hits the server error described on `ClientFilterOptions.cs:543`),
- kills a client via `WithIdle` (same class of bug),
- exercises any 9.0-only filter (`WithName`, `WithFlag(s)`, `WithLibraryName`, `WithLibraryVersion`, `WithDatabaseId`, `WithCapability`, `WithIpAddress`, or any `Without*` counterpart),
- exercises multi-id `WithIds` against an 8.1+ server.

Because the unit tests bake in wire arguments that Valkey will reject at runtime (e.g. `MAXAGE 0.5`, `MAXAGE 0`, `IDLE 0`, `IDLE 0.5`), and no integration test rebuilds those wire arguments end-to-end, the current suite provides no signal for the wire-format bugs. `TestConfiguration.IsVersionAtLeast("8.0.0")` / `"9.0.0"` already exists (`TestConfiguration.cs:51`), so version-gated integration tests are cheap to add. At minimum: one MAXAGE and one IDLE integration test on 8.0+ and 9.0+ respectively.

tests/Valkey.Glide.UnitTests/Options/ClientFilterOptionsTests.cs
```diff
@@ -0,0 +1,178 @@
...
+        Assert.Equal(["USER", "user"], new ClientFilterOptions().WithUser("user").ToClientKillArgs());
+        Assert.Equal(["ADDR", "addr:1234"], new ClientFilterOptions().WithAddress("addr", 1234).ToClientKillArgs());
+        Assert.Equal(["ADDR", "127.0.0.1:1234"], new ClientFilterOptions().WithAddress("127.0.0.1", 1234).ToClientKillArgs());
+        Assert.Equal(["ADDR", "[::1]:1234"], new ClientFilterOptions().WithAddress("::1", 1234).ToClientKillArgs());
+        Assert.Equal(["LADDR", "laddr:4321"], new ClientFilterOptions().WithLocalAddress("laddr", 4321).ToClientKillArgs());
+        Assert.Equal(["LADDR", "127.0.0.1:4321"], new ClientFilterOptions().WithLocalAddress("127.0.0.1", 4321).ToClientKillArgs());
+        Assert.Equal(["LADDR", "[::1]:4321"], new ClientFilterOptions().WithLocalAddress("::1", 4321).ToClientKillArgs());
+        Assert.Equal(["SKIPME", "yes"], new ClientFilterOptions().WithSkipMe(true).ToClientKillArgs());
+        Assert.Equal(["SKIPME", "no"], new ClientFilterOptions().WithSkipMe(false).ToClientKillArgs());
+        Assert.Equal(["MAXAGE", "0"], new ClientFilterOptions().WithMaxAge(TimeSpan.Zero).ToClientKillArgs());
```
xShinnRyuu: **Unit tests assert wire arguments that Valkey will refuse.**

Lines 28-29 assert `["MAXAGE", "0"]` and `["MAXAGE", "0.5"]`, and lines 34-35 assert `["IDLE", "0"]` and `["IDLE", "0.5"]`. Both wire formats are wrong for Valkey's parser (see the anchored comments on `ClientFilterOptions.cs:543` and `:636`). The tests will pass in CI because they only check what the builder emits, but the moment either filter reaches a real server the server will error out with `"maxage is not an integer or out of range"` / `"maxage should be greater than 0"` etc.

Once the fixes on `ClientFilterOptions.cs` land (integer wire arg + `ThrowIfLessThanOrEqual` guard), these assertions should flip to expect integer wire args (and the zero cases should become throwing-cases in `ToClientKillArgs_Failure`).

sources/Valkey.Glide/Commands/Options/ClientFilterOptions.cs
currantw: Good catch. Addressed in a410bca6.

sources/Valkey.Glide/Commands/Options/ClientFilterOptions.cs
currantw: Good catch. Addressed in a410bca6.

sources/Valkey.Glide/Commands/Options/ClientFilterOptions.cs
currantw: Good catch. Addressed in a410bca6.

sources/Valkey.Glide/Commands/Options/ClientFilterOptions.cs
currantw: Good catch. Addressed in a410bca6.

tests/Valkey.Glide.UnitTests/Options/ClientFilterOptionsTests.cs
currantw: Good catch. Addressed in a410bca6.

sources/Valkey.Glide/Commands/Options/ClientFilterOptions.cs
currantw: Good catch on the version semantics. Looking at the Valkey source, on 8.1+ the parser uses an `intset` and reads multiple IDs after a single `ID` token (`ID 1 2 3`), but our code emits repeated `ID 1 ID 2 ID 3` — which also works since the parser accepts the `ID` keyword multiple times and adds to the same intset.

On pre-8.1, the parser used a scalar assignment (`id = tmp`) so repeated `ID` tokens would overwrite, leaving only the last. This is already documented on `WithIds` ("Since Valkey 8.1.0"), but you're right that chaining `WithId` multiple times has the same effect. I'll add a doc note to `WithId` clarifying this.

Note: we can't detect server version at build time, so a runtime guard isn't feasible here. The approach is consistent with other GLIDE language clients.

currantw: Updated in 5a6cb48f. `WithId` and `WithIds` now overwrite any previous values, which matches the behaviour of other `With*` methods, and `WithIds` is clearly marked at 8.0+ only. Beyond that, there is no validation. I think this is pretty consistent with our general behaviour, in that we don't validate server compatibility issues for the user – we just try to make sure we clearly spell them out in the documentation.

sources/Valkey.Glide/Commands/Options/ClientFilterOptions.cs
currantw: While this is true, I would suggest that this is something we _should_ handle on the server side: I think the general approach is to pass values through to the server without any extra validation. (The `TimeSpan` cases are different in my view, because we _already_ need to cast the intervals to raw numbers before passing to server).

sources/Valkey.Glide/Abstract/ValkeyServer.cs
currantw: Updated to use `MakeRoute` in e69f05f4.

sources/Valkey.Glide/Abstract/ValkeyServer.cs
currantw: Updated to use `MakeRoute` in e69f05f4.

tests/Valkey.Glide.IntegrationTests/ConnectionManagementCommandTests.cs
currantw: Hmmmm. I understand the desire to test this comprehensively... but, in this case, I think that would be a lot more effort than it is worth:

1. There are _a lot_ of methods (35 by my count).
2. Many of them would be quite tricky to setup. For example, testing MaxAge would require us to create a new client, wait for it be be at least one second old, and then kill it using another client. Nothing _crazy_, but certainly not trivial. And some of the other methods would be significantly more tricky to test.

Given that, I see these integration tests as proof that this command works end to end, rather than comprehensive testing of every permutation. I think that can be left to the unit tests for `ClientFilterOptions`. Are there any specific cases you think are missing there that could be added to make sure all variations are tests?

affonsov (approved):

PR #514, opened 2026-08-07
feat: add LibName and ClientInfoTag to client configuration

Jonathan-Improving:
## Summary

Adds two client-identification options — reported via `CLIENT SETINFO LIB-NAME` — to the standalone, cluster, **and MONITOR** clients:

1. **`LibraryName`** — full override for the lib-name value. Defaults to `GlideC#`; when set, replaces that identifier entirely.
2. **`ClientInfoTag`** — appends a parenthesized tag to the resolved library name while preserving the underlying GLIDE identity, e.g. `GlideC#(my-framework:1.0)`.

Available on `ConnectionConfiguration` (standalone/cluster) and on `MonitorConfig` (MONITOR connection).

## Motivation

Libraries and frameworks embedding GLIDE need to identify themselves to the server for adoption tracking, observability, and debugging. `ClientInfoTag` is the recommended option: it keeps the `GlideC#` token intact for GLIDE-level dashboards while adding per-framework attribution.

## Behavior matrix

| Config | Resulting `lib-name` |
|---|---|
| _(none)_ | `GlideC#` |
| `ClientInfoTag="lmcache:1.2"` | `GlideC#(lmcache:1.2)` |
| `ClientInfoTag="   "` _(whitespace-only)_ | `GlideC#` |
| `LibraryName="custom"` | `custom` |
| `LibraryName="custom"`, `ClientInfoTag="lmcache:1.2"` | `custom(lmcache:1.2)` |

## What changed

- **`ConnectionConfiguration.cs`** — Add `LibraryName` and `ClientInfoTag` builder options (`WithLibraryName`, `WithClientInfoTag`), backed by the `ConnectionConfig` record with a `ResolvedLibName` computed property. No client-side validation is performed on the tag — validation is deferred to the server. As output-composition hygiene, a whitespace-only tag degrades to the base lib-name (so `CLIENT SETINFO LIB-NAME`, which rejects spaces/newlines/special characters, never receives a broken value). `LibraryName` exposes the `GlideC#` default via its getter.
- **`MonitorConfig.cs` / `MonitorClient.cs`** — Add the same `LibraryName` / `ClientInfoTag` options (+ `WithLibraryName` / `WithClientInfoTag`) to the MONITOR client, and send the resolved lib-name for the monitor connection instead of the hard-coded `GLIDE_NAME`.
- **`Internals/LibNameResolver.cs`** (new) — Shared composition (base name + optional tag, incl. whitespace-only handling) used by both `ConnectionConfiguration` and `MonitorConfig`, so all connection types stay consistent.
- **`Internals/FFI.structs.cs`** — Add the non-optional `LibName` field (`LPUTF8Str`) to the `ConnectionRequest` struct and to `MonitorConfigFfi`. The resolved lib-name is always provided.
- **`rust/src/ffi.rs` / `rust/src/lib.rs`** — Add required `lib_name` to the connection and monitor FFI structs; always send the C#-resolved value (no compile-time `GLIDE_NAME` fallback — glide-core honors the request `lib_name`). glide-core is untouched.
- **Tests** — Unit tests for `ConnectionConfiguration` and `MonitorConfig` (defaults, overrides, tag composition, whitespace-only, cluster); integration tests asserting `CLIENT INFO` (standard/cluster) and `CLIENT LIST` (monitor connection) report the expected lib-name.

## Testing

- Unit tests green (671 total).
- Integration tests (standalone + cluster) verify `CLIENT INFO` reports the expected `lib-name`; a MONITOR integration test verifies the monitor connection's `lib-name` via `CLIENT LIST` (gated to Valkey 7.2+ for `CLIENT SETINFO`).
- Follows the same pattern as the merged Python PR (valkey-io/valkey-glide#6389).

## Notes

- **LIB-NAME only.** `LIB-VER` remains core-controlled via `GLIDE_VERSION`.
- **No client-level tag validation** — deferred to the server per review discussion; only whitespace-only tags are normalized away at composition.
- The `lib_name` field already exists in the shared `ConnectionRequest` (used by other language clients). This change wires C# (including the MONITOR connection) to use it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support additional configuration options:
   - All nodes read-from strategy (#207)
   - Circuit breaker (#474)
+  - Client library name and info tag (#514)
   - Inflight requests limit (#484)
   - Mutual TLS (#488)
   - Periodic topology checks (#485)
--- a/rust/src/ffi.rs
+++ b/rust/src/ffi.rs
@@ -28,7 +28,7 @@ use redis::{
 /// # Safety
 ///
 /// * `ptr` must be able to be safely casted to a valid [`CStr`] via [`CStr::from_ptr`]. See the safety documentation of [`std::ffi::CStr::from_ptr`].
-unsafe fn ptr_to_str(ptr: *const c_char) -> Result<String, String> {
+pub(crate) unsafe fn ptr_to_str(ptr: *const c_char) -> Result<String, String> {
     if !ptr.is_null() {
         unsafe { CStr::from_ptr(ptr) }
             .to_str()
@@ -155,6 +155,9 @@ pub struct ConnectionConfig {
     pub protocol: redis::ProtocolVersion,
     /// zero pointer is valid, means no client name is given (`None`)
     pub client_name: *const c_char,
+    /// Library name override for CLIENT SETINFO LIB-NAME. Always provided by the C# layer
+    /// (defaults to "GlideC#"); must be non-null.
+    pub lib_name: *const c_char,
     pub lazy_connect: bool,
     pub refresh_topology_from_initial_nodes: bool,
     pub pubsub_config: PubSubConfigInfo,
@@ -311,7 +314,7 @@ pub(crate) unsafe fn create_connection_request(
             None
         },
         client_name: unsafe { ptr_to_opt_str(config.client_name) }?,
-        lib_name: Some(env!("GLIDE_NAME").to_string()),
+        lib_name: Some(unsafe { ptr_to_str(config.lib_name) }?),
         authentication_info: if config.has_authentication_info {
             let auth_info = config.authentication_info;
             let iam_config = if auth_info.has_iam_credentials {
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -2039,6 +2039,7 @@ pub struct MonitorConfig {
     pub database: u16,
     pub username: *const c_char,
     pub password: *const c_char,
+    pub lib_name: *const c_char,
 }
 
 struct MonitorAdapter {
@@ -2074,16 +2075,26 @@ pub unsafe extern "C-unwind" fn create_monitor_client(
 ) -> *const MonitorConnectionResponse {
     let config = unsafe { &*config };
 
-    let host = match unsafe { CStr::from_ptr(config.host) }.to_str() {
-        Ok(s) => s.to_owned(),
-        Err(e) => {
-            let err_msg = CString::new(format!("Invalid UTF-8 in host: {e}")).unwrap_or_default();
-            return Box::into_raw(Box::new(MonitorConnectionResponse {
-                conn_ptr: std::ptr::null(),
-                connection_error_message: err_msg.into_raw(),
-            }));
-        }
-    };
+    // Converts a required C string field to a Rust `String`, or returns an error response
+    // immediately if it is not valid UTF-8. Shared by `host` and `lib_name`, which are the two
+    // required (non-optional) string fields on `MonitorConfig`.
+    macro_rules! required_field {
+        ($ptr:expr, $field_name:literal) => {
+            match unsafe { crate::ffi::ptr_to_str($ptr) } {
+                Ok(s) => s,
+                Err(e) => {
+                    let err_msg = CString::new(format!("Invalid UTF-8 in {}: {e}", $field_name))
+                        .unwrap_or_default();
+                    return Box::into_raw(Box::new(MonitorConnectionResponse {
+                        conn_ptr: std::ptr::null(),
+                        connection_error_message: err_msg.into_raw(),
+                    }));
+                }
+            }
+        };
+    }
+
+    let host = required_field!(config.host, "host");
 
     let address = NodeAddress {
         host,
@@ -2114,13 +2125,15 @@ pub unsafe extern "C-unwind" fn create_monitor_client(
         }
     };
 
+    let lib_name = required_field!(config.lib_name, "lib_name");
+
     let redis_conn_info = redis::RedisConnectionInfo {
         db: config.database as i64,
         username,
         password,
         protocol: redis::ProtocolVersion::RESP2,
         client_name: None,
-        lib_name: Some(env!("GLIDE_NAME").to_string()),
+        lib_name: Some(lib_name),
         server_assisted_cache: false,
         cache: None,
     };
--- a/sources/Valkey.Glide/ConnectionConfiguration.cs
+++ b/sources/Valkey.Glide/ConnectionConfiguration.cs
@@ -52,6 +52,8 @@ public abstract class ConnectionConfiguration
         public uint DatabaseId;
         public Protocol? Protocol;
         public string? ClientName;
+        public string? LibName;
+        public string? ClientInfoTag;
         public bool LazyConnect;
         public bool RefreshTopologyFromInitialNodes;
         public BasePubSubSubscriptionConfig? PubSubSubscriptions;
@@ -82,6 +84,9 @@ public abstract class ConnectionConfiguration
         public PeriodicChecksMode? PeriodicChecksMode;
         public uint? PeriodicChecksIntervalSecs;
 
+        /// <summary>
+        /// Computes the resolved library name to send via CLIENT SETINFO LIB-NAME.
+        /// </summary>
         internal FFI.ConnectionConfig ToFfi() => new(
             Addresses,
             ClusterMode,
@@ -120,7 +125,10 @@ public abstract class ConnectionConfiguration
 
             // Periodic checks
             PeriodicChecksMode,
-            PeriodicChecksIntervalSecs
+            PeriodicChecksIntervalSecs,
+
+            // CLIENT SETINFO LIB-NAME
+            Internals.Utils.ResolveLibraryName(LibName, ClientInfoTag)
         );
     }
 
@@ -866,6 +874,60 @@ public abstract class ConnectionConfiguration
         public T WithClientName(string? clientName)
         {
             ClientName = clientName;
+            return (T)this;
+        }
+
+        #endregion
+        #region Library Name
+
+        /// <summary>
+        /// Full override for the library name reported via <c>CLIENT SETINFO LIB-NAME</c> during connection establishment.
+        /// Defaults to <c>GlideC#</c>. When set, this replaces the default identifier entirely.
+        /// <para/>
+        /// Use <see cref="ClientInfoTag"/> instead if you want to preserve the GLIDE identity while adding a framework tag.
+        /// </summary>
+        /// <seealso cref="ClientInfoTag"/>
+        public string? LibraryName
+        {
+            get => Config.LibName ?? Internals.Utils.DefaultLibraryName;
+            set => Config.LibName = value;
+        }
+
+        /// <inheritdoc cref="LibraryName" />
+        public T WithLibraryName(string? libraryName)
+        {
+            LibraryName = libraryName;
+            return (T)this;
+        }
+
+        #endregion
+        #region Client Info Tag
+
+        /// <summary>
+        /// A tag appended to the library name reported via <c>CLIENT SETINFO LIB-NAME</c>.
+        /// The tag is appended in parentheses to preserve the underlying GLIDE identity for adoption tracking.
+        /// <para/>
+        /// For example, setting <c>ClientInfoTag = "my-framework:1.0"</c> produces <c>GlideC#(my-framework:1.0)</c>.
+        /// <para/>
+        /// A <see langword="null"/> value is treated as absent: the base library name is reported
+        /// unchanged, with no <c>()</c> suffix. Any other value — including empty or
+        /// whitespace-only — is passed through as supplied. Character validation is performed
+        /// entirely by the GLIDE core, which permits only printable ASCII from <c>!</c> through
+        /// <c>~</c> (excluding space) plus at most one matched, non-empty trailing <c>(tag)</c>
+        /// group; a malformed value fails client creation with a configuration error rather than
+        /// being silently dropped.
+        /// </summary>
+        /// <seealso cref="LibraryName"/>
+        public string? ClientInfoTag
+        {
+            get => Config.ClientInfoTag;
+            set => Config.ClientInfoTag = value;
+        }
+
+        /// <inheritdoc cref="ClientInfoTag" />
+        public T WithClientInfoTag(string? clientInfoTag)
+        {
+            ClientInfoTag = clientInfoTag;
             return (T)this;
         }
 
--- a/sources/Valkey.Glide/Internals/FFI.structs.cs
+++ b/sources/Valkey.Glide/Internals/FFI.structs.cs
@@ -34,6 +34,7 @@ internal partial class FFI
             if (_ptr != IntPtr.Zero)
             {
                 FreeMemory();
+                DestroyStruct(_ptr);
                 FreeStructPtr(_ptr);
                 _ptr = IntPtr.Zero;
             }
@@ -43,6 +44,15 @@ internal partial class FFI
         protected abstract IntPtr AllocateAndCopy();
 
         protected abstract void FreeMemory();
+
+        /// <summary>
+        /// Releases unmanaged memory the runtime marshaller allocated for reference-typed struct
+        /// fields (e.g. <see cref="UnmanagedType.LPUTF8Str"/> strings), which <see cref="FreeMemory"/>
+        /// (manual <see cref="IntPtr"/> frees) does not cover. No-op by default; override for structs
+        /// that contain such fields. Called after <see cref="FreeMemory"/> and before the block is freed.
+        /// </summary>
+        /// <param name="ptr">Pointer to the marshalled struct whose reference-typed fields are released.</param>
+        protected virtual void DestroyStruct(IntPtr ptr) { }
     }
 
     // A wrapper for a command, resposible for marshalling (allocating and freeing) the required data
@@ -213,6 +223,12 @@ internal partial class FFI
         /// </summary>
         internal NodeDiscoveryMode NodeDiscoveryMode => _request.NodeDiscoveryMode;
 
+        /// <summary>
+        /// The resolved library name marshalled into the underlying FFI request. Exposed for testing
+        /// that <see cref="Utils.ResolveLibraryName"/> is correctly wired through to the FFI layer.
+        /// </summary>
+        internal string ResolvedLibName => _request.ResolvedLibName;
+
         public ConnectionConfig(
             List<NodeAddress> addresses,
             bool clusterMode,
@@ -251,7 +267,10 @@ internal partial class FFI
 
             // Periodic checks
             PeriodicChecksMode? periodicChecksMode,
-            uint? periodicChecksIntervalSec)
+            uint? periodicChecksIntervalSec,
+
+            // Resolved library name
+            string resolvedLibName)
         {
             _request = new()
             {
@@ -313,6 +332,9 @@ internal partial class FFI
                 HasPeriodicChecksConfig = periodicChecksMode.HasValue,
                 PeriodicChecksMode = periodicChecksMode ?? default,
                 PeriodicChecksIntervalSec = periodicChecksIntervalSec ?? 0,
+
+                // CLIENT SETINFO LIB-NAME
+                ResolvedLibName = resolvedLibName,
             };
         }
 
@@ -380,6 +402,9 @@ internal partial class FFI
 
         protected override IntPtr AllocateAndCopy()
             => StructToPtr(_request);
+
+        protected override void DestroyStruct(IntPtr ptr)
+            => Marshal.DestroyStructure(ptr, typeof(ConnectionRequest));
 
         /// <summary>
         /// Marshals the node addresses.
@@ -739,6 +764,13 @@ internal partial class FFI
 
         [MarshalAs(UnmanagedType.LPUTF8Str)]
         public string? ClientName;
+
+        /// <summary>
+        /// The fully composed value sent to <c>CLIENT SETINFO LIB-NAME</c> (already includes any
+        /// <c>ClientInfoTag</c>). Named distinctly from the raw <c>LibName</c> to avoid re-composing.
+        /// </summary>
+        [MarshalAs(UnmanagedType.LPUTF8Str)]
+        public string ResolvedLibName;
 
         [MarshalAs(UnmanagedType.U1)]
         public bool LazyConnect;
@@ -1281,6 +1313,12 @@ internal partial class FFI
         /// or <see langword="null"/> if not set.</summary>
         [MarshalAs(UnmanagedType.LPUTF8Str)]
         public string? Password;
+
+        /// <summary>
+        /// The resolved library name reported via CLIENT SETINFO LIB-NAME for the MONITOR connection.
+        /// </summary>
+        [MarshalAs(UnmanagedType.LPUTF8Str)]
+        public string LibName;
     }
 
     /// <summary>
--- a/sources/Valkey.Glide/MonitorClient.cs
+++ b/sources/Valkey.Glide/MonitorClient.cs
@@ -63,6 +63,7 @@ public sealed class MonitorClient : IAsyncDisposable, IDisposable
             Database = config.Database,
             Username = config.Username,
             Password = config.Password is not null ? new string(config.Password) : null,
+            LibName = Internals.Utils.ResolveLibraryName(config.LibraryName, config.ClientInfoTag),
         };
 
         IntPtr configPtr = Marshal.AllocHGlobal(Marshal.SizeOf<MonitorConfigFfi>());
--- a/tests/Valkey.Glide.IntegrationTests/ClientInfoTagTests.cs
+++ b/tests/Valkey.Glide.IntegrationTests/ClientInfoTagTests.cs
@@ -1,0 +1,171 @@
+// Copyright Valkey GLIDE Project Contributors - SPDX Identifier: Apache-2.0
+
+using Valkey.Glide.TestUtils;
+
+namespace Valkey.Glide.IntegrationTests;
+
+/// <summary>
+/// Integration tests verifying CLIENT INFO reports the correct lib-name
+/// when LibraryName and/or ClientInfoTag are configured.
+/// Requires Valkey server >= 7.2.0 (CLIENT SETINFO support).
+/// </summary>
+[Collection("GlideTests")]
+public class ClientInfoTagTests(TestConfiguration config)
+{
+    public TestConfiguration Config { get; } = config;
+
+    /// <summary>
+    /// LibraryName override only → CLIENT INFO contains lib-name=custom-client
+    /// </summary>
+    [Theory(DisableDiscoveryEnumeration = true)]
+    [MemberData(nameof(Data.ClusterMode), MemberType = typeof(Data))]
+    public async Task TestClientInfo_WithLibNameOnly_ReportsCustomLibName(bool useCluster)
+    {
+        Skip.IfClientSetInfoNotSupported();
+
+        const string customLibName = "custom-client";
+
+        using BaseClient client = useCluster
+            ? await GlideClusterClient.CreateClient(
+                TestConfiguration.DefaultClusterClientConfig()
+                    .WithLibraryName(customLibName)
+                    .Build())
+            : await GlideClient.CreateClient(
+                TestConfiguration.DefaultClientConfig()
+                    .WithLibraryName(customLibName)
+                    .Build());
+
+        string info = await Client.GetClientInfo(client);
+
+        Assert.Contains($"lib-name={customLibName}", info);
+        // Should NOT have parenthesized tag when no ClientInfoTag is set
+        Assert.DoesNotContain($"lib-name={customLibName}(", info);
+    }
+
+    /// <summary>
+    /// ClientInfoTag only → CLIENT INFO contains lib-name=GlideC#(tag)
+    /// </summary>
+    [Theory(DisableDiscoveryEnumeration = true)]
+    [MemberData(nameof(Data.ClusterMode), MemberType = typeof(Data))]
+    public async Task TestClientInfo_WithClientInfoTagOnly_ReportsDefaultLibNameWithTag(bool useCluster)
+    {
+        Skip.IfClientSetInfoNotSupported();
+
+        const string tag = "my-framework:1.0";
+
+        using BaseClient client = useCluster
+            ? await GlideClusterClient.CreateClient(
+                TestConfiguration.DefaultClusterClientConfig()
+                    .WithClientInfoTag(tag)
+                    .Build())
+            : await GlideClient.CreateClient(
+                TestConfiguration.DefaultClientConfig()
+                    .WithClientInfoTag(tag)
+                    .Build());
+
+        string info = await Client.GetClientInfo(client);
+
+        Assert.Contains($"lib-name=GlideC#({tag})", info);
+    }
+
+    /// <summary>
+    /// Both LibraryName and ClientInfoTag → CLIENT INFO contains lib-name=custom-client(tag)
+    /// </summary>
+    [Theory(DisableDiscoveryEnumeration = true)]
+    [MemberData(nameof(Data.ClusterMode), MemberType = typeof(Data))]
+    public async Task TestClientInfo_WithLibNameAndClientInfoTag_ReportsCombined(bool useCluster)
+    {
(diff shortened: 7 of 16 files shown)


valkey-glide
currantw, line 1: The `valkey-glide` submodule was recently bumped in #529, so this change may not be necessary any more.

sources/Valkey.Glide/ConnectionConfiguration.cs
```diff
@@ -866,6 +890,63 @@ public T WithClientName(string? clientName)
...
+            set
+            {
+                if (value is not null && value.Any(char.IsWhiteSpace))
+                {
+                    throw new ArgumentException("ClientInfoTag must not contain whitespace characters.", nameof(value));
+                }
+
+                Config.ClientInfoTag = value;
+            }
+        }
```
currantw: For this client, we are generally trying to avoid extra validation of user input at the client level, and leave this for the server (though I'm sure we aren't totally consistent in this everywhere). See for example `ClientName` and `WithClientName`.


```suggestion
            set => Config.ClientInfoTag = value;
```

rust/src/ffi.rs
```diff
@@ -311,7 +315,10 @@ pub(crate) unsafe fn create_connection_request(
             None
         },
         client_name: unsafe { ptr_to_opt_str(config.client_name) }?,
-        lib_name: Some(env!("GLIDE_NAME").to_string()),
+        // Fallback: C# always provides ResolvedLibName, but kept defensively
+        // in case the FFI contract changes.
+        lib_name: unsafe { ptr_to_opt_str(config.lib_name) }?
+            .or_else(|| Some(env!("GLIDE_NAME").to_string())),
```
currantw: Should we just remove dependency on `GLIDE_NAME` completely? To me, this is potentially misleading since `GLIDE_NAME` will never actually be used, right?

rust/src/ffi.rs
currantw, line 1: I think we may also need to update the library name for the MONITOR client? We might also want to consider adding similar client name and tag methods to `MonitorConfig`?

sources/Valkey.Glide/Internals/FFI.structs.cs
```diff
@@ -244,7 +244,10 @@ public ConnectionConfig(
 
             // Periodic checks
             PeriodicChecksMode? periodicChecksMode,
-            uint? periodicChecksIntervalSec)
+            uint? periodicChecksIntervalSec,
+
+            // CLIENT SETINFO LIB-NAME
+            string? libName)
```
currantw: Similar to other comment above, I think we should make this non-optional if there isn't a current scenario where it would be null. This is internal, so we can always make it optional in the future if needed, right?

sources/Valkey.Glide/Internals/FFI.structs.cs
```diff
@@ -811,6 +817,12 @@ private struct ConnectionRequest
         public PeriodicChecksMode PeriodicChecksMode;
         public uint PeriodicChecksIntervalSec;
 
+        #endregion
+        #region CLIENT SETINFO LIB-NAME
```
currantw: Nit. Just matching pattern above. 🤷 


```suggestion
        #region Library Name
```

sources/Valkey.Glide/ConnectionConfiguration.cs
```diff
@@ -866,6 +890,63 @@ public T WithClientName(string? clientName)
...
+        /// When set, this replaces the default <c>GlideC#</c> identifier entirely.
+        /// <para/>
+        /// Use <see cref="ClientInfoTag"/> instead if you want to preserve the GLIDE identity while adding a framework tag.
+        /// </summary>
+        /// <seealso cref="ClientInfoTag"/>
+        public string? LibName
+        {
+            get => Config.LibName;
+            set => Config.LibName = value;
+        }
```
currantw: Would it make sense to just default to "GlideC#" directly here? I don't see any value in hiding this from users?


```suggestion
        public string? LibName
        {
            get => Config.LibName;
            set => Config.LibName = value;
        } = DefaultLibName;
```

tests/Valkey.Glide.IntegrationTests/ClientInfoTagTests.cs
```diff
@@ -0,0 +1,116 @@
...
+{
+    private static readonly Version Valkey7_2 = new("7.2.0");
+    private static readonly GlideString[] InfoCommand = ["CLIENT", "INFO"];
+
+    public TestConfiguration Config { get; } = config;
+
+    private static void SkipIfClientSetInfoNotSupported()
+        => Assert.SkipWhen(
+            TestConfiguration.SERVER_VERSION < Valkey7_2,
+            "CLIENT SETINFO requires Valkey 7.2+");
```
currantw: We've got a `Skip` module in `TestUtils` with similar logic. Probably worth moving this version/skip logic there?

tests/Valkey.Glide.IntegrationTests/ClientInfoTagTests.cs
```diff
@@ -0,0 +1,116 @@
...
+    public async Task TestClientInfo_WithLibNameOnly_ReportsCustomLibName(bool useCluster)
+    {
+        SkipIfClientSetInfoNotSupported();
+
+        const string customLibName = "custom-client";
+
+        using BaseClient client = useCluster
+            ? await GlideClusterClient.CreateClient(
+                TestConfiguration.DefaultClusterClientConfig()
+                    .WithLibName(customLibName)
```
currantw: The C# client _generally_ uses more verbose names without abbreviated (again, I'm sure there are plenty of counter-examples). As such, perhaps we should go with `WithLibraryName`?


```suggestion
                    .WithLibraryName(customLibName)
```

tests/Valkey.Glide.IntegrationTests/ClientInfoTagTests.cs
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    private static async Task<string> GetClientInfo(BaseClient client)
+    {
+        object? result = client is GlideClusterClient clusterClient
+            ? (await clusterClient.CustomCommand(InfoCommand, Route.Random)).SingleValue
+            : await ((GlideClient)client).CustomCommand(InfoCommand);
+
+        return result!.ToString()!;
+    }
```
currantw: Perhaps this could live in `TestUtils.Client`? There is already some similar logic in there for `GetVersion`.

currantw (commented):
Reviewed to bec61e8.

Thanks @Jonathan-Improving! Please let me know if anything is unclear or you want to discuss in more detail. Please hit the "Re-request review" button when it's ready for another look and I'll try to get to it ASAP.

currantw:
@Jonathan-Improving please let us know if you've got the bandwidth to complete this. Otherwise, I am happy to take over and get it across the finish line 🏁

currantw:
Also occurred to me that the monitor client also uses the GLIDE_NAME env var for its client name.

Jonathan-Improving:
Still working on this @currantw please standby...

Jonathan-Improving:
### F1b: whitespace-only `ClientInfoTag` no longer drops the whole lib-name (`2613d81`)

Follow-up to removing the client-side whitespace guard on the `ClientInfoTag` setter (per @currantw's "defer validation to the server").

**Problem:** `ResolvedLibName` used `string.IsNullOrEmpty(ClientInfoTag)`, so a whitespace-only tag (e.g. `"   "`) composed `GlideC#(   )`. `CLIENT SETINFO LIB-NAME` rejects any value containing spaces/newlines/special characters (`validateClientAttr` in valkey `networking.c`), and glide-core issues SETINFO with `.ignore()` and discards the pipeline result (`connection.rs`: *"result is ignored"*). Net effect: the **entire** lib-name is silently dropped — you lose `GlideC#` too.

**Fix:** use `string.IsNullOrWhiteSpace(ClientInfoTag)` so a whitespace-only tag degrades to the plain base name (`GlideC#`, or a custom `LibraryName`) with no parentheses. This is **output-composition hygiene**, not input validation, so it does not reintroduce a client-side guard on the setter — a tag with *internal* spaces still flows through to the server and is rejected there, consistent with defer-to-server.

Added unit tests for whitespace-only tags (standalone, cluster, and with a custom library name). All 660 unit tests pass.

@currantw — could you confirm you agree with treating a **whitespace-only** tag as *absent* (degrade to base lib-name), rather than the strict defer-to-server behaviour where the whole `CLIENT SETINFO LIB-NAME` is dropped? Happy to change if you'd prefer the latter. Re-requesting review.

sources/Valkey.Glide/ConnectionConfiguration.cs
Jonathan-Improving: Confirming this stays as you directed: the client-side whitespace guard on the `ClientInfoTag` setter remains **removed** (defer validation to the server), as of `e9d5013`. I won't re-add it.

To keep the server from ever receiving a broken lib-name, the follow-up `2613d81` (F1b) handles the one pathological case at **composition** time only: a whitespace-only tag (e.g. `"   "`) now degrades to the plain base lib-name instead of composing `GlideC#(   )` — which `CLIENT SETINFO LIB-NAME` would reject, causing glide-core to silently drop the entire lib-name. A tag with *internal* spaces still passes through untouched and is validated by the server, consistent with defer-to-server.

Jonathan-Improving:
@currantw good catch — you're right that the MONITOR client also derives its client name from the `GLIDE_NAME` env var, so it wouldn't pick up `LibraryName` / `ClientInfoTag` today.

I'd like to keep this PR scoped to the standard/cluster `client_info_tag` feature (its contract), and handle MONITOR lib-name/tag support (wiring `MonitorConfig` + the monitor client's lib-name through the same resolved value) as a **separate, tracked follow-up** issue/PR. I'll open that follow-up so it isn't lost. Does that split work for you?

valkey-glide
Jonathan-Improving: Kept as-is (not reverted) — intentional. Main's submodule revision (`9a268ea`, from #529) does **not** yet include the glide-core fix `a16f3ad` *"Flip GLIDE_NAME precedence so request lib_name wins over compile-time env"*, which `CLIENT SETINFO LIB-NAME` needs in order to honor the request's lib-name over the compiled `GLIDE_NAME`. Reverting to main's revision would silently drop any custom lib-name/tag. The required revision is retained via `6480614`. (Follow-up: upstream that fix to glide-core `main`, then re-bump.) **Not resolved by reverting — deliberate.**

rust/src/ffi.rs
Jonathan-Improving: Done in `6c70ae9` (`rust/src/ffi.rs`): removed the `GLIDE_NAME` env fallback. The C# layer always supplies the resolved lib-name, so the FFI now sends it unconditionally (`lib_name: Some(ptr_to_str(config.lib_name)?)`). **Resolved.**

rust/src/ffi.rs
Jonathan-Improving: Deferring MONITOR-client lib-name/tag support (and `MonitorConfig` methods) as a separate, tracked follow-up so this PR stays scoped to the standard/cluster feature — see #514 (comment) https://github.com/valkey-io/valkey-glide-csharp/pull/514#issuecomment-5372253544. **Not addressed here by design (tracked follow-up).**

sources/Valkey.Glide/Internals/FFI.structs.cs
Jonathan-Improving: Done in `6c70ae9`: `lib_name` is now non-optional across the FFI constructor param, the `ConnectionRequest` struct field, and `ResolvedLibName` (always non-null). **Resolved.**

sources/Valkey.Glide/Internals/FFI.structs.cs
Jonathan-Improving: Done in `6c70ae9`: renamed to `#region Library Name` to match the pattern above. **Resolved.**

sources/Valkey.Glide/ConnectionConfiguration.cs
Jonathan-Improving: Done in `6c70ae9`: the `LibraryName` getter now surfaces the default (`get => Config.LibName ?? DefaultLibName`), so `GlideC#` is visible to users rather than hidden. **Resolved.**

tests/Valkey.Glide.IntegrationTests/ClientInfoTagTests.cs
Jonathan-Improving: Done in `6c70ae9`: moved the version gate into the IntegrationTests `Skip` module as `Skip.IfClientSetInfoNotSupported()`. **Resolved.**

tests/Valkey.Glide.IntegrationTests/ClientInfoTagTests.cs
Jonathan-Improving: Done in `6c70ae9`: renamed to the verbose form — `LibName` → `LibraryName` and `WithLibName` → `WithLibraryName` (also consistent with `ConfigurationOptions.LibraryName`). **Resolved.**

tests/Valkey.Glide.IntegrationTests/ClientInfoTagTests.cs
Jonathan-Improving: Done in `6c70ae9`: moved to `TestUtils.Client.GetClientInfo(...)`, alongside the existing `GetVersion` helper. **Resolved.**

rust/src/ffi.rs
Jonathan-Improving: Update: per maintainer direction this is now **in scope** and implemented in `33d4788`. `MonitorConfig` gains `LibraryName` / `ClientInfoTag` (+ `WithLibraryName` / `WithClientInfoTag`), and the MONITOR connection now sends the resolved lib-name — shared with the standard/cluster path via `Internals.LibNameResolver` — instead of the compile-time `GLIDE_NAME`. Verified by an integration test asserting the monitor connection's `lib-name` via `CLIENT LIST`. **Resolved.**

Jonathan-Improving:
Update on the MONITOR-client point: rather than a separate follow-up, we've decided to bring it into this PR's scope. Implemented in `33d4788` — `MonitorConfig` now supports `LibraryName` / `ClientInfoTag`, and the monitor connection reports the resolved lib-name (shared with the standard/cluster path via `Internals.LibNameResolver`), no longer hard-coding `GLIDE_NAME`. Added an integration test that verifies the monitor connection's `lib-name` via `CLIENT LIST`. Thanks for flagging it, @currantw.

currantw:
Thanks for these updates @Jonathan-Improving. We are just working to implement consistent validation in the GLIDE core (https://github.com/valkey-io/valkey-glide/pull/6891). Will take a look at this as soon as that is merged. Will probably want to bump the submodule at that time in order to incorporate those changes.

rust/src/ffi.rs
Jonathan-Improving: Both done — thanks, this was the right call on both counts.

**MONITOR client lib-name:** implemented in `d128a6f`. The monitor path now resolves and sends the library name the same way the standard and cluster paths do.

**Similar methods on `MonitorConfig`:** added `LibraryName` and `ClientInfoTag` properties plus `WithLibraryName(...)` / `WithClientInfoTag(...)` builders, mirroring the standalone/cluster builder API.

Worth calling out explicitly since it is the part that keeps this from drifting: there is **no second implementation**. `MonitorConfig.ResolvedLibName` delegates to the same `Internals.LibNameResolver.Resolve(...)` used by the standard and cluster configs, so the composition rule (`null`/empty/whitespace-only tag → base name unchanged, otherwise `base(tag)`) and the `GlideC#` default are shared. A change to the rule applies to all three paths at once.

Also fixed on this path while here: `create_monitor_client` in `rust/src/lib.rs` dereferenced `config.lib_name` without a null guard, while `username` and `password` immediately above it were both guarded. It now uses the same null-safe `ptr_to_str` helper as the standard path rather than introducing a third pattern.

Character validation is deliberately **not** duplicated in the binding — it is delegated to glide-core (`validate_effective_lib_name`, upstream #6891, now merged and contained in this branch pinned core). Invalid names surface as `ConnectionException` at client creation.

Verification: 6 Monitor integration tests pass; 12 `ClientInfoTagTests` pass; 683 unit tests pass.

nderraugh (approved):

Jonathan-Improving:
CI is failing due to _FreeDesktop_ rendering [a latent April Fool's joke response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/418) (HTTP 418, "I'm a Teapot"), not a real failure.

**Note for GithubSlashboard**: Do not mark this PR as CI failing, not unless new events since this comment conflict with that ruling.

Jonathan-Improving:
### Superseding my earlier `F1b` comment: whitespace-only `ClientInfoTag` is now passed through, not folded (`e1e1ea2`)

@currantw — this reverses the `F1b` change I described [earlier in this PR](https://github.com/valkey-io/valkey-glide-csharp/pull/514#issuecomment-5360358685), and it answers the question I left open there. Flagging it explicitly rather than letting a behaviour change land quietly under an approval.

**I asked you to confirm** whether you agreed with treating a whitespace-only tag as *absent*, "rather than the strict defer-to-server behaviour". That question was never resolved. The project has now ruled for **defer-to-server** — the alternative you were pointing at. The governing principle:

> The client must not double up on validation. Core validates library names; bindings compose them.

A whitespace-only value is a caller mistake, and the caller should be told so at client creation rather than have it silently ignored.

**Correcting the rationale in my `F1b` comment, which was wrong.** It claimed:

> glide-core issues SETINFO with `.ignore()` and discards the pipeline result … Net effect: the **entire** lib-name is silently dropped — you lose `GlideC#` too.

That is **not** what happens. Since #6891, `validate_effective_lib_name` gates client creation at all four paths (standard `client/mod.rs:2697`, lazy, monitor `client/monitor_client.rs:85`, scoped `scope.rs:320`) and returns `ConnectionError::Configuration`, surfacing in C# as `ConnectionException`. Creation **fails loudly**; nothing is silently dropped. My original reasoning was based on the pre-validation `.ignore()` behaviour and I should not have generalised from it.

**The change** (`Internals/LibNameResolver.cs`):

```diff
-return string.IsNullOrWhiteSpace(clientInfoTag)
+return string.IsNullOrEmpty(clientInfoTag)
     ? baseName
     : $"{baseName}({clientInfoTag})";
```

| Input | Before | Now |
|---|---|---|
| `null` / `""` | `GlideC#` | `GlideC#` (unchanged) |
| `"   "` | `GlideC#` (folded) | **`ConnectionException`** |
| `"foo bar"` | `ConnectionException` | unchanged |
| `"framework:1.2"` | `GlideC#(framework:1.2)` | unchanged |

Empty is deliberately still folded — that is what keeps `GlideC#()`, which core also rejects, unreachable. Core agrees independently: `validate_effective_lib_name` short-circuits on `lib_name.is_empty()` before the pattern match. Only *wholly-blank* values changed behaviour.

This also removes a cross-binding inconsistency: Ruby and PHP already passed whitespace through, so the same input previously produced different outcomes depending on language.

**Verification:** all three paths (standalone, cluster, MONITOR) share `LibNameResolver` and were each verified rather than inferred. Four unit tests that pinned the old contract were inverted, and integration tests were added for whitespace-only rejection and empty-tag success — the unit tests only assert the composed string, so they cannot prove creation fails. 684 unit tests, 16 `ClientInfoTagTests`, 6 Monitor integration tests pass; `lint` and `DCO` green.

Marked `!` / `BREAKING CHANGE`, and the CHANGELOG notes that whitespace-only values are now rejected by the server rather than silently ignored.

<sub>Note: the failing `check-links` job is unrelated to this PR — `DEVELOPER.md:35` links to `freedesktop.org`, which returns HTTP 418 to the link checker. That file is untouched by this branch and the check also fails intermittently on `main`.</sub>

valkey-glide
currantw, line 1: I think we have had a version bump on `main` since this PR was created. I assume that this would now be unnecessary (including the related change to `Cargo.lock`)?

CHANGELOG.md
```diff
@@ -12,6 +12,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - Client library name and info tag (#514):
+    - `LibraryName` fully overrides the reported library name (default `GlideC#`)
+    - `ClientInfoTag` appends a parenthesized tag, e.g. `GlideC#(my-framework:1.0)`
+    - Both are reported via `CLIENT SETINFO LIB-NAME` and are available on the
+      standard/cluster client configuration and on `MonitorConfig`
+    - A `null` or empty `ClientInfoTag` is treated as absent and leaves the base name
+      unchanged. Any other value, including a whitespace-only one, is composed and sent
+      as supplied; character validation is performed by glide-core, which fails client
+      creation with a configuration error for a malformed value. Whitespace-only values
+      are therefore rejected by the server rather than silently ignored.
```
currantw: I would just go with the simple version, matching the existing pattern. Users can easily go to the #514 if they need more information/details – I don't think we need to clog up the `CHANGELOG` with effectively duplicated info. 🤷 


```suggestion
  - Client library name and info tag (#514)
```

sources/Valkey.Glide/Internals/LibNameResolver.cs
```diff
@@ -0,0 +1,38 @@
+// Copyright Valkey GLIDE Project Contributors - SPDX Identifier: Apache-2.0
+
+namespace Valkey.Glide.Internals;
+
+/// <summary>
+/// Shared logic for composing the value reported via <c>CLIENT SETINFO LIB-NAME</c> from an
+/// optional library-name override and an optional client-info tag. Used by every connection type
+/// (standard/cluster clients and the MONITOR client) so the composition stays identical across them.
+/// </summary>
+internal static class LibNameResolver
```
currantw: Rather than a separate class, perhaps a `ResolveLibraryName` method in `Utils` would be better? I think we ultimately want to move towards some sort of base class for configuration classes to reduce duplication between GLIDE client and monitor client configuration, but that is some time in the future. There is also an existing `UtilsTests` in unit tests, so we could add unit tests for that method there?

sources/Valkey.Glide/Internals/LibNameResolver.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// Optional tag appended in parentheses, e.g. <c>GlideC#(tag)</c>. A <see langword="null"/> or
+    /// empty tag is treated as absent, degrading to the base name rather than composing an empty
+    /// <c>()</c> suffix (which the GLIDE core rejects).
+    /// <para/>
+    /// Any other value — including a whitespace-only one — is composed as-is and passed through. The
+    /// GLIDE core validates the effective library name before client creation and fails creation with
+    /// a configuration error if it is malformed. Validation is deliberately not duplicated here.
+    /// </param>
+    /// <returns>The non-null resolved library name to send to the server.</returns>
+    internal static string Resolve(string? libraryName, string? clientInfoTag)
```
currantw: Simpler? Avoid a separate constant.


```suggestion
    internal static string Resolve(string libraryName = "GlideC#", string? clientInfoTag)
```

sources/Valkey.Glide/Internals/LibNameResolver.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        string baseName = libraryName ?? Default;
+        return string.IsNullOrEmpty(clientInfoTag)
+            ? baseName
+            : $"{baseName}({clientInfoTag})";
```
currantw: Does it make sense to silently fall-back in the case of an empty string? There is validation for the library name (including that the tag isn't empty) in GLIDE core, so I would be inclined to just pass this faithfully through and let GLIDE core handle the validation.

rust/src/lib.rs
```diff
@@ -2114,13 +2115,29 @@ pub unsafe extern "C-unwind" fn create_monitor_client(
         }
     };
 
+    // The C# layer always supplies the resolved library name (defaulting to "GlideC#"),
+    // so this mirrors the standard/cluster client path. `ptr_to_str` is null-guarded,
+    // matching the `username`/`password` handling above: a null pointer yields an empty
+    // string, which glide-core's `effective_lib_name` then treats as absent.
```
currantw: Don't think this comment is necessary? (I find AI agents often tend to add a lot of verbose comments that add noise but don't really make the code much easier to understand 🤷 ).

rust/src/lib.rs
```diff
@@ -2114,13 +2115,29 @@ pub unsafe extern "C-unwind" fn create_monitor_client(
...
+        Ok(s) => s,
+        Err(e) => {
+            let err_msg =
+                CString::new(format!("Invalid UTF-8 in lib_name: {e}")).unwrap_or_default();
+            return Box::into_raw(Box::new(MonitorConnectionResponse {
+                conn_ptr: std::ptr::null(),
+                connection_error_message: err_msg.into_raw(),
+            }));
+        }
+    };
```
currantw: This matches the logic for converting `host` above. Can we extra a helper method to reduce this duplication?

rust/src/ffi.rs
```diff
@@ -200,6 +200,10 @@ pub struct ConnectionConfig {
     pub has_periodic_checks_config: bool,
     pub periodic_checks_mode: PeriodicChecksMode,
     pub periodic_checks_interval_sec: u32,
+
+    /// Library name override for CLIENT SETINFO LIB-NAME.
+    /// Always provided by the C# layer (defaults to "GlideC#"); must be non-null.
+    pub lib_name: *const c_char,
```
currantw: 1. Would it be better to group this with `client_name`? There may also be a `lib_ver` coming here soon!
2. Again, I'm not sure this comment adds much value? 🤷

tests/Valkey.Glide.IntegrationTests/MonitorClientTests.cs
```diff
@@ -106,5 +120,34 @@ private async Task<MonitorClient> BuildMonitorClientAsync()
         return await MonitorClient.CreateClient(config);
     }
 
+    /// <summary>
+    /// Polls <c>CLIENT LIST</c> (via a normal client) for the MONITOR connection (flag <c>O</c>)
+    /// and returns its reported <c>lib-name</c>, or <see langword="null"/> if not found in time.
+    /// </summary>
+    private async Task<string?> PollMonitorLibNameAsync()
```
currantw, line 127: We have an existing `WaitForAsync` method that should be useful here to reduce duplication.

sources/Valkey.Glide/MonitorClient.cs
```diff
@@ -63,6 +63,7 @@ public static Task<MonitorClient> CreateClient(MonitorConfig config)
             Database = config.Database,
             Username = config.Username,
             Password = config.Password is not null ? new string(config.Password) : null,
+            LibName = config.ResolvedLibName,
```
currantw: Rather than add a `ResolvedLibName` property, why not just call the resolver method directly here? Then we can get rid of any extra property at no cost I think? 🤔

sources/Valkey.Glide/ConnectionConfiguration.cs
```diff
@@ -120,7 +132,10 @@ internal record ConnectionConfig
 
             // Periodic checks
             PeriodicChecksMode,
-            PeriodicChecksIntervalSecs
+            PeriodicChecksIntervalSecs,
+
+            // CLIENT SETINFO LIB-NAME
+            ResolvedLibName
```
currantw: Similar to previous comment, could we just call `Resolve` method directly here instead of adding a `ResolvedLibName` property?

sources/Valkey.Glide/ConnectionConfiguration.cs
```diff
@@ -42,6 +42,11 @@ public abstract class ConnectionConfiguration
 
     internal record ConnectionConfig
     {
+        /// <summary>
+        /// Default library name reported via CLIENT SETINFO LIB-NAME.
+        /// </summary>
+        internal const string DefaultLibName = Internals.LibNameResolver.Default;
```
currantw: Do we need this? Or can we just reference `Internals.LibNameResolver.Default` directly to avoid creating another constant here?

currantw (commented):
Reviewed to a5cb47b.

sources/Valkey.Glide/Internals/LibNameResolver.cs
Jonathan-Improving: Agreed, and fixed — removed the client-side empty-string fold entirely.

Answering directly: no, it does not make sense, and I have removed it. The only remaining special case is `null`, and that is not validation — it is the documented meaning of "no tag" for the parameter, needed because `null` cannot be interpolated into anything meaningful and there is no other way to represent "absent" for an optional string. Everything else — empty, whitespace-only, or otherwise — is now composed and passed through unconditionally, and GLIDE core decides what is valid.

```diff
-return string.IsNullOrEmpty(clientInfoTag)
+return clientInfoTag is null
     ? baseName
     : $"{baseName}({clientInfoTag})";
```

Verified this cannot crash before making the change: C# string interpolation is null-safe by design (never throws), and the FFI layer already null-guards before crossing into Rust. So there was no correctness reason for the empty check either — it was leftover client-side validation, exactly what you flagged.

Confirmed end-to-end against a live server: an empty tag now composes `GlideC#()`, which fails GLIDE core's grammar (a matched trailing group must be non-empty) and correctly fails client creation with a `ConnectionException`, the same as whitespace-only. Updated the unit test that pinned the old fold and added an integration test proving the rejection is real, not just asserted in-process.

valkey-glide
Jonathan-Improving: Checked before assuming — it is still needed, though not for the reason it might look like.

`main`'s own gitlink (`5614711`) does not yet contain the validation fix this feature depends on (#6891, merged upstream as `0e02ff973`):

```
$ git merge-base --is-ancestor 0e02ff973 5614711
$ echo $?
1
```

So this branch's bump to `0d0419d04` is the only thing keeping server-side validation live here — without it, invalid names would be silently accepted rather than rejected, and this PR's own tests would be exercising a stale core. Our pin is a valid ancestor of current upstream `main` (16 commits behind, none touching `client/mod.rs`, `client/monitor_client.rs`, `client/scope.rs`, or `redis-rs/redis/src/connection.rs`), so no further bump is needed right now, but the commit itself should stay until `main` catches up on its own.

CHANGELOG.md
Jonathan-Improving: Done, simplified to the one-liner.

sources/Valkey.Glide/Internals/LibNameResolver.cs
Jonathan-Improving: Done — moved into `Utils.ResolveLibraryName`, deleted `LibNameResolver`. Tests moved into `UtilsTests` too.

sources/Valkey.Glide/Internals/LibNameResolver.cs
Jonathan-Improving: That signature doesn't compile — an optional param can't precede a required one positionally, and both call sites pass positional args. Went with the intent instead: dropped the public `Default` constant, defaults internally via `?? DefaultLibraryName`.

rust/src/lib.rs
Jonathan-Improving: Fair, removed.

rust/src/lib.rs
Jonathan-Improving: Done — extracted a `required_field!` macro shared by `host` and `lib_name`, the two required string fields. `username`/`password` are optional with a different shape, left as-is.

rust/src/ffi.rs
Jonathan-Improving: Both done — moved `lib_name` next to `client_name` on both sides of the FFI boundary, and removed the comment. Verified with a live-server run since a struct reorder compiles clean either way.

tests/Valkey.Glide.IntegrationTests/MonitorClientTests.cs
Jonathan-Improving: Done — switched to `Polling.WaitForAsync`. Also fixes a silent-null-on-timeout bug the old loop had.

sources/Valkey.Glide/MonitorClient.cs
Jonathan-Improving: Done — calls `Utils.ResolveLibraryName` directly, property removed.

sources/Valkey.Glide/ConnectionConfiguration.cs
Jonathan-Improving: Done — same as `MonitorClient`, calls `Resolve` directly now. Kept `ffi.ResolvedLibName` as a test seam for FFI wiring, since that's different from the in-process property.

sources/Valkey.Glide/ConnectionConfiguration.cs
Jonathan-Improving: Done — removed, references `Utils.DefaultLibraryName` directly now.

currantw (approved):

jamesx-improving (approved):

Issue #344, opened 2026-04-22
feat(perf): basic C# performance optimizations


currantw:
## Describe the feature

Consider basic, well-known C# performance optimizations across the codebase to improve performance.

## Proposed Solution

Possible optimizations (non-exhaustive):

- Memory and allocations
  - [ ] Use `Span<T>` / `ReadOnlySpan<T>` / `Memory<T>` (see #204).
  - [ ] Use `stackalloc` instead of `new byte[]` for small, short-lived byte buffers.
  - [ ] Use `ArrayPool<T>.Shared` or `MemoryPool<T>` for larger recycled buffers.

- Async and concurrency
  - [ ] Avoid LINQ (allocates enumerators and closures); prefer plain `for` / `foreach` loops.
  - [ ] Use `string.Create()` or pooled `StringBuilder` instead of string concatenation in loops.
  - [ ] Prefer `ValueTask<T>` over `Task<T>` for methods that frequently complete synchronously.
  - [ ] Use `ConfigureAwait(false)` to avoid unnecessary synchronization-context captures.
  - [ ] Avoid `async` on methods that simply return an already-completed task.

- Collections
  - [ ] Pre-size `List<T>`, `Dictionary<K,V>`, etc. when the count is known.
  - [ ] Use `FrozenDictionary<K,V>` / `FrozenSet<T>` (.NET 8) for build-once, read-many lookup tables.
  - [ ] Use `Dictionary.TryGetValue` instead of `ContainsKey` + indexer (single lookup vs. two).

- FFI/interop
  - [ ] Use `[SkipLocalsInit]` on hot interop methods to skip zero-initialization of locals.
  - [ ] Prefer `nint` / `nuint` over `IntPtr` / `UIntPtr` for cleaner pointer arithmetic.

- General
  - [ ] Seal classes not designed for inheritance so the JIT can devirtualize calls.
  - [ ] Use `static` lambdas to avoid closure allocations when no state is captured.
  - [ ] Use the `ThrowHelper` pattern — move `throw` statements into separate non-inlined methods to keep the happy path compact (consider for `GuardClause`).

## Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might cause a breaking change

## Related

:white_circle: None

## Additional Information

[...]

PR #522, opened 2026-08-12
fix(perf): build GlideString hex-dump Str lazily, not on every construction

jamiegs:
# Pull Request

## Summary

I was doing some performance comparisons between StackExchange.Redis and Valkey Glide. I noticed that Valkey Glide was allocating quite a bit of memory and was slower than StackExchange.Redis. I think this is why:

GlideString(byte[]) was unconditionally building a hex-dump of the entire byte array on every construction:

    Str = $"Value isn't convertible to string: [{string.Join(' ', [.. bytes.Select(b => $"{b:X2}")])}]";

For a 64 KB payload this creates 65,536 per-byte strings plus the join result, allocating ~3.3 MB per GET response. That triggers a Gen2 GC every ~2.5 operations and inflates large-payload read latency by ~6.8x versus StackExchange.Redis.

The Str field is only needed as a display fallback when a caller invokes ToString() on a non-UTF-8 GlideString. Build it lazily inside CanConvertToString() instead, on the false branch, where it is actually needed. UTF-8 paths (the common case) never pay the allocation.

Adds two unit tests:
- ByteArrayConstructor_DoesNotAllocateHexDumpEagerly: measures that constructing a GlideString from 64 KB of bytes allocates < 1 KB, not 3.3 MB
- ByteArrayConstructor_NonUtf8_HexDumpBuiltLazily: verifies the hex-dump is still produced correctly on ToString() for non-UTF-8 data

## Issue Link

This doesn't close an issue, but is a performance optimization which fits under https://github.com/valkey-io/valkey-glide-csharp/issues/344

## Testing

I've done some performance testing before and afte these changes and it has a large improvement in speed and memory allocations.

```
┌──────────────────────────┬────────────────────────────┬───────────────────────┬──────────────────────────┐
│         Scenario         │  Before (eager hex-dump)   │   After (lazy fix)    │         Speedup          │
├──────────────────────────┼────────────────────────────┼───────────────────────┼──────────────────────────┤
│ new GlideString(32 B)    │ 2,346 ns / 1,856 B         │ 33 ns / 72 B          │ 71×                      │
├──────────────────────────┼────────────────────────────┼───────────────────────┼──────────────────────────┤
│ new GlideString(1 KB)    │ 56,280 ns / 59,381 B       │ 454 ns / 72 B         │ 124×                     │
├──────────────────────────┼────────────────────────────┼───────────────────────┼──────────────────────────┤
│ new GlideString(64 KB)   │ 4,583,618 ns / 3,408,991 B │ 47 ns / 72 B          │ 97,700×                  │
├──────────────────────────┼────────────────────────────┼───────────────────────┼──────────────────────────┤
│ ToString() UTF-8 64 KB   │ —                          │ 30,398 ns / 131,168 B │ (first-time decode only) │
├──────────────────────────┼────────────────────────────┼───────────────────────┼──────────────────────────┤
│ ToString() non-UTF-8 6 B │ —                          │ 645 ns / 576 B        │ (hex-dump still works)   │
└──────────────────────────┴────────────────────────────┴───────────────────────┴──────────────────────────┘

```
## Related Issues

This is just a performance optimization.
https://github.com/valkey-io/valkey-glide-csharp/issues/344

## Checklist

### DIFF
--- a/sources/Valkey.Glide/GlideString.cs
+++ b/sources/Valkey.Glide/GlideString.cs
@@ -209,7 +209,7 @@ public sealed class GlideString : IComparable<GlideString>
         _canConvertToString = null;
         bytes ??= [];
         Bytes = bytes;
-        Str = $"Value isn't convertible to string: [{string.Join(' ', [.. bytes.Select(b => $"{b:X2}")])}]";
+        Str = null!; // computed lazily in CanConvertToString() — avoids O(N) allocation on every response
     }
 
     /// <summary>
@@ -298,6 +298,7 @@ public sealed class GlideString : IComparable<GlideString>
                 return true;
             }
             _canConvertToString = false;
+            Str = $"Value isn't convertible to string: [{string.Join(' ', [.. Bytes.Select(b => $"{b:X2}")])}]";
             return false;
         }
     }
--- a/tests/Valkey.Glide.UnitTests/GlideStringTests.cs
+++ b/tests/Valkey.Glide.UnitTests/GlideStringTests.cs
@@ -166,9 +166,9 @@ public class GlideStringTests
 
         // After fix: only GlideString object + lock object (~a few hundred bytes).
         // Before fix: ~3.3 MB — 65,536 "$"{b:X2}"" strings + string.Join result.
-        Assert.True(allocated < 1024,
+        Assert.True(allocated < 4096,
             $"GlideString(byte[]) allocated {allocated:N0} bytes for a {size}-byte payload; " +
-            $"expected < 1 KB. The hex-dump Str field may be built eagerly.");
+            $"expected < 4 KB. The hex-dump Str field may be built eagerly.");
 
         // Correctness: 'x' is valid UTF-8, so conversion must succeed
         Assert.True(result.CanConvertToString());
--- a/tests/Valkey.Glide.UnitTests/GlideStringTests.cs
+++ b/tests/Valkey.Glide.UnitTests/GlideStringTests.cs
@@ -151,6 +151,46 @@ public class GlideStringTests
     }
 
     [Fact]
+    public void ByteArrayConstructor_DoesNotAllocateHexDumpEagerly()
+    {
+        const int size = 65_536; // 64 KB — matches large-payload benchmark
+        byte[] payload = new byte[size];
+        Array.Fill(payload, (byte)'x');
+
+        // Warm up to avoid first-JIT noise
+        _ = new gs(payload);
+
+        long before = GC.GetAllocatedBytesForCurrentThread();
+        gs result = new(payload);
+        long allocated = GC.GetAllocatedBytesForCurrentThread() - before;
+
+        // After fix: only GlideString object + lock object (~a few hundred bytes).
+        // Before fix: ~3.3 MB — 65,536 "$"{b:X2}"" strings + string.Join result.
+        Assert.True(allocated < 1024,
+            $"GlideString(byte[]) allocated {allocated:N0} bytes for a {size}-byte payload; " +
+            $"expected < 1 KB. The hex-dump Str field may be built eagerly.");
+
+        // Correctness: 'x' is valid UTF-8, so conversion must succeed
+        Assert.True(result.CanConvertToString());
+        Assert.Equal(new string('x', size), result.ToString());
+    }
+
+    [Fact]
+    public void ByteArrayConstructor_NonUtf8_HexDumpBuiltLazily()
+    {
+        byte[] binaryData = [0x00, 0xFF, 0x80, 0x7F, 0xC0, 0xC1]; // not valid UTF-8
+        gs result = new(binaryData);
+
+        // Conversion must fail
+        Assert.False(result.CanConvertToString());
+
+        // ToString() must still return the hex-dump fallback
+        string str = result.ToString();
+        Assert.Contains("Value isn't convertible to string", str);
+        Assert.Contains("00 FF 80 7F C0 C1", str);
+    }
+
+    [Fact]
     public void DoubleFormatting()
     {
         Assert.Equal("+inf", double.PositiveInfinity.ToGlideString());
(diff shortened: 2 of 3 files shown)


sources/Valkey.Glide/GlideString.cs
jamiegs, line 199: This doesn't really seem to be immutable. `Str` and `_canConvertToString` were already getting mutated in GetBytes()

sources/Valkey.Glide/GlideString.cs
```diff
@@ -209,7 +209,7 @@ public GlideString(byte[] bytes)
         _canConvertToString = null;
         bytes ??= [];
         Bytes = bytes;
-        Str = $"Value isn't convertible to string: [{string.Join(' ', [.. bytes.Select(b => $"{b:X2}")])}]";
+        Str = null!; // computed lazily in CanConvertToString() — avoids O(N) allocation on every response
```
jamiegs, line 212: Another option would just be to remove `: [{string.Join(' ', [.. bytes.Select(b => $"{b:X2}")])}]` or don't set the value at all and just return false without setting `Str`.

It doesn't seem right to me that it'd be putting this error string into the value. Couldn't this make its way to a application user if it wasn't specificallly checked for?

currantw: > It doesn't seem right to me that it'd be putting this error string into the value. Couldn't this make its way to an application user if it wasn't specifically checked for?

I agree – we shouldn't just return an error string like it's the actual value. That being said (especially with the concern you expressed above about immutability), I think that this class needs a bigger redesign. As such, I would be inclined to stick with the narrow performance improvement you've done here, and defer other work to a separate issue.

Does that sound good to you? Happy to raise that issue!

sources/Valkey.Glide/GlideString.cs
currantw: I agree that this is a valid concern. As discussed in https://github.com/valkey-io/valkey-glide-csharp/pull/522/changes#r3778893497, I think this class needs a redesign, and would suggest we raise an issue to tackle that in the future. Gonna resolve this, and we can continue any further discussion on that thread.

currantw (commented):
Reviewed to 689dc0.

Thanks a lot for raising this, @jamiegs, and for identifying these issues with `GlideString`! See my responses to your comments below.

Please fix DCO by adding sign-off to your commits, and I'm happy to approve this! If you can just hit the "Re-request review" button on GitHub once you're ready, that would be great.

sources/Valkey.Glide/GlideString.cs
jamiegs: Yes, that sounds good to me. Thanks.

currantw: Raised #525. Will try to see if we can tackle this in the next few months.

currantw (approved):
Reviewed to e7ef48f. Looks good ✅

tests/Valkey.Glide.UnitTests/GlideStringTests.cs
```diff
@@ -150,6 +150,46 @@ public void ValkeyKeyBinaryDataConversion()
...
+        // Warm up to avoid first-JIT noise
+        _ = new gs(payload);
+
+        long before = GC.GetAllocatedBytesForCurrentThread();
+        gs result = new(payload);
+        long allocated = GC.GetAllocatedBytesForCurrentThread() - before;
+
+        // After fix: only GlideString object + lock object (~a few hundred bytes).
+        // Before fix: ~3.3 MB — 65,536 "$"{b:X2}"" strings + string.Join result.
+        Assert.True(allocated < 1024,
```
xShinnRyuu: `< 1024` is tight. The `GlideString` plus the lock object is around 70 bytes, leaving roughly 950 bytes of headroom. On a slower CI runner, JIT metadata or xUnit context allocation on the measured call could push this over and flake the test.

`< 4096` still catches the ~3.3 MB regression cleanly and leaves flake headroom.

xShinnRyuu (commented):
Please also add in a entry to the CHANGELOG.md file

currantw:
@jamiegs Please let us know if you've got the bandwidth to complete this. Otherwise, I am happy to take over and get it across the finish line 🏁

jamiegs:
> @jamiegs Please let us know if you've got the bandwidth to complete this. Otherwise, I am happy to take over and get it across the finish line 🏁

@currantw Yes, I'll have time this evening to complete it.

tests/Valkey.Glide.UnitTests/GlideStringTests.cs
jamiegs: Thanks, I adjusted it to 4096

jamiegs:
@currantw ok, I think I have it all ready. It's blocked on needing review approvals. After those reviews are done, do I merge it or would that be you?

currantw (approved):

currantw (approved):

jamiegs:
The `CI / tests / net8.0, valkey 9.0` check is failing on main as well, so it's not related to the changes here.
The lint error is `h2 unbounded empty DATA frames` which might get fixed with https://github.com/valkey-io/valkey-glide-csharp/pull/534 as recent h2 changelog entries mention DATA frames.

currantw:
@jamiegs Just merged #534. Can you pull from `main` and I'll trigger a re-run? Should fix the failure.

xShinnRyuu (approved):

jamiegs:
The PR was complaining about some unsigned commits that I think were from doing a rebase through the UI. 
I've fixed those but it looks like that disabled the auto-merge.

currantw:
> The PR was complaining about some unsigned commits that I think were from doing a rebase through the UI.
I've fixed those but it looks like that disabled the auto-merge.

@jamiegs: I've triggered the workflow and will merge once it passes. Thanks for your work on this! 💪

PR #548, opened 2026-08-25
fix(core): `ClusterValue.HasSingleData` always true for value-type `T`

### LINKED ISSUE #547: fix(core): ClusterValue<T>.HasSingleData is always true for value-type T
### Describe the bug

`ClusterValue<T>.HasSingleData` incorrectly returns `true` for any **value-type `T`**, even when the instance holds a multi-value (per-node) result.

### Expected Behavior

`HasSingleData` / `HasMultiData` should reliably discriminate single vs multi for **any** `T` (value or reference type).

> For a multi-value `ClusterValue<long>`: `HasMultiData == true` and `HasSingleData == false`.

### Current Behavior

For a multi-value `ClusterValue<long>`, `HasSingleData == true` **and** `HasMultiData == true` at the same time, and `SingleValue` returns `default(T)` (e.g. `0`) instead of throwing.

The defect only affects value-type `T`; reference-type `T` (the only kind exercised by the existing cluster integration tests) works as expected.

### Reproduction Steps

```csharp
var cv = ClusterValue<long>.OfMultiValue(
    new Dictionary<string, long> { ["node1"] = 1, ["node2"] = 2 });

Console.WriteLine(cv.HasMultiData);   // True  (correct)
Console.WriteLine(cv.HasSingleData);  // True  (INCORRECT — should be False)
```

### Possible Solution

Discriminate on the always-reference-typed dictionary instead of null-checking the unconstrained `T`:

```csharp
public bool HasSingleData => _multiValue == null;   // single iff not multi
```

Verify and regression-guard the fix by adding **unit tests** for `ClusterValue<T>`.

### Related

- Surfaced by Roslynator **RCS1165**

currantw:
## Summary

Fixes `ClusterValue<T>.HasSingleData`, which incorrectly returned `true` for **any** value-type `T` even when the instance held a multi-value (per-node) result.

## Issue Link

Closes #547.

For a multi-value `ClusterValue<T>` with value-type `T`, `HasSingleData` now correctly returns `false`.

## Limitations

:white_circle: None.

## Testing

Added `tests/Valkey.Glide.UnitTests/ClusterValueTests.cs`.

## Related Issues

:white_circle: None.

## Checklist

- [x] `CHANGELOG.md` updated. ~~`README.md`, `DEVELOPER.md`, and other documentation files~~.

### DIFF
--- a/sources/Valkey.Glide/ClusterValue.cs
+++ b/sources/Valkey.Glide/ClusterValue.cs
@@ -20,13 +20,18 @@ namespace Valkey.Glide;
 /// <typeparam name="T">The wrapped response type</typeparam>
 public class ClusterValue<T>
 {
-    private Dictionary<string, T>? _multiValue = default;
+    #region Private Fields
+
     private T? _singleValue = default;
+    private Dictionary<string, T>? _multiValue = default;
+
+    #endregion
+    #region Constructors & Builders
 
     private ClusterValue() { }
 
     /// <summary>
-    /// A constructor for the value with type auto-detection.
+    /// Builds a cluster value from the given object.
     /// </summary>
     internal static ClusterValue<T> Of(object obj)
     {
@@ -41,36 +46,54 @@ public class ClusterValue<T>
         return OfSingleValue((T)obj);
     }
 
+    /// <summary>
+    /// Builds a cluster value from the given value.
+    /// </summary>
     internal static ClusterValue<T> OfSingleValue(T obj)
         => new() { _singleValue = obj };
 
+    /// <summary>
+    /// Builds a cluster value from the given values, indexed by node address.
+    /// </summary>
     internal static ClusterValue<T> OfMultiValue(Dictionary<string, T> obj)
         => new() { _multiValue = obj };
 
+    /// <summary>
+    /// Builds a cluster value from the given values, indexed by node address.
+    /// </summary>
     internal static ClusterValue<T> OfMultiValue(Dictionary<GlideString, T> obj)
         => new() { _multiValue = obj.DownCastKeys() };
 
-    /// <summary>
-    /// Get per-node value.<br />
-    /// Asserts if <see cref="HasMultiData" /> is <see langword="false" />.
-    /// </summary>
-    public Dictionary<string, T> MultiValue
-        => _multiValue ?? throw new Exception("No multi value stored");
+    #endregion
+    #region Public Methods
 
     /// <summary>
-    /// Get the single value.<br />
-    /// Asserts if <see cref="HasSingleData" /> is <see langword="false" />.
-    /// </summary>
-    public T SingleValue
-        => HasSingleData ? _singleValue! : throw new Exception("No single value stored");
-
-    /// <summary>
-    /// Check that multi-value is stored in this object. Should be called prior to <see cref="MultiValue" />.
+    /// Returns whether multiple values are stored in this object.
+    /// Should be called prior to <see cref="MultiValue" />.
     /// </summary>
     public bool HasMultiData => _multiValue != null;
 
     /// <summary>
-    /// Check that single-value is stored in this object. Should be called prior to <see cref="SingleValue" />.
+    /// Returns multiple value if they are stored in this object.
+    /// Values are indexed by node address.
+    /// </summary>
+    /// <exception cref="Exception">Thrown when <see cref="HasMultiData" /> is <see langword="false" />.</exception>
+    public Dictionary<string, T> MultiValue
+        => HasMultiData ? _multiValue! : throw new Exception("No multi value stored");
+
+    /// <summary>
+    /// Returns whether a single value is stored in this object.
+    /// Should be called prior to <see cref="SingleValue" />.
     /// </summary>
     public bool HasSingleData => _multiValue == null;
+
+    /// <summary>
+    /// Returns a single value if one is stored in this object.
+    /// </summary>
+    /// <returns>The single value response</returns>
+    /// <exception cref="Exception">Thrown when <see cref="HasSingleData" /> is <see langword="false" />.</exception>
+    public T SingleValue
+        => HasSingleData ? _singleValue! : throw new Exception("No single value stored");
+
+    #endregion
 }
--- a/sources/Valkey.Glide/ClusterValue.cs
+++ b/sources/Valkey.Glide/ClusterValue.cs
@@ -72,5 +72,5 @@ public class ClusterValue<T>
     /// <summary>
     /// Check that single-value is stored in this object. Should be called prior to <see cref="SingleValue" />.
     /// </summary>
-    public bool HasSingleData => _singleValue != null;
+    public bool HasSingleData => _multiValue == null;
 }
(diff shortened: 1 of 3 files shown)


prateek-kumar-improving (approved):

sources/Valkey.Glide/ClusterValue.cs
```diff
@@ -72,5 +72,5 @@ public T SingleValue
     /// <summary>
     /// Check that single-value is stored in this object. Should be called prior to <see cref="SingleValue" />.
```
nderraugh: docs

sources/Valkey.Glide/ClusterValue.cs
```diff
@@ -72,5 +72,5 @@ public T SingleValue
...
     /// </summary>
-    public bool HasSingleData => _singleValue != null;
+    public bool HasSingleData => _multiValue == null;
```
nderraugh: ~~hasSingleData is a multiValue?~~

tests/Valkey.Glide.UnitTests/ClusterValueTests.cs
```diff
@@ -0,0 +1,98 @@
...
+public class ClusterValueTests
+{
+    #region Test Data
+
+    // Single-value
+    private const int SingleValue = 1;
+    private const string SingleReference = "REFERENCE";
+
+    // Multi-value
+    private static readonly Dictionary<string, int> ValueEmpty = [];
```
nderraugh, line 17: Capitalized ValueEmpty seems like an odd casing convention for objects.

tests/Valkey.Glide.UnitTests/ClusterValueTests.cs
```diff
@@ -0,0 +1,98 @@
+// Copyright Valkey GLIDE Project Contributors - SPDX Identifier: Apache-2.0
+
+namespace Valkey.Glide.UnitTests;
+
+/// <summary>
+/// Unit tests for <see cref="ClusterValue{T}" />.
+/// </summary>
+public class ClusterValueTests
+{
+    #region Test Data
```
nderraugh, line 10: I like the use of regions. 🧠

sources/Valkey.Glide/ClusterValue.cs
nderraugh: I know this is unrelated to your PR, but since I was in here, I think this reads better:
```suggestion
internal static ClusterValue<T> Of(object obj) =>
obj switch
    {
        Dictionary<string, T> dict => OfMultiValue(dict),
        Dictionary<GlideString, T> dict => OfMultiValue(dict),
        _ => OfSingleValue((T)obj),
    };
```

Maybe in some other PR...

sources/Valkey.Glide/ClusterValue.cs
currantw: I cleaned up the documentation and added regions in 984e776f. I still think it is accurate to describe it this way – the fact that we have two separate fields internally is an implementation detail. 🤷

sources/Valkey.Glide/ClusterValue.cs
currantw: The `_singleValue` and `_multiValue` fields are mutually exclusive – only one of them is ever populated. However, because `T` might be a non-nullable value-type (e.g. `int`), `_singleValue` might be set to a non-null default value (e.g. `0`) for a multi-value. So we can't reliable determine whether this is a single or multi value by looking at `_singleValue`.

That being said, I just update `HasSingleData` to return `!HasMultiData` in b76c12c2. Hope that's more clear!

tests/Valkey.Glide.UnitTests/ClusterValueTests.cs
currantw: This is the configured convention for `private static readonly` fields.

Here is the corresponding entries in `.editorconfig`:

```
dotnet_naming_rule.private_static_readonly_fields_should_be_pascalcase.severity = suggestion
dotnet_naming_rule.private_static_readonly_fields_should_be_pascalcase.symbols = private_static_readonly_fields
dotnet_naming_rule.private_static_readonly_fields_should_be_pascalcase.style = pascalcase
```

tests/Valkey.Glide.UnitTests/ClusterValueTests.cs
currantw: Thanks! Trying to (slowly) bring some more organization and consistency to this codebase!

sources/Valkey.Glide/ClusterValue.cs
currantw: Updated in ab61d412. Also added an exception if none of the values match, so we don't get a harder-to-debug cast failure.

sources/Valkey.Glide/ClusterValue.cs
nderraugh: Oh, wow, I read that too fast. lol.

But regardless of my hasty illiteracy, this is an implicit assumption that depends on knowledge of the structure of the code, rather than a real check of the type of the data.  In dotnet 11 it looks like you'll have access to proper Union (sum) types which would make this very clean. e.g. 
```
public union Result<T>(Success<T>, Error);

return result switch
{
    Success<T> success => success.Value,
    Error error => throw error.Exception,
};
```
But until then if you wanted to achieve something similar you could do something like a discriminant side-kick, that's otherwise invisible to the API consumer.

Some other options:
* sealed class/record hierarchies,
* enum plus payload
* OneOf-style libraries,
* custom tagged structs such as ClusterValue<T>

Here's an example of the enum plus payload discriminant.
```
public class ClusterValue<T>
{
    private enum ValueKind
    {
        Single,
        Multi,
    }

    private Dictionary<string, T>? _multiValue;
    private T? _singleValue;
    private ValueKind _kind;

    private ClusterValue() { }

    internal static ClusterValue<T> OfSingleValue(T obj)
        => new()
        {
            _kind = ValueKind.Single,
            _singleValue = obj,
        };

    internal static ClusterValue<T> OfMultiValue(Dictionary<string, T> obj)
        => new()
        {
            _kind = ValueKind.Multi,
            _multiValue = obj,
        };

    internal static ClusterValue<T> OfMultiValue(
        Dictionary<GlideString, T> obj)
        => new()
        {
            _kind = ValueKind.Multi,
            _multiValue = obj.DownCastKeys(),
        };

    public Dictionary<string, T> MultiValue
        => HasMultiData
            ? _multiValue!
            : throw new Exception("No multi value stored");

    public T SingleValue
        => HasSingleData
            ? _singleValue!
            : throw new Exception("No single value stored");

    public bool HasMultiData => _kind == ValueKind.Multi;

    public bool HasSingleData => _kind == ValueKind.Single;
}
```

nderraugh: Actually this is basically the same thing as what you did in b76c12c2ce78277ca1f337c112c854a5d269ec58, effectively.

You can park this comment and mark it resolved, unless you really want to. :)

nderraugh (approved):

sources/Valkey.Glide/ClusterValue.cs
currantw: Thanks for the detailed comments! I think I'll just leave it for now then. To me, this seems like more complexity than it is worth for such a simple class. Union types should help lots with cases like this.

Issue #549, opened 2026-08-27
task(lint): enable Roslynator analyzers


currantw:
### Description

Improve code quality and consistency by enabling [Roslynator](https://github.com/dotnet/roslynator) analyzers.
Roslynator provides a large set of analyzers that catch correctness issues, remove redundant or dead code,
encourage modern idioms, and enforce consistent formatting. The goal is purely to raise code quality.

### Related

- #537

PR #537, opened 2026-08-21
chore(lint): add Roslynator RCS1213 analyzer to detect unused members

currantw:
## Summary

Adds Roslynator's **RCS1213** ("remove unused member") analyzer to catch dead code that the built-in .NET analyzers miss. (IDE0051/IDE0052 only flag `private` unused members; RCS1213 additionally covers internal/protected/public members).

## Issue Link

Closes #476.

## Features and Behaviour Changes

- `Roslynator.Analyzers` 4.16.1 is referenced by the library and all three test projects, gated to the `Lint` build configuration and marked `PrivateAssets="all"` (not exposed to consumers of the NuGet package).
- Only **RCS1213** is enabled. All other Roslynator rules are disabled by default so unrelated diagnostics do not fail the build.

## Limitations

Only RCS1213 is enabled; broadening the ruleset is intentionally deferred to a follow-up pull request.

## Testing

- `task lint` passes (all linters, including the `Lint` build with `Roslynator` active
  across all four projects).
- Verified RCS1213 actually runs, not just that the build is clean:
  - `dotnet build -p:ReportAnalyzer=true` lists
    `Roslynator.CSharp.Analysis.UnusedMember.UnusedMemberAnalyzer (RCS1213)` in the analyzer execution report.
  - Introducing an unused member fails the `Lint` build with `error RCS1213`; removing it restores a clean build.
- No new unit/integration tests — this is analyzer/tooling configuration with no runtime behaviour change.

## Related Issues

:white_circle: None.

## Checklist

- [x] This Pull Request is related to one issue.
- [x] Commit message has a detailed description of what changed and why.
- [x] Tests are added or updated and all checks pass.
- [x] ~~`CHANGELOG.md`, `README.md`, `DEVELOPER.md`, and other documentation files are updated.~~ (Dev-only tooling; no user-facing change to document.)
- [x] Destination branch is correct - `main` or release
- [x] Create merge commit if merging release branch into `main`, squash otherwise.

Issue #538, opened 2026-08-22
task(batch): remove `Async` suffix from sync batch commands


currantw:
### Description

The batch command methods on `IBatch*` interfaces and `BaseBatch<T>` carry an `Async`
suffix but are **synchronous** — they return `T` (for fluent chaining) or `IBatch`, never
a `Task`. This makes the suffix misleading and is flagged by Roslynator
[RCS1047](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1047) (*Non-asynchronous method name should not end with 'Async'*).

| File                                    | Methods | Kind             |
| --------------------------------------- | ------- | ---------------- |
| `BaseBatch.StringCommands.cs`           | 22      | implementation   |
| `BaseBatch.ServerManagementCommands.cs` | 15      | implementation   |
| `IBatchServerManagementCommands.cs`     | 15      | public interface |
| `BaseBatch.BitmapCommands.cs`           | 8       | implementation   |
| `BaseBatch.ConnectionManagement.cs`     | 7       | implementation   |
| `IBatchConnectionManagementCommands.cs` | 7       | public interface |

Because these members live on **public** `IBatch*` interfaces and `BaseBatch<T>`, renaming
them is a **breaking change**.

### Checklist

- [ ] Rename all `Batch` `*Async` methods and remove the RCS1047 suppressions and `TODO` comments.
- [ ] Add a `CHANGELOG.md` entry.

### Related

:white-circle: None

### Additional Information

:white-circle: None

PR #550, opened 2026-08-27
refactor(lint): enable Roslynator analyzers

currantw:
## Summary

Enables [Roslynator](https://github.com/dotnet/roslynator) analyzers using a **deliberate opt-in model**:

- All analyzers are disabled by default in `.editorconfig`
- Rules are enabled (marked as `error`) only if they are judged to add value.
- All resulting violations are fixed.

This pull request starts by evaluating and enabling a subset of analyzers.
Additional analyzers are left to future work – see `TODO` comments in `.editorconfig`.

The change is *overwhelmingly non-functional* (e.g. formatting, dead-code cleanup, doc comments, etc.).
A small number of public behaviour changes are recorded in `CHANGELOG.md`. I have also done my best to
identify any noteworthy changes to reviewers with comments.

## Issue Link

Part of #549.

- Enable **110 Roslynator (`RCS`) rules** as `error`, including some common analyzers (`RCS1xxx`) and formatting analyzers (`RCS0xxx`).
- Public behaviour/API changes (see `CHANGELOG.md`):
  - `Logger`, `InfoOptions`, and `Options` are now `static` classes. They only had static methods, so there is no expected or reasonable use case.
  - `ScoreBound.CompareTo(null)` now returns a positive value per the `IComparable` contract instead of throwing.

## Implementation

- `Roslynator.Analyzers` and `Roslynator.Formatting.Analyzers` are referenced as build-time-only analyzers.
- `.editorconfig` holds the opt-in list of enabled rules.
- Formatting-rule fixes were applied with the Roslynator analyzer code fixes.

## Limitations

- Additional analyzers are deferred to future PRs and tracked with TODO comments.

## Testing

- `Lint` build
- C# formatting: clean (`dotnet format --verify-no-changes`).
- Unit tests: all passing.

## Related Issues

- #537

## Checklist


sources/Valkey.Glide/Abstract/Condition.cs
```diff
@@ -407,9 +411,9 @@ internal bool Validate(object? res)
     /// <summary>
     /// Validates the result of the condition check against the expected outcome.
     /// </summary>
-    /// <param name="res">The result from the condition command execution.</param>
+    /// <param name="result">The result from the condition command execution.</param>
     /// <returns><see langword="true"/> if the condition is satisfied; otherwise, <see langword="false"/>.</returns>
-    protected abstract bool ValidateImpl(object? res);
+    protected abstract bool ValidateImpl(object? result);
```
currantw, line 416: **RCS1168** — Parameter name differs from base name.
Updated to match parameter name in descendants.

sources/Valkey.Glide/Abstract/ConfigurationOptions.cs
```diff
@@ -94,7 +96,6 @@ public static string TryNormalize(string value)
 
     private bool? _ssl;
     private Proxy? _proxy;
-    private RetryStrategy? _reconnectRetryPolicy;
```
currantw, line 97: Remove unnecessary field. Replaced with `ReconnectRetryPolicy`.

sources/Valkey.Glide/Abstract/Condition.cs
```diff
@@ -530,8 +535,7 @@ internal class ListCondition : Condition
         private readonly ValkeyValue? expectedValue;
         private readonly ValkeyKey key;
 
-        [System.Diagnostics.CodeAnalysis.SuppressMessage("Roslynator", "RCS1242:Do not pass non-read-only struct by read-only reference.", Justification = "Attribute")]
-        public ListCondition(in ValkeyKey key, long index, bool expectedResult, in ValkeyValue? expectedValue)
+        public ListCondition(in ValkeyKey key, long index, bool expectedResult, ValkeyValue? expectedValue)
```
currantw, line 538: **RCS1242** – Don't pass non-readonly struct by in
Updated to remove `in` for `expectedValue`. Internal so non-breaking change.

sources/Valkey.Glide/Abstract/ConnectionMultiplexer.cs
```diff
@@ -14,7 +14,7 @@ namespace Valkey.Glide;
 /// Connection methods common to both standalone and cluster clients.<br />
 /// See also <see cref="GlideClient" /> and <see cref="GlideClusterClient" />.
 /// </summary>
-public sealed class ConnectionMultiplexer : IConnectionMultiplexer, IDisposable, IAsyncDisposable
+public sealed class ConnectionMultiplexer : IConnectionMultiplexer
```
currantw, line 17: Redundant. Already inherited from `IConnectionMultiplexer`.

sources/Valkey.Glide/Abstract/ConnectionMultiplexer.cs
```diff
@@ -141,22 +140,21 @@ public IServer[] GetServers()
 
             return [.. servers];
         }
-        else
```
currantw, line 144: RCS1211: Remove unnecessary 'else'

sources/Valkey.Glide/Abstract/Condition.cs
```diff
@@ -541,21 +545,29 @@ public ListCondition(in ValkeyKey key, long index, bool expectedResult, in Valke
...
+                + index.ToString()
+                + "]"
+                + (expectedValue.HasValue ? (expectedResult ? " == " : " != ")
+                    + expectedValue.Value : (expectedResult ? " exists" : " does not exist"));
 
         internal sealed override List<ICmd> CreateCommands()
             => [
                 Request.Watch([key]),
-                Request.CustomCommand([ValkeyCommand.LINDEX.ToString(), key, index.ToString()]),
+                Request.ListGetByIndex(key, index),
```
currantw, line 558: RCS1015: Use 'nameof' operator
Ended up realizing that we have now implemented `ListGetByIndex`, so `CustomCommand` isn't even needed.

sources/Valkey.Glide/Abstract/Database.GenericCommands.cs
```diff
@@ -261,7 +259,7 @@ public async Task<ValkeyKey> KeyRandomAsync(CommandFlags flags = CommandFlags.No
     public Task<ValkeyValue[]> SortAsync(ValkeyKey key, long skip, long take, Order order, SortType sortType, ValkeyValue by, IEnumerable<ValkeyValue>? get, CommandFlags flags)
     {
         GuardClauses.ThrowIfCommandFlags(flags);
-        return ((IGenericBaseCommands)this).SortAsync(key, skip, take, order, sortType, by, get);
+        return SortAsync(key, skip, take, order, sortType, by, get);
```
currantw, line 262: Remove unnecessary cast

sources/Valkey.Glide/Abstract/Database.ListCommands.cs
```diff
@@ -10,19 +10,19 @@ internal partial class Database
...
 
     /// <inheritdoc cref="IDatabaseAsync.ListRightPushAsync(ValkeyKey, ValkeyValue, When)"/>
     public Task<long> ListRightPushAsync(ValkeyKey key, ValkeyValue value, When when)
-        => Command(Request.ListRightPushAsync(key, value, when));
+        => Command(Request.ListRightPush(key, value, when));
 
     /// <inheritdoc cref="IDatabaseAsync.ListRightPushAsync(ValkeyKey, IEnumerable{ValkeyValue}, When)"/>
     public Task<long> ListRightPushAsync(ValkeyKey key, IEnumerable<ValkeyValue> values, When when)
-        => Command(Request.ListRightPushAsync(key, [.. values], when));
+        => Command(Request.ListRightPush(key, [.. values], when));
```
currantw, line 25: RCS1047: Non-asynchronous method name should not end with 'Async'

sources/Valkey.Glide/Abstract/IDatabaseAsync.BitmapCommands.cs
```diff
@@ -1,13 +1,15 @@
...
-/// by StackExchange.Redis databases but NOT by Valkey GLIDE clients. Methods implemented
-/// by both should be added to <see cref="IBitmapBaseCommands"/> instead.
+// ATTENTION: Methods should only be added to this interface if they are implemented
+// by StackExchange.Redis databases but NOT by Valkey GLIDE clients. Methods implemented
+// by both should be added to IBitmapBaseCommands instead.
 
+/// <summary>
+/// Bitmap commands for StackExchange.Redis databases.
+/// </summary>
+/// <seealso href="https://valkey.io/commands/#bitmap">Valkey – Bitmap Commands</seealso>
```
currantw, line 12: Added for consistency across partials.

sources/Valkey.Glide/Abstract/LuaScript.cs
```diff
@@ -127,8 +124,8 @@ public static LuaScript Prepare(string script)
     /// <param name="flags">Command flags (currently not supported by GLIDE).</param>
     /// <exception cref="NotImplementedException">Thrown if <paramref name="flags"/> is not <see cref="CommandFlags.None"/>.</exception>
     /// <returns>A task representing the asynchronous operation, containing the result of the script execution.</returns>
-    /// <exception cref="ArgumentNullException">Thrown when db is null.</exception>
-    /// <exception cref="ArgumentException">Thrown when parameters object is missing required properties or has invalid types.</exception>
+    /// <exception cref="ArgumentNullException">Thrown if db is null.</exception>
+    /// <exception cref="ArgumentException">Thrown if parameters object is missing required properties or has invalid types.</exception>
```
currantw, line 128: Updated for consistency. Exceptions here use "Throw if" instead of "Throw when". Not a big deal but I like consistency 🤷

sources/Valkey.Glide/abstract_Enums/SER/CommandFlags.cs
```diff
@@ -32,7 +32,7 @@ public enum CommandFlags
     /// This operation should be performed on the primary if it is available, but read operations may
     /// be performed on a replica if no primary is available. This is the default option.
     /// </summary>
-    PreferMaster = 0,
+    PreferMaster = None,
```
currantw, line 35: Enums with same value.

sources/Valkey.Glide/Internals/FFI.enums.cs
```diff
@@ -12,7 +12,7 @@ internal partial class FFI
     /// The cache metric to retrieve from the client-side cache.<p/>
     /// Must match the glide-core cache metric methods.
     /// </summary>
-    internal enum CacheMetricsType : uint
+    internal enum CacheMetricsType : int
```
currantw, line 15: This was something that that I found by coincidence because of [RCS1042](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1042) (remove enum default underlying type). 

While these don't actually use the default type (`int`), investigating that elsewhere led to be finding that several of our enums actually use `uint` instead of `int`. This doesn't match the actual underlying Rust core, so updated these. Didn't even end up deciding to enable that rule, because I thought it was important to be able to be explicit about this matching to Rust core.

sources/Valkey.Glide/Internals/Request.GenericCommands.cs
```diff
@@ -28,7 +28,7 @@ public static Cmd<bool, bool> CopyAsync(ValkeyKey source, ValkeyKey destination,
         return Simple<bool>(RequestType.Copy, [.. args]);
     }
 
-    public static Cmd<bool, bool> CopyAsync(ValkeyKey source, ValkeyKey destination, int destinationDatabase, bool replace = false)
+    public static Cmd<bool, bool> Copy(ValkeyKey source, ValkeyKey destination, int destinationDatabase, bool replace = false)
```
currantw, line 31: As mentioned in another comments, these renames are triggered by 
RCS1047: Non-asynchronous method name should not end with 'Async'

While these `Request` methods are all internal and so safe to change, we actually expose some of these errors for our Batch commands, and raised https://github.com/valkey-io/valkey-glide-csharp/issues/538 to make this change in the next major release.

tests/Valkey.Glide.UnitTests/GlideStringTests.cs
```diff
@@ -9,7 +9,7 @@ public void Sorting()
     {
         gs[] arr = ["abc", "abcd", "abcde", "abd", "abb", "ab1"];
         Assert.Equal(new gs[] { "ab1", "abb", "abc", "abd", "abcd", "abcde" }, [.. arr.Order()]);
-        Assert.Equal(new gs[] { "ab1", "abb", "abc", "abd", "abcd", "abcde" }, [.. arr.OrderBy(s => s)]);
+        Assert.Equal(new gs[] { "abc", "abd", "abb", "ab1", "abcd", "abcde" }, [.. arr.OrderBy(s => s.Length)]);
```
currantw, line 12: Updated to a non-trivial `OrderBy`

sources/Valkey.Glide/Abstract/LuaScript.cs
```diff
@@ -69,10 +69,7 @@ internal LuaScript(string originalScript, string executableScript, string[] argu
     /// </example>
     public static LuaScript Prepare(string script)
     {
-        if (string.IsNullOrEmpty(script))
-        {
-            throw new ArgumentException("Script cannot be null or empty", nameof(script));
-        }
+        ArgumentNullException.ThrowIfNullOrEmpty(script, nameof(script));
```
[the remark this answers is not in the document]
currantw: Addressed in 1442c63.

sources/Valkey.Glide/GlideString.cs
```diff
@@ -336,13 +338,22 @@ public static implicit operator GlideString(ValkeyKey key)
         return keyBytes == null ? new GlideString([]) : new GlideString(keyBytes);
     }
 
+    /// <inheritdoc/>
+    int IComparable.CompareTo(object? obj) => obj switch
+    {
+        null => CompareTo(null),
```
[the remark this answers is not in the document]
currantw: Addressed in 1442c63.

currantw:
The vast majority of changes in this pull request are straight-forward, mechanical changes to resolve newly-enabled Roslynator violations. However, there are a small number of non-trivial changes that were made in the course of fixing these violations, and which are called out below:

### 1. `ScoreBound.CompareTo(null)` threw instead of returning positive

- File(s): `ScoreBound.cs`
- Rule: [RCS1256: Invalid argument null check](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1256/)
- Commit(s): dbc07d79

`ScoreBound.CompareTo(null)` threw `ArgumentNullException`. It now returns `1`, which is consistent with the
`IComparable` contract (any instance sorts greater than `null`). A `CHANGELOG.md` entry was also added.

### 2. `GlideString.CompareTo(null)` returned `-1` (violated `IComparable`)

- File(s): `GlideString.cs`
- Rule: [RCS1241: Implement non-generic counterpart](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1241/)
- Commit(s): 58d9566c, 1442c634

RCS1241 prompted adding a non-generic `IComparable` that delegates to the generic overload (58d9566c), which
surfaced that `CompareTo(null)` returned `-1`. Copilot flagged that non-generic sorting would then order real
values as smaller than `null`, inconsistent with the `ScoreBound` fix (#1). It now returns a positive value
(1442c634, "Addressed in 1442c63"). A `CHANGELOG.md` entry was added.

### 3. `MessageContainer.Dispose` silently swallowed exceptions

- File(s): `MessageContainer.cs`
- Rule: [RCS1075: Avoid empty catch clause that catches System.Exception](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1075/)
- Commit(s): a05a3c2c

`Dispose` used an empty catch while cancelling pending messages. It now logs the exception via
`Logger.Log(Level.Warn, ...)`.

### 4. `ValkeyValue.CompareTo` used `throw ex;` (reset the stack trace)

- File(s): `ValkeyValue.cs`
- Rule: [RCS1044: Remove original exception from throw statement](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1044/)
- Commit(s): 4420865c

The surrounding catch-all `try`/`catch` existed only to re-throw , so its only real effect was resetting the stack trace. The whole `try`/`catch` was removed — behavior-identical but the full stack trace is now preserved.

### 5. FFI enums retyped `uint` → `int` to match the Rust core

- File(s): `FFI.enums.cs`, `ResponseHandler.cs`
- Rule: [RCS1042: Remove enum default underlying type](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1042/) (investigated, ultimately *not* enabled)
- Commit(s): 90990e13

`CacheMetricsType`, `PeriodicChecksMode`, `RouteType`, `ServiceType`, `TlsMode`, and `ValueType` were declared
`uint`, but their glide-core counterparts are signed `#[repr(C)]` / protobuf `int32`. They were switched to
`int` to mirror the Rust core.

### 6. `ListCondition` uses typed `ListGetByIndex` instead of `CustomCommand`

- File(s): `Condition.cs`
- Rule: [RCS1015: Use 'nameof' operator](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1015/)
- Commit(s): 0a11ea94

Rather than add `nameof` to a literal-command-name `CustomCommand`, the code now calls
`Request.ListGetByIndex(key, index)`, removing the command-name string entirely.

### 7. `RCS1037`/`RCS1085` cleanup removed dead code in `ValkeyValue` and `ConfigurationOptions`

- File(s): `ConfigurationOptions.cs`, `ValkeyValue.cs`
- Rule: [RCS1085: Use auto-implemented property](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1085/)
- Commit(s): 17667aa3

`ConfigurationOptions.ReconnectRetryPolicy` was converted to an auto-property (the `_reconnectRetryPolicy`
backing field was removed and `Clone` updated). In `ValkeyValue`, the `OverlappedValueInt64` passthrough was inlined and two unused dead accessors (`DirectObject`, `DirectOverlappedBits64`) were removed.

### 8. `expectedValue` no longer passed by `in` (`Condition`)

- File(s): `Condition.cs`
- Rule: [RCS1242: Do not pass non-read-only struct by read-only reference](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1242/)
- Commit(s): 2c776785

RCS1242 (adopted in the bulk zero-violation commit) flagged the `in` on `expectedValue` in the `ListCondition`
constructor. The `in` was removed. The member is internal, so this is non-breaking.

sources/Valkey.Glide/Valkey.Glide.csproj
```diff
@@ -55,6 +55,10 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="Roslynator.Formatting.Analyzers" Version="4.16.1">
```
nderraugh, line 58: Roslynator.Formatting.Analyzers is referenced only when Configuration=Lint, but task format:csharp runs plain dotnet format under the default configuration. So it can't load or fix the newly enforced RCS0 rules.

The Lint build still detects violations, so enforcement works, but the repository’s formatter cannot remediate them. Run dotnet format with the Lint MSBuild configuration/property or make the analyzer references available to the formatter.

currantw: Yes! Plan to do that in a follow-up PR for #549 🚀

sources/Valkey.Glide/abstract_APITypes/ValkeyValue.cs
```diff
@@ -365,56 +367,48 @@ internal StorageType Type
...
-
-            // otherwise, compare as strings
-            return string.CompareOrdinal((string?)x, (string?)y);
         }
-        catch (Exception ex)
-        {
-            //ConnectionMultiplexer.TraceWithoutContext(ex.Message);
-            throw ex;
+
+        switch (xType)
```
nderraugh, line 394: Need unit tests here for comparisons (and ordering).

There's some conversions here that will drop precision, like a long as a double.

Two values can have CompareTo() == 0 while Equals() is false because equality rejects differing numeric storage kinds.

NaN compares equal for ordering purposes but is unequal under the current ==.

Raw/string comparison decodes bytes as UTF-8. Different invalid byte sequences can decode to the same replacement characters, potentially violating equality transitivity.

And later (L:557), retained byte[] storage is mutable. And mutating it after inserting the value into a dictionary or hash set can change its equality and hash code.

This would be a great place to introduce property based testing.  I did a quick google and https://github.com/AnthonyLloyd/CsCheck look promising.

nderraugh (approved):

sources/Valkey.Glide/abstract_APITypes/ValkeyValue.cs
currantw: Raised #555 and tentatively assigned it to the next release milestone.

.editorconfig
```diff
@@ -562,14 +562,256 @@ dotnet_diagnostic.xUnit3002.severity = error
...
+dotnet_diagnostic.RCS0050.severity = error
+dotnet_diagnostic.RCS0054.severity = error
+dotnet_diagnostic.RCS0055.severity = error
+dotnet_diagnostic.RCS0060.severity = error
+dotnet_diagnostic.RCS0061.severity = error
+dotnet_diagnostic.RCS0063.severity = error
+
+roslynator_blank_line_between_single_line_accessors = false
+roslynator_blank_line_after_file_scoped_namespace_declaration = true
+roslynator_blank_line_between_switch_sections = true
```
Aryex: Seems that `= true` isn't a valid value.

The [RCS0061 docs](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS0061/) list `include|omit|omit_after_block`. Right now this rule is disabled and not checked. Can you confirm?

Aryex (commented):
Reviewed up to dc9793f.

.editorconfig
currantw: Thanks for catching this! Did not want to implement this rule (I think it doesn't really provide much value for simple `switch` statements), but looks like it accidentally crept in! Removed in c2f05a26.

Aryex (approved):

PR #557, opened 2026-09-04
refactor(lint): enable RCS0053 and consolidate response converters

currantw:
## Summary

- Enables the Roslynator [RCS0053](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS0053/) analyzer ("Fix formatting of a list")
- Applies the resulting list-formatting fixes across the codebase.
- Consolidates some response-parsing helpers in the `Request` layer that were surfaced while working in that code.

## Issue Link

Part of #549.

## Implementation

- RCS0053
  - Enabled RCS0053 in `.editorconfig`.
  - Ran the formatter to apply list-wrapping fixes.
- Response-converter consolidation:
  - Added `ToLongs`, `ToNullableLongs`, and `ToDouble` parsers, and reused them in `BitField(ReadOnly)`, `ListPositions`, `WaitAof`, and the Set/SortedSet scan cursor + score parsing.
  - Routed MGET, LRANGE, and the Set/SortedSet/Stream `ValkeyValue[]` builders through the shared `ToValkeyValues` factory.
  - Had `TryGetDouble` delegate to `ToDouble` (fixes a locale-dependent parse).
  - Removed a duplicate `ToValkeyValues` definition.

## Limitations

:white_circle: None.

## Testing

Verify that pipeline passes.

## Checklist

- [x] `CHANGELOG.md` updated with the culture-invariant numeric-parsing fix. `README.md`/`DEVELOPER.md` not applicable.


tests/Valkey.Glide.UnitTests/PubSubPerformanceTests.cs
```diff
@@ -265,11 +262,9 @@ public void ChannelBasedProcessing_LongRunning_MaintainsStablePerformance()
         var throughputStdDev = Math.Sqrt(throughputSamples.Average(t => Math.Pow(t - avgThroughput, 2)));
 
         // Verify stable performance over time
-        Assert.True(avgThroughput >= targetRate * 0.8,
-            $"Average throughput {avgThroughput:F0} is below 80% of target rate {targetRate}. Processed {messagesReceived} messages over {duration} seconds. Min: {minThroughput:F0}, Max: {maxThroughput:F0}, StdDev: {throughputStdDev:F0}");
+        Assert.True(avgThroughput >= targetRate * 0.8);
```
currantw: We don't have test failures message in the vast majority of assertions, and I am skeptical they add much value here. Removed for consistency and brevity. 🤷

tests/Valkey.Glide.IntegrationTests/ClusterClientTests.cs
```diff
@@ -444,8 +444,7 @@ public async Task TimeAsync_ReturnsTimePerNode(GlideClusterClient client)
         foreach (var serverTime in allTimes.Values)
         {
             TimeSpan diff = (serverTime - localTime).Duration();
-            Assert.True(diff < TimeSpan.FromSeconds(10),
-                $"Server time {serverTime} differs from local time {localTime} by {diff}");
+            Assert.True(diff < TimeSpan.FromSeconds(10));
```
xShinnRyuu: **Medium.** This assertion lost the only information that made a failure actionable.

The loop at line 444 walks every cluster node. When one node's clock drifts past 10 seconds, the failure now reads `Assert.True() Failure` and nothing else: no node, no server time, no delta. The deleted message carried all three. Clock comparisons against a live server fail intermittently in CI, and by the time anyone opens the run the servers are gone, so the log text is the whole investigation.

RCS0053 wraps argument lists. It does not require deleting arguments, so the message can come back as-is:

```csharp
Assert.True(
    diff < TimeSpan.FromSeconds(10),
    $"Server time {serverTime} differs from local time {localTime} by {diff}");
```

tests/Valkey.Glide.UnitTests/PubSubPerformanceTests.cs
```diff
@@ -95,13 +94,11 @@ public void ChannelBasedProcessing_ReducedAllocationPressure_MinimizesGCImpact()
         var gen2Collections = finalGen2 - initialGen2;
 
         // Verify reasonable memory growth (should be less than 10MB for 50k messages)
-        Assert.True(memoryGrowth < 10_000_000,
-            $"Memory grew by {memoryGrowth:N0} bytes ({memoryGrowth / 1024.0 / 1024.0:F2} MB) - excessive allocation pressure. Gen0: {gen0Collections}, Gen1: {gen1Collections}, Gen2: {gen2Collections}");
+        Assert.True(memoryGrowth < 10_000_000);
```
xShinnRyuu: **Medium.** Dropping the messages in this file left five locals that are computed and never read: `gen0Collections` (92), `gen1Collections` (93), `maxBurstTime` (188), `minThroughput` (260), `maxThroughput` (261). Two `GC.CollectionCount` deltas and three LINQ passes still run and nothing consumes the result. I built the solution with `Configuration=Lint` and it reports 0 warnings, so nothing in the gate catches it. Either the values go back into the messages or the locals should go.

On the removal itself, the rest of this PR argues for keeping them. `PubSubFFIMemoryLeakTests.cs:62`, `PubSubThreadSafetyTests.cs`, and `CompressionTests.cs` all keep the same style of interpolated message and just reflow it. These particular assertions are throughput and GC-pressure thresholds, where the measured number is what tells you whether a red build is a real regression or a busy runner.

sources/Valkey.Glide/Internals/Request.cs
```diff
@@ -261,13 +287,7 @@ private static double GetDouble(Dictionary<GlideString, object> map, string key)
...
     /// <exception cref="RequestException">Thrown if the value for <paramref name="key"/> is not a double or string.</exception>
     private static double? TryGetDouble(Dictionary<GlideString, object> map, string key)
-        => map.TryGetValue(key, out var value)
-            ? value switch
-            {
-                double d => d,
-                GlideString gs => double.Parse(gs.ToString()),
-                _ => throw new RequestException($"Response field '{key}' expected double or string, got {value.GetType()}"),
-            } : null;
+        => map.TryGetValue(key, out var value) ? ToDouble(value) : null;
```
xShinnRyuu: **Low.** Delegating to `ToDouble` drops the field name from the exception.

`ToDouble` throws `Expected a double or numeric string, got {type}` with no key. `GetDouble` has seven callers: the six `MEMORY STATS` ratio fields at `Request.ServerManagement.cs:281-286` and `backfill_complete_percent` at `Request.FtCommands.cs:530`. If a server returns any of those six as a RESP integer, the caller now gets `Expected a double or numeric string, got System.Int64` with no way to tell which field failed. The old message named it.

The neighbours still do: `TryGetLong` at line 323 reports `Response field '{key}' expected long or string`, and so do `GetChar` (271) and the array helper (417). Threading the key through keeps the shared parser and the diagnostic:

```csharp
private static double ToDouble(object value, string? key = null) => value switch
{
    double d => d,
    GlideString gs => double.Parse(gs.ToString(), CultureInfo.InvariantCulture),
    _ => throw new RequestException(key is null
        ? $"Expected a double or numeric string, got {value.GetType()}"
        : $"Response field '{key}' expected double or string, got {value.GetType()}"),
};
```

sources/Valkey.Glide/Internals/Request.cs
```diff
@@ -199,6 +199,32 @@ private static int ToInt(object value)
...
 
+    /// <summary>
+    /// Parses a response value as a <see langword="double"/>.
+    /// </summary>
+    /// <param name="value">The response value to parse.</param>
+    /// <exception cref="RequestException">Thrown if <paramref name="value"/> is not a double or numeric string.</exception>
+    private static double ToDouble(object value) => value switch
+    {
+        double d => d,
+        GlideString gs => double.Parse(gs.ToString(), CultureInfo.InvariantCulture),
```
xShinnRyuu: **Low.** This line is a user-visible fix, and the PR checklist is filed as though nothing user-visible changed.

Adding `CultureInfo.InvariantCulture` changes results on any machine whose culture uses a comma decimal separator. Under `de-DE`, `ParseScanResponse` previously turned a ZSCAN score of `1.5` into `15`, and the `MEMORY STATS` ratio fields were wrong the same way. Anyone who hit that would want to see it in the release notes, but the CHANGELOG checkbox is struck out as "no user-facing or documented behavior changed".

#550, also `refactor(lint)`, took the other route and added `CompareTo(null) returns a positive value for ScoreBound and GlideString (#550)` under `### Fixed` for the same kind of incidental fix.

Nothing pins the new behavior either. CA1305 is not enabled: `TryGetLong` at line 329 still calls `long.Parse(gs.ToString())` with no provider and the Lint build stays clean. A later pass can drop the provider again with nothing to catch it. Worth a `### Fixed` entry plus a test that parses a score under a comma-decimal culture.

tests/Valkey.Glide.IntegrationTests/ServerModules/JsonCommandTests.cs
```diff
@@ -1255,8 +1255,7 @@ public async Task TypeAsync_NumberType_ReturnsNumberOrInteger(BaseClient client)
         Assert.NotNull(intResult);
         Assert.NotNull(floatResult);
         // Integer values may return "integer" or "number" depending on the JSON module version
-        Assert.True(intResult is "integer" or "number",
-            $"Expected 'integer' or 'number' but got '{intResult}'");
+        Assert.True(intResult is "integer" or "number");
```
xShinnRyuu: **Low.** Here the deleted message was the point of the assertion.

The comment on line 1257 says the value varies by JSON module version, so a failure means the module returned some third thing. The old message told you what. Now you get `Assert.True() Failure` and have to reproduce against the same module build to learn anything. Same at 1319 and 1420.

`Assert.Equal` would carry the actual value for free and reads better than `Assert.True` on a string comparison, if you want the shorter call without losing the diagnostic.

xShinnRyuu (commented):
Reviewed the head commit. The RCS0053 reflow is clean and the `.editorconfig` change is correctly scoped, but the converter consolidation and a handful of deleted assertion messages need another look. Notes inline.

tests/Valkey.Glide.IntegrationTests/ClusterClientTests.cs
currantw: Addressed in `ac12359f` — switched to `Assert.InRange` so the actual delta is reported on failure, but we follow the general codebase pattern of not using custom failure messages where not needed.

tests/Valkey.Glide.UnitTests/PubSubPerformanceTests.cs
currantw: Addressed in `ac12359f` — restored the messages.

tests/Valkey.Glide.IntegrationTests/ServerModules/JsonCommandTests.cs
currantw: Addressed in `ac12359f` — restored strict `is "integer" or "number"` checks with interpolated messages that report the actual value.

sources/Valkey.Glide/Internals/Request.cs
```diff
@@ -397,7 +417,7 @@ private static ValkeyValue[] GetValkeyValues(Dictionary<GlideString, object> map
             _ => throw new RequestException($"Response field '{key}' expected array, got {value.GetType()}"),
         };
 
-        return ToValkeyValueArray(items);
+        return ToValkeyValues(items);
```
Aryex: Is the `HashSet<object>` arm on line 415 still meant to be reachable?

`items` is `IEnumerable<object>`, which has no implicit conversion to `object[]`, so this binds to
`ToValkeyValues(object)` and its `(object[])value` cast. Neither `HandleServerValue` nor
`Cmd.GetConverter()` catches, so the failure surfaces at the caller instead of as the `RequestException`
on line 417.

I built the same overload shape standalone with the project's settings (net8.0, SDK 10.0.105,
`TreatWarningsAsErrors`). It compiles with 0 warnings, so the Lint configuration never flags it:

```
object[]         -> ToValkeyValues(object) -> [a],[b]
HashSet<object>  -> ToValkeyValues(object) -> InvalidCastException: Unable to cast
    'System.Collections.Generic.HashSet`1[System.Object]' to 'System.Object[]'
```

RESP3 sets arrive as `HashSet<object>` (`ffi.rs:677`, `ResponseHandler.cs:81`), RESP3 is the default
(`ConnectionConfiguration.cs:731`), and `FTInfoReturnType` recurses only into `fields` and
`vector_params`, so `prefixes` and `stop_words` reach here unconverted.

I could not confirm whether valkey-search sends `~` for those two fields. The module job skips 676 of 680
tests, including all 40 `FtInfoTests` variants.

sources/Valkey.Glide/Internals/Request.cs
Aryex: The error no longer names the field, but the doc comment on line 288 still says it does ("Thrown if the
value for `key` is not a double or string").

```
before:  Response field 'allocator-rss.ratio' expected double or string, got System.Object[]
after:   Expected a double or numeric string, got System.Object[]
```

Six of the seven `GetDouble` callers read from the same `MEMORY STATS` dictionary
(`Request.ServerManagement.cs:281-286`), so the message no longer says which one failed. `TryGetLong` on
line 330 still includes the key.

sources/Valkey.Glide/Internals/Request.cs
```diff
@@ -199,6 +199,32 @@ private static int ToInt(object value)
...
+    /// <summary>
+    /// Parses a response value as a <see langword="double"/>.
+    /// </summary>
+    /// <param name="value">The response value to parse.</param>
+    /// <exception cref="RequestException">Thrown if <paramref name="value"/> is not a double or numeric string.</exception>
+    private static double ToDouble(object value) => value switch
+    {
+        double d => d,
+        GlideString gs => double.Parse(gs.ToString(), CultureInfo.InvariantCulture),
+        _ => throw new RequestException($"Expected a double or numeric string, got {value.GetType()}"),
```
Aryex: `null` matches the `_` arm, and `value.GetType()` then dereferences it, so a nil element throws
`NullReferenceException` rather than the documented `RequestException`. Same shape at line 199 in
`ToLong`.

Not reachable from the three new `ToLongs` callers: only `BitFieldGet` implements
`IBitFieldReadOnlySubCommand`, WAITAOF elements are `Value::Int`, and `LPOS ... COUNT` returns no nil
members.

tests/Valkey.Glide.IntegrationTests/ClusterClientTests.cs
```diff
@@ -443,9 +443,8 @@ public async Task TimeAsync_ReturnsTimePerNode(GlideClusterClient client)
         // Each node should have a reasonable time
         foreach (var serverTime in allTimes.Values)
         {
-            TimeSpan diff = (serverTime - localTime).Duration();
-            Assert.True(diff < TimeSpan.FromSeconds(10),
-                $"Server time {serverTime} differs from local time {localTime} by {diff}");
+            var diff = (serverTime - localTime).Duration();
+            Assert.InRange(diff, TimeSpan.Zero, TimeSpan.FromSeconds(10));
```
Aryex: `Assert.InRange` has no message overload in xunit.v3 3.2.2, only `(T,T,T)` and `(T,T,T,IComparer<T>)`, so
`serverTime` and `localTime` are no longer reported on failure.

It is also inclusive on both bounds, so a 10.000s drift now passes where `diff <
TimeSpan.FromSeconds(10)` failed, and `TimeSpan.Zero` cannot fail because `.Duration()` is non-negative.

`StandaloneClientTests.cs:453-455` keeps its message in RCS0053-compliant form.

tests/Valkey.Glide.IntegrationTests/ServerModules/FtUtils.cs
```diff
@@ -58,17 +58,19 @@ public static async Task WaitForIndexingAsync(BaseClient client, string indexNam
...
-            Prefixes = [prefix],
-        });
+        await Ft.CreateAsync(
+            client,
+            index,
+            [
+                    new Ft.CreateTextField("title"),
+                    new Ft.CreateNumericField("price"),
+                    new Ft.CreateTagField("category"),
+                ],
```
Aryex: The `[` moved 4 columns but the block interior moved 8, so elements sit at +8 from the opener and `]` at
+4. Here `[` is at column 12, elements at 20, and `],` at 16. The base had 8/12/8.

RCS0053 does not require this. Against Roslynator.Formatting.Analyzers 5.0.0 with this `.editorconfig`,
aligning to 12/16/12 builds with 0 errors, and `dotnet format whitespace` leaves it unchanged.

Also `FtUtils.cs` 88, 96, 104, 131, 153, 161, 169, 177, 185; `BatchTestUtils.cs` 1263, 1486, 1493, 1500,
1509; `StackExchange/SortedSetCommandTests.cs` 879, 905, 980;
`StackExchange/GeospatialCommandsTests.cs` 338.

sources/Valkey.Glide/PubSubMessage.cs
```diff
@@ -71,10 +71,8 @@ public override string ToString()
             Pattern = Pattern?.ToString()
         };
 
-        return JsonSerializer.Serialize(messageObject, new JsonSerializerOptions
-        {
-            WriteIndented = false
-        });
+        var options = new JsonSerializerOptions { WriteIndented = false };
```
Aryex: This and `Abstract/ScriptParameterMapper.cs` (lines 79, 153, 201) are the only non-whitespace changes
outside what the description covers, and they are not mentioned in the body or in any commit message.
Both look semantics-preserving.

sources/Valkey.Glide/Internals/Request.cs
currantw: Restored exception with field name in 6cf526ad

sources/Valkey.Glide/Internals/Request.cs
currantw: Added a CHANGELOG entry in 338e740a. This is a desirable change – the format returned from the Valkey server does not vary by culture, so neither should our parsing.

sources/Valkey.Glide/Internals/Request.cs
currantw: Good catch — that `HashSet<object>` arm was dead (and broken): the only fields routed here, FT.INFO `prefixes`/`stop_words`, are RESP arrays, and set/map responses use `ToValkeyValueSet`/dictionary converters instead. Removed the arm in d08f0885, so a non-array now yields a clean `RequestException` rather than an `InvalidCastException`.

Aryex (approved):

sources/Valkey.Glide/Internals/Request.cs
```diff
@@ -299,16 +342,16 @@ private static long GetLong(Dictionary<GlideString, object> map, string key)
...
-                GlideString gs => long.Parse(gs.ToString()),
-                _ => throw new RequestException($"Response field '{key}' expected long or string, got {value.GetType()}"),
-            } : null;
+    {
+        if (!map.TryGetValue(key, out var value))
+        {
+            return null;
+        }
+
+        return TryToLong(value) ?? throw new RequestException($"Could not convert '{key}' field {value.GetType()} value '{value}' to long.");
```
xShinnRyuu: **Blocking.** This drops the explicit `null` arm the old `TryGetLong` had, and the replacement dereferences `value` while building the error message.

Before, the switch had `null => null`, so a field present in the map with a nil value returned `null`. Now `TryToLong(null)` falls through to `_ => null`, the `??` fires, and `value.GetType()` throws `NullReferenceException` instead of a `RequestException`. A nil map value is reachable: `ResponseHandler.TraverseValue` maps `ValueType.Null` to `null` and `CreateMap` stores it under the key, so `TryGetValue` returns true with a null value.

`XINFO GROUPS` hits this. The server reports `entries-read` and `lag` as nil when the group's lag cannot be determined, and `Request.StreamCommands.cs:265-266` reads both with `TryGetLong`. The failing `tests` check on this head is this bug, not a flake:

```
System.NullReferenceException : Object reference not set to an instance of an object.
  at Valkey.Glide.Internals.Request.TryGetLong(Dictionary`2 map, String key) in sources/Valkey.Glide/Internals/Request.cs:353
  at Request.<ConvertStreamGroupInfoResponses>b__408_0(Object response) in sources/Valkey.Glide/Internals/Request.StreamCommands.cs:260
```

`StreamConsumerGroupTests.StreamInfoGroupsAsync_Basic` fails on all four client variants (GlideClient RESP2 and RESP3, GlideClusterClient RESP2 and RESP3). `Request.StreamCommands.cs:249-250`, `286` and `299` read the same fields and fail the same way.

Guarding the null restores the old semantics exactly:

```csharp
private static long? TryGetLong(Dictionary<GlideString, object> map, string key)
{
    if (!map.TryGetValue(key, out var value) || value is null)
    {
        return null;
    }

    return TryToLong(value) ?? throw new RequestException($"Could not convert '{key}' field {value.GetType()} value '{value}' to long.");
}
```

`TryGetDouble` at line 312 has the same null dereference. That one predates this PR, but the same guard covers it and it is cheap to fix while you are here.

sources/Valkey.Glide/Internals/Request.cs
```diff
@@ -390,14 +433,9 @@ private static ValkeyValue[] GetValkeyValues(Dictionary<GlideString, object> map
...
-        {
-            object[] arr => arr,
-            HashSet<object> set => set,
-            _ => throw new RequestException($"Response field '{key}' expected array, got {value.GetType()}"),
-        };
-
-        return ToValkeyValueArray(items);
+        return value is object[] arr
+            ? ToValkeyValues(arr)
+            : throw new RequestException($"Could not convert '{key}' field {value.GetType()} value '{value}' to array.");
```
xShinnRyuu, line 442: **Blocking.** Narrowing the shared `ToValkeyValues` from `IEnumerable<object>` to `object[]` silently deleted a case this helper needs.

The old body was:

```csharp
IEnumerable<object> items = value switch
{
    object[] arr => arr,
    HashSet<object> set => set,
    _ => throw new RequestException($"Response field '{key}' expected array, got {value.GetType()}"),
};
return ToValkeyValueArray(items);
```

The `HashSet<object>` arm is reachable. `ResponseHandler.TraverseValue` maps `ValueType.Set` to `CreateArray(value).ToHashSet()`, so any RESP3 set reaching this helper is a `HashSet<object>` rather than `object[]`, and now takes the `throw`.

Both callers are `FT.INFO`: `Request.FtCommands.cs:522` reads `prefixes` via `GetValkeyValues`, and `Request.FtCommands.cs:535` reads `stop_words` via `TryGetValkeyValues`. If valkey-search returns either as a set under RESP3, `FT.INFO` now throws `RequestException` where it previously returned the values.

Nothing in CI would catch it. Every `FtInfoTests` case is skipped in the failing `tests` job with "Valkey Search module is not loaded on the server", and the separate module job runs only the four JSON tests.

Keeping the shared helper's parameter as `IEnumerable<object>`, or restoring the set arm here, preserves the behaviour. If the arm really is dead, that is worth its own change that says so, rather than a side effect of a signature narrowing.

sources/Valkey.Glide/Internals/Request.BitmapCommands.cs
```diff
@@ -52,19 +52,17 @@ public static Cmd<long, long> BitOp(Bitwise operation, ValkeyKey destination, Va
         }
 
         // Preserve null values to indicate overflow with OVERFLOW FAIL
-        return new(RequestType.BitField, [.. args], false, response =>
-            [.. response.Select(item => item is null ? (long?)null : Convert.ToInt64(item))]);
+        return new(RequestType.BitField, [.. args], false, ToNullableLongs);
```
xShinnRyuu, line 55: **Non-blocking, but worth deciding before merge.** `ToNullableLongs` makes `null` mean two different things, and in `BITFIELD` `null` is load-bearing.

The comment two lines up is right that `null` marks an overflow under `OVERFLOW FAIL`. The old converter produced `null` only for a literal null element and ran `Convert.ToInt64` on everything else, so an element it could not read raised. `TryToLong` returns `null` both for a nil element and for any element that is neither `long` nor `GlideString`, so an unreadable element is now handed to the caller as a successful "this one overflowed".

Concretely: if a `BITFIELD` element ever arrives as something other than `long` or `GlideString` (`ResponseHandler.TraverseValue` also produces `double` for `ValueType.Float` and `bool` for `ValueType.Bool`), `BitFieldAsync` returns `null` at that index and the caller reads it as an overflow rather than seeing an error. The old `Convert.ToInt64` would have converted a `double` or thrown.

`ToLongs` on the next builder has the opposite behaviour, because it goes through `ToLong`, which throws. Having the nullable variant swallow the same condition is the part worth making explicit: if the intent is "nil means overflow, anything else is a protocol error", the nullable parser has to keep those two apart.

CHANGELOG.md
```diff
@@ -77,6 +77,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `GlideString(byte[])` no longer builds the hex-dump representation on construction (#522)
 - `CompareTo(null)` returns a positive value for `ScoreBound` and `GlideString` (#550)
 - `ClusterValue<T>.HasSingleData` returns expected result for value-type `T` (#547)
+- Culture-invariant parsing of numeric string responses (#557)
```
xShinnRyuu: Following up on my earlier comment about the `CultureInfo.InvariantCulture` change: your argument is right, and this is the entry I asked for, but the entry does not say what changed for a user.

The server's number format does not vary by culture, so parsing it invariantly is the correct behaviour, agreed. What a reader needs from the changelog is the observable effect: results change on a machine whose culture uses a comma decimal separator. Under `de-DE`, a ZSCAN score of `1.5` previously came back as `15`, and the `MEMORY STATS` ratio fields were wrong the same way. As written, "Culture-invariant parsing of numeric string responses" states the mechanism, so someone who hit the old behaviour would not recognise their bug in it.

Compare the #550 entry three lines up, which names the observable behaviour rather than the mechanism. Something like:

```
- Numeric string responses such as sorted set scores and `MEMORY STATS` ratios are parsed with the invariant culture, and are no longer misread on hosts whose culture uses a comma decimal separator (#557)
```

sources/Valkey.Glide/Internals/Request.cs
```diff
@@ -213,6 +231,30 @@ private static DateTimeOffset ToDateTimeOffset(object value)
...
 
+    /// <summary>
+    /// Attempts to parse a response value as a <see langword="double"/>.
+    /// </summary>
+    /// <param name="value">The response value to parse.</param>
+    /// <exception cref="FormatException">Thrown if <paramref name="value"/> could not be parsed as a <see langword="double"/>.</exception>
+    private static double? TryToDouble(object? value) => value switch
+    {
+        double d => d,
+        GlideString gs => double.Parse(gs.ToString(), CultureInfo.InvariantCulture),
```
xShinnRyuu: Second half of my earlier comment on this change, still open: nothing pins the invariant behaviour, and two parsers in this layer were left culture-sensitive.

No test was added for it. Every test file in this PR is whitespace-only apart from `ClusterClientTests.cs` and `JsonCommandTests.cs`, so a later pass can drop the provider here and the gate stays green. CA1305 is still not enabled.

Two response parsers still use the ambient culture, which makes the CHANGELOG claim of culture-invariant parsing of numeric string responses broader than what the diff does:

- `Request.GeospatialCommands.cs:103` builds a `GeoPosition` from two bare `double.Parse` calls. `GEOPOS` returns latitude and longitude as strings, so under `de-DE` a longitude of `13.361389` parses as `13361389`. Same bug class as the one being fixed, in a file this PR already touches.
- `Request.cs:392` parses the server duration string in `GetTimeSpan` with a bare `double.Parse`.

Routing both through `TryToDouble` closes the class, and a test that parses a score with `CultureInfo.CurrentCulture` set to a comma-decimal culture is what keeps it closed.

xShinnRyuu (dismissed):

sources/Valkey.Glide/Internals/Request.cs
currantw: Fixed in adb9bcb9 — restored the `|| value is null` guard.

sources/Valkey.Glide/Internals/Request.cs
currantw: The FT.INFO docs list both `prefixes` and `stop_words` as arrays, so I don't think this change is needed (https://valkey.io/commands/ft.info/). Only parsing this as an array also seems consistent with how other clients parse this response.

sources/Valkey.Glide/Internals/Request.BitmapCommands.cs
currantw: Addressed in adb9bcb9 — `TryToLong`/`TryToDouble` now map `null` → `null` and throw on any other unexpected type.

xShinnRyuu (approved):

PR #564, opened 2026-09-16
feat: add AzAffinityAllNodes read strategy

### LINKED ISSUE #533: [Task] CSharp: Add AzAffinityAllNodes read strategy
### Description

GLIDE added a new read strategy, `AZAffinityAllNodes`, in valkey-io/valkey-glide#6721, which closed valkey-io/valkey-glide#6653. Only the Python binding was wired up, so every other client needs the same surface. This issue tracks the C# side.

The strategy spreads read requests equally, in round robin, across all nodes in the client's Availability Zone, primary and replicas alike. When no node in the client's AZ is available it falls back to a round robin across all nodes. It differs from `AzAffinityReplicasAndPrimary` in that it does not prioritise replicas ahead of the primary, which is what makes an even per-node read distribution possible on ElastiCache deployments that hit the 5 replica cap. The motivation is written up in valkey-io/valkey-glide#6653.

**Naming, please read before starting.** `ReadFromStrategy` already has an `AllNodes` member, documented as "Read from all nodes (primary and replicas) in round-robin". That is the AZ agnostic strategy and it is a different thing. The new one is AZ scoped, so it needs its own member, `AzAffinityAllNodes`, and the two must not be conflated in the enum, the FFI mapping, or the docs.

### Prerequisite

The `valkey-glide` submodule is currently pinned at `9a268ea`, which predates the merge of valkey-io/valkey-glide#6721. It has to be bumped to a revision that carries `AZAffinityAllNodes = 6` in `glide-core/src/protobuf/connection_request.proto` before the C# side can select the strategy.

### Checklist

- [ ] Bump the `valkey-glide` submodule to a revision containing valkey-io/valkey-glide#6721.
- [ ] Add `AzAffinityAllNodes` to `ConnectionConfiguration.ReadFromStrategy` in `sources/Valkey.Glide/ConnectionConfiguration.cs`, with a doc comment describing the round robin behaviour and the fallback.
- [ ] Add the matching variant to `ReadFromStrategy` in `rust/src/enums.rs`. That enum carries a comment requiring it to match `glide_core::client::ReadFrom`, so keep the discriminants in lockstep rather than assuming they follow the protobuf numbering.
- [ ] Update `sources/Valkey.Glide/Internals/FFI.enums.cs` and the conversion in `rust/src/ffi.rs` so the new value crosses the boundary.
- [ ] Extend the helper in `ConnectionConfiguration.cs` that decides whether a strategy requires an AZ. It currently matches only `AzAffinity` and `AzAffinityReplicasAndPrimary`, so the new strategy would silently skip the `ClientAz` requirement.
- [ ] Extend any read-only mode incompatibility check to cover the new strategy.
- [ ] Add tests. `tests/Valkey.Glide.UnitTests/ReadFromTests.cs` and `ConnectionMultiplexerReadFromMappingTests.cs` cover the mapping, and `tests/Valkey.Glide.IntegrationTests/AzAffinityTests.cs` covers behaviour against a real topology. All three have unit and integration counterparts that enumerate the existing strategies.

### Additional Notes

The Python diff in valkey-io/valkey-glide#6721 is a close reference for the binding surface, including the exact wording of the validation errors and the docstrings.

The same parity gap in the other clients:

- Java valkey-io/valkey-glide#6790
- Node valkey-io/valkey-glide#6791
- Go valkey-io/valkey-glide#6792
- PHP valkey-io/valkey-glide-php#316
- Ruby valkey-io/valkey-glide-ruby#294

prateek-kumar-improving:
## Summary

Adds the `AzAffinityAllNodes` read strategy to the C# client, bringing it to parity with the Python, Java, and Go bindings. The strategy spreads read requests equally, in round robin, across all nodes (primary and replicas) within the client's Availability Zone (AZ), falling back to a round robin across all nodes when no in-AZ node is available.

Unlike `AzAffinityReplicasAndPrimary`, it does not prioritize replicas ahead of the primary within the AZ, which is what enables an even per-node read distribution — useful on ElastiCache deployments that hit the replica cap.

## Issue Link

Closes #533.

The core routing, the vendored `redis-rs` changes, and the `AZAffinityAllNodes` proto value were already merged on the upstream `main` (valkey-io/valkey-glide#6721); this PR only wires up the C# binding surface.

- New public `ReadFromStrategy.AzAffinityAllNodes` value, usable both programmatically (`new ReadFrom(ReadFromStrategy.AzAffinityAllNodes, az)`) and via connection strings (`readFrom=AzAffinityAllNodes,az=...`).
- Like the other AZ strategies, it requires an Availability Zone; constructing it without one throws `ArgumentException`.

## Implementation

- **`ConnectionConfiguration.cs`**: added `AzAffinityAllNodes` to the `ReadFromStrategy` enum (value `5`, appended so all existing values are preserved) with XML docs, and included it in `IsAzReadFromStrategy()` so the AZ requirement is enforced by the `ReadFrom` constructors. Connection-string parsing works automatically via `Enum.Parse`.
- **Rust FFI (`rust/src/enums.rs`, `rust/src/ffi.rs`)**: unlike the protobuf-based bindings (Node/PHP/Java/Go), this repo has its own `rust/` crate that marshals the read-from strategy as a raw `#[repr(u32)]` enum. Added `AZAffinityAllNodes = 5` and the match arm mapping it to `glide_core::client::ReadFrom::AZAffinityAllNodes(az)`.

Reviewers: the main area worth attention is the FFI enum value alignment — the C# enum value and the Rust `ReadFromStrategy` value must stay in sync (`5` in both).

## Limitations

None. The read-only-mode incompatibility with AZ strategies is enforced by glide-core during client creation, consistent with the other AZ strategies.

## Testing

- **Unit tests** (`ReadFromTests`, `ConnectionMultiplexerReadFromMappingTests`): parse with/without AZ, empty/whitespace-AZ rejection, `ToString` formatting, round-trip serialization, property-setter validation, standalone + cluster config mapping, the all-strategies theory, and a `ToFfi()` flow test exercising native marshaling.
- **Integration tests** (`ConnectionMultiplexerReadFromMappingTests`, `ReadFromTests`, `AzAffinityTests`): standalone + cluster mapping, connection-string-to-FFI flow, and two end-to-end routing tests mirroring the Go PR — reads split evenly across the in-AZ primary and replica, and fallback distributes across every shard node when no in-AZ node exists (skipped on server < 8.0.0).
- The Rust crate builds cleanly (`cargo build`).

## Related Issues

- Tracking issue: #533
- Upstream Java reference: valkey-io/valkey-glide#7059
- Upstream Go reference: valkey-io/valkey-glide#6927
- Core/Python: valkey-io/valkey-glide#6721

## Checklist


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Support additional configuration options:
   - All nodes read-from strategy (#207)
+  - AZ affinity across all nodes read-from strategy (#533)
```
jamesx-improving, line 14: `Az = az.Trim()` at `ConnectionConfiguration.cs:236` sits in the shared constructor, so it also changes `AzAffinity` and `AzAffinityReplicasAndPrimary`: `ReadFrom.Az` and `ToString()` now differ for a padded value, and a config that previously engaged AZ affinity while matching no node starts working.

That's a behaviour change to shipped API with nothing under `### Fixed`. It's a different gap from the Java/PHP one, incidentally — there the pre-existing AZ strategies newly throw without an AZ, whereas C# already threw.

rust/src/enums.rs
```diff
@@ -52,6 +52,7 @@ pub enum ReadFromStrategy {
     AZAffinity = 2,
     AZAffinityReplicasAndPrimary = 3,
     AllNodes = 4,
+    AZAffinityAllNodes = 5,
```
jamesx-improving, line 55: The "Must match" comment covers the Rust→core leg, which is name-based and exhaustively matched, so a submodule pin lacking the variant fails to compile.

The C#→Rust leg has none of that: two hand-maintained enums, no generated symbol, no parity test, and 4 of the 6 C# members take implicit values — so inserting one above `AzAffinityAllNodes` renumbers C# without touching this file, and the result is silent misrouting rather than a build error. Values are correct today, 5 on both sides.

jamesx-improving (approved):
This is the cleanest of the three bindings to ship this strategy: the padded-`clientAz` hole that a bot caught in Node and that PHP still ships is fixed here — the value that gets validated, the one that gets trimmed and the one that crosses the FFI are all the same one, for the two pre-existing AZ strategies too. The routing tests earn their keep as well, reading per-node `cmdstat_get:calls` with assertions that only a (2,2) split satisfies and that would also fail under `AzAffinityReplicasAndPrimary`.

Two non-blocking notes inline. The FFI values themselves are correct — C# `AzAffinityAllNodes` = 5 matches `rust/src/enums.rs:55`.

rust/src/enums.rs
currantw: I'm a bit confused about this comment. What exactly is it critiquing or suggesting? 🤔

sources/Valkey.Glide/ConnectionConfiguration.cs
```diff
@@ -226,7 +226,14 @@ public ReadFrom(ReadFromStrategy strategy, string az)
...
             Strategy = strategy;
-            Az = az;
+
+            // The core compares availability zones with exact equality and never trims, so a padded
+            // value such as "us-east-1a " (easily produced by an environment variable or file read)
+            // would engage the strategy, match no node, and silently fall back to spreading reads
+            // across all nodes. No real availability-zone name carries surrounding whitespace, so
+            // trimming here cannot break a value that works today, and keeps parity with the other
+            // GLIDE clients (e.g. Java's ConnectionManager.resolveClientAz).
+            Az = az.Trim();
```
currantw, line 236: What is the specific rationale for adding this? Do we do this with any other string inputs? Was there a specific failure that triggered it – or was it just compatibility with other GLIDE clients? If the latter, then can you please verify that the same behaviour is applied to _all_ client, and raise issues for any where the behaviour diverges?

Similarly, I think this comment can be simplified significantly. I think we can probably explain this in a sentence or two? As mentioned, I think we should generally avoid references to specific other GLIDE clients, and even more so to specific methods in other clients (which may change).

sources/Valkey.Glide/ConnectionConfiguration.cs
```diff
@@ -260,6 +267,27 @@ public enum ReadFromStrategy : uint
...
+        /// <para>
+        /// Choose this over <see cref="AzAffinityReplicasAndPrimary"/> when the in-AZ primary should
+        /// take a share of the reads. That strategy sends every read to an in-AZ replica whenever one is
+        /// connected, so the primary serves reads only as a fallback.
+        /// </para>
+        /// <para>
+        /// Requires an Availability Zone (AZ) to be set on the client configuration.
+        /// </para>
+        /// </summary>
+        AzAffinityAllNodes,
```
currantw, line 290: I think this is too verbose. The documentation for `ReadFromStrategy` already links to the Valkey GLIDE documentation website. I don't think it makes sense to duplicate all that information here. We should provide a short, concise description of the value (like the other enum values already do), and leave more detailed description to the website.


```suggestion
        /// <summary>
        /// Reads from all nodes within the client's Availability Zone (AZ) in a round-robin manner,
        /// falling back to other nodes if needed.
        /// </summary>
        AzAffinityAllNodes,
```

tests/Valkey.Glide.IntegrationTests/AzAffinityTests.cs
```diff
@@ -239,4 +239,132 @@ public async Task TestAzAffinityReplicasAndPrimaryRoutesToPrimary(ConnectionConf
             }
         }
     }
+
+    [Theory]
+    [InlineData(ConnectionConfiguration.Protocol.RESP2)]
+    [InlineData(ConnectionConfiguration.Protocol.RESP3)]
+    public async Task TestAzAffinityAllNodesSplitsBetweenPrimaryAndReplica(ConnectionConfiguration.Protocol protocol)
+    {
+        Assert.SkipWhen(TestConfiguration.IsVersionLessThan("8.0.0"), "AZ affinity requires server version 8.0.0 or higher");
```
currantw, line 248: There is a `TestUtils` module for skipping. Can you please add a method for this case, use it here, and also update all the other methods in this class to use it?

tests/Valkey.Glide.IntegrationTests/AzAffinityTests.cs
```diff
@@ -239,4 +239,132 @@ public async Task TestAzAffinityReplicasAndPrimaryRoutesToPrimary(ConnectionConf
             }
         }
     }
+
+    [Theory]
+    [InlineData(ConnectionConfiguration.Protocol.RESP2)]
+    [InlineData(ConnectionConfiguration.Protocol.RESP3)]
```
currantw, line 245: There should be a data constant for this in `TestUtils.Data`. If there isn't, please add one, and update test in this file to use it.

tests/Valkey.Glide.IntegrationTests/ConnectionMultiplexerReadFromMappingTests.cs
```diff
@@ -105,6 +105,52 @@ public async Task ConfigurationOptions_ReadFromAzAffinityReplicasAndPrimary_Maps
...
+        // Assert
+        Assert.NotNull(connectionMultiplexer);
+        Assert.NotNull(connectionMultiplexer.RawConfig);
+        Assert.True(connectionMultiplexer.RawConfig.ReadFrom.HasValue);
+        Assert.Equal(ReadFromStrategy.AzAffinityAllNodes, connectionMultiplexer.RawConfig.ReadFrom.Value.Strategy);
+        Assert.Equal(testAz, connectionMultiplexer.RawConfig.ReadFrom.Value.Az);
+    }
+
+    [Fact]
+    public async Task ConfigurationOptions_ReadFromAzAffinityAllNodes_MapsToClusterClientConfigurationBuilder()
```
currantw, line 132: None of the tests above have cluster versions. Why are we doing it for this one? Should we add them for all other tests above? It seems like the only difference between `ConfigurationOptions_ReadFromAzAffinityAllNodes_MapsToStandaloneClientConfigurationBuilder` and `ConfigurationOptions_ReadFromAzAffinityAllNodes_MapsToClusterClientConfigurationBuilder` is the address. If we want to keep the latter, can we parametrize the test? 🤔

tests/Valkey.Glide.IntegrationTests/ConnectionMultiplexerReadFromMappingTests.cs
```diff
@@ -306,6 +353,7 @@ public async Task EndToEnd_ReadFromConfiguration_FlowsFromConnectionStringToConn
             ReadFromStrategy.AllNodes => ",readFrom=AllNodes",
             ReadFromStrategy.AzAffinity => $",readFrom=AzAffinity,az={az}",
             ReadFromStrategy.AzAffinityReplicasAndPrimary => $",readFrom=AzAffinityReplicasAndPrimary,az={az}",
+            ReadFromStrategy.AzAffinityAllNodes => $",readFrom=AzAffinityAllNodes,az={az}",
             _ => throw new ArgumentException("Invalid ReadFromStrategy for this test"),
         };
```
currantw, line 358: Instead of parametrizing on the ReadFromStrategy and az string, and THEN using a `switch` statement to map this to a connection string, why don't we simply parametrize on the connection string segment that we want to use?

tests/Valkey.Glide.IntegrationTests/ReadFromTests.cs
```diff
@@ -266,5 +269,28 @@ public async Task BackwardCompatibility_LegacyConfigurationWithoutReadFromStillW
...
+    [InlineData(ReadFromStrategy.AzAffinity)]
+    [InlineData(ReadFromStrategy.AzAffinityReplicasAndPrimary)]
+    [InlineData(ReadFromStrategy.AzAffinityAllNodes)]
+    public async Task ReadOnlyMode_WithAzAffinityStrategy_IsRejectedAtClientCreation(ReadFromStrategy strategy)
+    {
+        // Read-only mode is incompatible with every AZ-affinity read strategy. The rejection is
+        // enforced by the Rust core during (standalone) client creation and surfaces as a
+        // ConnectionException. ReadOnly has no public builder method yet, so it is set through the
+        // internal config, consistent with how the field is wired to the FFI.
+        StandaloneClientConfigurationBuilder builder = TestConfiguration.DefaultClientConfig()
```
currantw, line 285: (1)

- I don't think this level of detail is necessary in a test method. To me, the bigger gap is whether this is *clearly documented* for customers. Is it? Is this behaviour clearly documented *in the code itself*? That is where we should be explaining this sort of detail, I think, if needed? 🤷 

(2)

> ReadOnly has no public builder method yet

Why is this? Is this tracked in an issue? 

>  enforced by the Rust core during (standalone) client creation

Why is this only applicable to standalone clients?

tests/Valkey.Glide.UnitTests/ConnectionMultiplexerReadFromMappingTests.cs
```diff
@@ -90,6 +90,27 @@ public void CreateClientConfigBuilder_WithReadFromAzAffinityReplicasAndPrimary_M
         Assert.Equal("eu-west-1b", clusterConfig.Request.ReadFrom!.Value.Az);
     }
 
+    [Fact]
+    public void CreateClientConfigBuilder_WithReadFromAzAffinityAllNodes_MapsCorrectly()
```
currantw, line 94: This seems very similar to the test method above. Can we parametrize these and combine into a single test method(s)?

tests/Valkey.Glide.UnitTests/ConnectionMultiplexerReadFromMappingTests.cs
```diff
@@ -149,12 +170,33 @@ public void CreateClientConfigBuilder_ReadFromFlowsToFfiLayer()
         Assert.Null(connectionConfig.ReadFrom.Value.Az);
     }
 
+    [Fact]
+    public void CreateClientConfigBuilder_AzAffinityAllNodesFlowsToFfiLayer()
```
currantw, line 174: Similar to previous comment. This seems virtually identical to the method above. Can we reduce duplication with parametrization?

tests/Valkey.Glide.UnitTests/ReadFromTests.cs
```diff
@@ -235,6 +239,18 @@ public void ToString_WithAzAffinityReplicasAndPrimaryStrategy_IncludesCorrectAzF
         Assert.Contains($"az={azValue}", result);
     }
 
+    [Theory]
+    [InlineData("us-east-1a")]
+    [InlineData("ap-south-1c")]
+    public void ToString_WithAzAffinityAllNodesStrategy_IncludesCorrectAzFormat(string azValue)
```
currantw, line 245: This looks very similar to the method(s) above. Can we parametrize these tests?

tests/Valkey.Glide.UnitTests/ReadFromTests.cs
```diff
@@ -440,6 +457,72 @@ public void ReadFromProperty_SetValidAzAffinityReplicasAndPrimaryStrategy_DoesNo
         Assert.Equal("eu-west-1", options.ReadFrom.Value.Az);
     }
 
+    [Fact]
+    public void ReadFromProperty_SetValidAzAffinityAllNodesStrategy_DoesNotThrow()
```
currantw, line 461: Very similar to test method(s) above. Can we parametrize to reduce duplication?

CHANGELOG.md
currantw: Agreed. If we want to keep this, we should add an entry to Fixed or Changed.

tests/Valkey.Glide.IntegrationTests/AzAffinityTests.cs
```diff
@@ -239,4 +239,132 @@ public async Task TestAzAffinityReplicasAndPrimaryRoutesToPrimary(ConnectionConf
...
+    {
+        Assert.SkipWhen(TestConfiguration.IsVersionLessThan("8.0.0"), "AZ affinity requires server version 8.0.0 or higher");
+
+        await using GlideClusterClient configClient = await GlideClusterClient.CreateClient(
+            TestConfiguration.DefaultClusterClientConfig().WithProtocolVersion(protocol).Build());
+        string key = Guid.NewGuid().ToString();
+
+        // Clear any AZ so that "non-existing-az" matches nothing, triggering the all-nodes fallback.
+        _ = await configClient.CustomCommand(["config", "set", "availability-zone", ""], AllNodes);
+        _ = await configClient.CustomCommand(["config", "resetstat"], AllNodes);
```
currantw, line 318: Can we extract this or similar logic to a helper?

tests/Valkey.Glide.IntegrationTests/AzAffinityTests.cs
```diff
@@ -239,4 +239,132 @@ public async Task TestAzAffinityReplicasAndPrimaryRoutesToPrimary(ConnectionConf
...
+        int callsPerNode = nGetCalls / nodesInSameAz;
+        string key = Guid.NewGuid().ToString();
+
+        // Reset stats and place every node outside the client's AZ...
+        _ = await configClient.CustomCommand(["config", "resetstat"], AllNodes);
+        _ = await configClient.CustomCommand(["config", "set", "availability-zone", otherAz], AllNodes);
+
+        // ...then move exactly the primary and replica owning the key into the client's AZ.
+        _ = await configClient.CustomCommand(["config", "set", "availability-zone", az], new SlotKeyRoute(key, SlotType.Primary));
+        _ = await configClient.CustomCommand(["config", "set", "availability-zone", az], new SlotKeyRoute(key, SlotType.Replica));
```
currantw, line 265: Can we extract helper(s) for this? Might help us also simplify tests above?

currantw (commented):
Reviewed to 76536d8.

CHANGELOG.md
prateek-kumar-improving: Fixed in https://github.com/valkey-io/valkey-glide-csharp/pull/564/commits/10d08e31c99e5c5bd80ab09e5946d921a71e2afe — added a `### Changed` entry documenting that `ReadFrom.Az` is now trimmed of surrounding whitespace, which affects the pre-existing `AzAffinity` and `AzAffinityReplicasAndPrimary` strategies as well.

prateek-kumar-improving: Fixed in https://github.com/valkey-io/valkey-glide-csharp/pull/564/commits/10d08e31c99e5c5bd80ab09e5946d921a71e2afe — added a `### Changed` entry documenting that `ReadFrom.Az` is now trimmed of surrounding whitespace, which affects the pre-existing `AzAffinity` and `AzAffinityReplicasAndPrimary` strategies as well.

rust/src/enums.rs
prateek-kumar-improving: Good catch — the C#→Rust leg had no guard: two hand-maintained enums, no generated symbol, and the C# members used implicit values, so inserting one above `AzAffinityAllNodes` would silently renumber the C# side and misroute rather than fail the build.

Fixed in https://github.com/valkey-io/valkey-glide-csharp/pull/564/commits/ea45601bd639f7a8af5bc8220a43ede03ea80166: the C# `ReadFromStrategy` now assigns explicit values (`Primary = 0` … `AzAffinityAllNodes = 5`), a `<remarks>` on the enum documents that the values are part of the FFI contract and must match `rust/src/enums.rs`, and the Rust comment now points back at the C# enum too. So a future insertion can't renumber the existing members silently.

prateek-kumar-improving: @currantw in short: the comment was flagging that only the Rust→core direction was protected (by the exhaustive, name-based match, which fails to compile if the variant is missing). The C#→Rust direction had no such protection — it relied on both enums happening to stay numerically aligned, and the C# side used implicit values.

Fixed in https://github.com/valkey-io/valkey-glide-csharp/pull/564/commits/ea45601bd639f7a8af5bc8220a43ede03ea80166 by giving the C# `ReadFromStrategy` explicit values and documenting the FFI contract on both sides, so a future insertion can't silently renumber and misroute.

sources/Valkey.Glide/ConnectionConfiguration.cs
prateek-kumar-improving: The trim is motivated by a real correctness gap, not just parity: glide-core compares availability-zone names by exact equality, so a padded value like `"us-east-1a "` passes the non-empty check, engages the strategy, matches no node, and silently spreads reads across all nodes. Trimming ensures the value that's validated is the one the core matches on.

Simplified the comment in https://github.com/valkey-io/valkey-glide-csharp/pull/564/commits/ea45601bd639f7a8af5bc8220a43ede03ea80166 and removed the reference to other clients' internals. We don't trim other string inputs — this is specific to AZ because of the exact-match routing. Node and Java already trim the AZ, so this also keeps C# consistent with them.

sources/Valkey.Glide/ConnectionConfiguration.cs
prateek-kumar-improving: Agreed. Applied your suggested wording in https://github.com/valkey-io/valkey-glide-csharp/pull/564/commits/ea45601bd639f7a8af5bc8220a43ede03ea80166 — the doc is now the concise two-line description, matching the other enum values, with the detailed explanation left to the linked docs site.

tests/Valkey.Glide.UnitTests/ConnectionMultiplexerReadFromMappingTests.cs
prateek-kumar-improving: Fixed in https://github.com/valkey-io/valkey-glide-csharp/pull/564/commits/1d12dacd8dba29d9a4f4c3634f0b0e72a08a18c0. Removed the five per-strategy `..._MapsCorrectly` `[Fact]`s entirely — they were already fully covered by the existing `CreateClientConfigBuilder_AllReadFromStrategies_MapCorrectly` `[Theory]`, which asserts both the standalone and cluster mappings across every strategy. Kept `WithNullReadFrom_HandlesCorrectly` since the theory doesn't cover the null case.

tests/Valkey.Glide.UnitTests/ConnectionMultiplexerReadFromMappingTests.cs
prateek-kumar-improving: Fixed in https://github.com/valkey-io/valkey-glide-csharp/pull/564/commits/1d12dacd8dba29d9a4f4c3634f0b0e72a08a18c0. Merged the two `...FlowsToFfiLayer` tests into a single `[Theory] CreateClientConfigBuilder_ReadFromFlowsToFfiLayer(strategy, az)` parametrized on `(strategy, az)`.

tests/Valkey.Glide.UnitTests/ReadFromTests.cs
prateek-kumar-improving: Fixed in https://github.com/valkey-io/valkey-glide-csharp/pull/564/commits/1d12dacd8dba29d9a4f4c3634f0b0e72a08a18c0. Collapsed the five `ReadFromProperty_SetValid{Strategy}_DoesNotThrow` `[Fact]`s into one `[Theory] ReadFromProperty_SetValidStrategy_DoesNotThrow(strategy, az)` covering all six strategies.

tests/Valkey.Glide.UnitTests/ReadFromTests.cs
prateek-kumar-improving: Fixed in https://github.com/valkey-io/valkey-glide-csharp/pull/564/commits/1d12dacd8dba29d9a4f4c3634f0b0e72a08a18c0 — this was part of the same duplicate group as the comment above; the five per-strategy Facts are now a single parametrized `[Theory]`.

tests/Valkey.Glide.IntegrationTests/ConnectionMultiplexerReadFromMappingTests.cs
prateek-kumar-improving: Fixed in https://github.com/valkey-io/valkey-glide-csharp/pull/564/commits/1d12dacd8dba29d9a4f4c3634f0b0e72a08a18c0. The theory now takes the `readFrom=...,az=...` connection-string segment directly as `InlineData` and appends it, so the `switch` is gone.

tests/Valkey.Glide.IntegrationTests/ConnectionMultiplexerReadFromMappingTests.cs
prateek-kumar-improving: This actually follows the existing pattern in this file — each strategy already has a matching standalone/cluster pair (`ConfigurationOptions_ReadFromPrimary_MapsToStandaloneClientConfigurationBuilder` / `...MapsToClusterClientConfigurationBuilder`, and likewise for `PreferReplica`, `AzAffinity`, and `AzAffinityReplicasAndPrimary`), just further down the file. The AzAffinityAllNodes standalone + cluster pair mirrors that, so dropping the cluster one would make it the only strategy without cluster coverage.

You're right that the only real difference is the address, so these are good candidates to parametrize into `[Theory]` cases (standalone + cluster × strategy). Note that would refactor the existing tests in this file, not just the AzAffinityAllNodes ones. Let me know if you'd rather I do the parametrization here.

tests/Valkey.Glide.IntegrationTests/AzAffinityTests.cs
prateek-kumar-improving: Fixed in https://github.com/valkey-io/valkey-glide-csharp/pull/564/commits/1a723db3b013f38b0223b83ddd1e49fdb0cbd927. Added `Skip.IfAzAffinityNotSupported()` to `Skip.cs` and replaced the inline `Assert.SkipWhen(TestConfiguration.IsVersionLessThan("8.0.0"), ...)` in all the tests in this class with it.

tests/Valkey.Glide.IntegrationTests/AzAffinityTests.cs
prateek-kumar-improving: Fixed in https://github.com/valkey-io/valkey-glide-csharp/pull/564/commits/1a723db3b013f38b0223b83ddd1e49fdb0cbd927. Added `Data.AvailabilityZone`, `Data.OtherAvailabilityZone`, and `Data.NonExistingAvailabilityZone` to `TestUtils.Data`, and replaced the hardcoded `"us-east-1a"` / `"us-east-1b"` / `"non-existing-az"` literals in this file with them.

tests/Valkey.Glide.IntegrationTests/AzAffinityTests.cs
prateek-kumar-improving: Fixed in https://github.com/valkey-io/valkey-glide-csharp/pull/564/commits/1a723db3b013f38b0223b83ddd1e49fdb0cbd927. Extracted a `GetCalls(infoValue)` helper (backed by a compiled regex) that parses `cmdstat_get:calls` from a node's INFO output, and reused it across all the tests in place of the repeated `Regex.Match(...)` blocks.

tests/Valkey.Glide.IntegrationTests/AzAffinityTests.cs
prateek-kumar-improving: Fixed in https://github.com/valkey-io/valkey-glide-csharp/pull/564/commits/1a723db3b013f38b0223b83ddd1e49fdb0cbd927. Extracted `SetAvailabilityZone(client, az, route)`, `ResetStats(client)`, and `CreateConfigClient(protocol)` helpers and applied them across the class — this removed the duplicated `config set availability-zone` / `config resetstat` command literals and the repeated config-client setup in the tests above as well.

tests/Valkey.Glide.IntegrationTests/ReadFromTests.cs
prateek-kumar-improving: Fixed in https://github.com/valkey-io/valkey-glide-csharp/pull/564/commits/e0024720e6a70ec2b8d588b19b5c97e9fc20f2c4. Addressing your points:

**(1) Documentation for customers** — it wasn't documented in code before. Added a `<remarks>` on the public `ReadFrom` type noting that the AZ-affinity strategies are incompatible with read-only mode and the combination is rejected at client creation, and trimmed the test comment down to a single line now that the behavior is documented where customers see it.

**(2) Why no public `ReadOnly` builder** — read-only mode simply isn't exposed through the public builder API yet in the C# client; the field exists internally and is wired to the FFI, but there's no `WithReadOnly(...)`. The test sets it via the internal config for that reason. I've documented this on the internal field.

**(3) Why standalone-only** — the read-only + AZ-affinity rejection lives in glide-core's standalone client path (`standalone_client.rs`); the cluster client doesn't perform the same check. So the test targets a `GlideClient` (standalone), which is where the error is actually raised.

sources/Valkey.Glide/ConnectionConfiguration.cs
currantw: Did you "verify that the same behaviour is applied to all client"? If missing for any others, please raise corresponding consistency issues.

tests/Valkey.Glide.IntegrationTests/AzAffinityTests.cs
currantw: Hmmmm. This comment was specifically about RESP protocols. I think it makes sense to add these to `TestUtils.Data`, since this is useful/applicable across many tests. I'm not sure whether the same is true for Availability Zones. I don't imagine there are any other tests that need to or might want to use these constants?

If the above is correct, I would suggest:

- Add `TestUtils.Data` for RESP protocols
- Remove `TestUtils.Data` for Availability Zones.

tests/Valkey.Glide.IntegrationTests/ConnectionMultiplexerReadFromMappingTests.cs
currantw: I see. They just aren't grouped together! That makes sense. 

I think it makes sense to do the parametrization here. It is a trivial change that will reduce probably 100+ lines of duplicate code, but take only a few minutes to implement. Seems worthwhile to me!

tests/Valkey.Glide.IntegrationTests/ReadFromTests.cs
currantw: (2) Can you please dig into this more? 

For example:

- What issue/pull request added this field? Does it have any rationale for this?
- Is there any open issue for adding this? Is it tracked in any way?

(3)

> the cluster client doesn't perform the same check

Why not? What behaviour does the cluster client have instead? If this is an inconsistency, should we raise an issue to fix it?

sources/Valkey.Glide/ConnectionConfiguration.cs
```diff
@@ -176,6 +182,12 @@ public struct RetryStrategy(uint numberOfRetries, uint factor, uint exponentBase
     /// <summary>
     /// Represents the client's read from strategy and Availability zone if applicable.
     /// </summary>
+    /// <remarks>
+    /// The AZ-affinity strategies (<see cref="ReadFromStrategy.AzAffinity"/>,
+    /// <see cref="ReadFromStrategy.AzAffinityReplicasAndPrimary"/>, and
+    /// <see cref="ReadFromStrategy.AzAffinityAllNodes"/>) are not compatible with read-only mode;
+    /// combining them is rejected when the client is created.
+    /// </remarks>
```
currantw: I don't think it makes sense to list all the strategies here, as this list is likely to get out of date when a new strategy is added. Let's just link to the enum if needed.

tests/Valkey.Glide.UnitTests/ReadFromTests.cs
currantw: This also looks virtually identical to the method(s) above. Can we parametrize these to reduce duplication?

tests/Valkey.Glide.UnitTests/ReadFromTests.cs
```diff
@@ -388,56 +405,75 @@ public void RoundTrip_LegacyConnectionString_RemainsCompatible()
...
-        Assert.Equal(ReadFromStrategy.AzAffinity, options.ReadFrom.Value.Strategy);
-        Assert.Equal("us-east-1", options.ReadFrom.Value.Az);
+        _ = Assert.NotNull(options.ReadFrom);
+        Assert.Equal(expectedStrategy, options.ReadFrom.Value.Strategy);
+        Assert.Equal(expectedAz, options.ReadFrom.Value.Az);
     }
 
     [Fact]
-    public void ReadFromProperty_SetValidAzAffinityReplicasAndPrimaryStrategy_DoesNotThrow()
+    public void ReadFromProperty_SetAzAffinityAllNodesWithoutAz_ThrowsArgumentException()
```
currantw: Shouldn't we have a similar test case to this for *all* the AZ affinity types? Can we parametrize this for all of them?

tests/Valkey.Glide.UnitTests/ReadFromTests.cs
```diff
@@ -388,56 +405,75 @@ public void RoundTrip_LegacyConnectionString_RemainsCompatible()
...
+            => new ReadFrom(ReadFromStrategy.AzAffinityAllNodes));
+        Assert.Contains("Availability zone must be specified", exception.Message);
+    }
+
+    [Theory]
+    [InlineData("")]
+    [InlineData(" ")]
+    [InlineData("\t")]
+    [InlineData("\n")]
+    public void ReadFromProperty_SetAzAffinityAllNodesWithEmptyOrWhitespaceAz_ThrowsArgumentException(string azValue)
```
currantw: Shouldn't we have a similar test case to this for *all* the AZ affinity types? Can we parametrize this for all of them?

tests/Valkey.Glide.UnitTests/ReadFromTests.cs
```diff
@@ -608,10 +644,12 @@ public void ReadFromProperty_NullValue_DoesNotAffectClone()
     [InlineData("", "", "cannot be empty")]
     [InlineData("AzAffinity", "", "Availability zone cannot be empty or whitespace")]
     [InlineData("AzAffinityReplicasAndPrimary", "", "Availability zone cannot be empty or whitespace")]
+    [InlineData("AzAffinityAllNodes", "", "Availability zone cannot be empty or whitespace")]
     [InlineData("AzAffinity", "   ", "Availability zone cannot be empty or whitespace")]
     [InlineData("AzAffinity", "\t", "Availability zone cannot be empty or whitespace")]
     [InlineData("AzAffinity", "\n", "Availability zone cannot be empty or whitespace")]
     [InlineData("AzAffinityReplicasAndPrimary", "   ", "Availability zone cannot be empty or whitespace")]
+    [InlineData("AzAffinityAllNodes", "   ", "Availability zone cannot be empty or whitespace")]
     public async Task ConnectionString_ArgumentExceptionScenarios(string readFromStrategy, string azValue, string expectedErrorSubstring)
```
currantw, line 572: We added `ReadFromProperty_SetAzAffinityAllNodesWithEmptyOrWhitespaceAz_ThrowsArgumentException` above. Is that test necessary then? It seems like we are testing the same thing here, no?

currantw (commented):
Reviewed to e002472.

sources/Valkey.Glide/ConnectionConfiguration.cs
prateek-kumar-improving: Checked the other clients:
- **Node** and **Java** trim the client AZ (same as this PR).
- **PHP** also guards against padded AZ values, but by **rejecting** them rather than trimming: `valkey_glide.c` throws `"client_az must not have leading or trailing whitespace."` for a padded value, treats empty/whitespace-only as absent.

So none of the clients forward an untrimmed AZ.

tests/Valkey.Glide.IntegrationTests/AzAffinityTests.cs
prateek-kumar-improving: Fixed in https://github.com/valkey-io/valkey-glide-csharp/pull/564/commits/3d74b7625fc34c5730ffde5c00901533d15757d1. The AZ values aren't reused broadly, so I removed the AZ constants from `TestUtils.Data` and inlined them as local `const`s in `AzAffinityTests`. Added `Data.Protocols` (RESP2/RESP3) instead and wired the theories here to `[MemberData(nameof(Data.Protocols), MemberType = typeof(Data))]`, since that's the cross-cutting one.

tests/Valkey.Glide.IntegrationTests/ConnectionMultiplexerReadFromMappingTests.cs
prateek-kumar-improving: Done in https://github.com/valkey-io/valkey-glide-csharp/pull/564/commits/3d74b7625fc34c5730ffde5c00901533d15757d1. Collapsed the standalone/cluster per-strategy `[Fact]`s into parametrized `[Theory]`s (on `useStandalone` + strategy + az), plus the null-ReadFrom and connection-string paths — removes the bulk of the duplication you flagged.

tests/Valkey.Glide.IntegrationTests/ReadFromTests.cs
prateek-kumar-improving: (2) Could not find more information about this.

(3) The read-only + AZ-affinity rejection lives only in glide-core's standalone client (`standalone_client.rs`); the cluster path has no equivalent check — that looks like a core-level inconsistency rather than a binding one.

sources/Valkey.Glide/ConnectionConfiguration.cs
prateek-kumar-improving: Fixed in https://github.com/valkey-io/valkey-glide-csharp/pull/564/commits/3d74b7625fc34c5730ffde5c00901533d15757d1 — the `<remarks>` now just links to the `ReadFromStrategy` enum instead of enumerating the strategies, so it won't go stale when a new one is added.

tests/Valkey.Glide.UnitTests/ReadFromTests.cs
prateek-kumar-improving: Fixed in https://github.com/valkey-io/valkey-glide-csharp/pull/564/commits/3d74b7625fc34c5730ffde5c00901533d15757d1. Removed the three redundant per-strategy `ToString_WithAzAffinity*_IncludesCorrectAzFormat` theories — they're already covered by the parametrized `ToString_WithReadFromStrategyWithAz_IncludesCorrectFormat`.

tests/Valkey.Glide.UnitTests/ReadFromTests.cs
prateek-kumar-improving: Fixed in https://github.com/valkey-io/valkey-glide-csharp/pull/564/commits/3d74b7625fc34c5730ffde5c00901533d15757d1. Parametrized into `ReadFromProperty_SetAzStrategyWithoutAz_ThrowsArgumentException` covering all three AZ strategies.

tests/Valkey.Glide.UnitTests/ReadFromTests.cs
prateek-kumar-improving: Fixed in https://github.com/valkey-io/valkey-glide-csharp/pull/564/commits/3d74b7625fc34c5730ffde5c00901533d15757d1. Parametrized into `ReadFromProperty_SetAzStrategyWithEmptyOrWhitespaceAz_ThrowsArgumentException` covering all three AZ strategies, and removed the per-strategy duplicates.

tests/Valkey.Glide.UnitTests/ReadFromTests.cs
prateek-kumar-improving: Fixed in https://github.com/valkey-io/valkey-glide-csharp/pull/564/commits/3d74b7625fc34c5730ffde5c00901533d15757d1. Consolidated — the AzAffinityAllNodes-only empty/whitespace test is now part of the parametrized `SetAzStrategyWithEmptyOrWhitespaceAz` theory, so the duplication is gone.

sources/Valkey.Glide/ConnectionConfiguration.cs
currantw: If we are going to be trimming and want to be consistent, why not do it in GLIDE core then?

prateek-kumar-improving: It could work in glide-core, but having this logic at the language binding level is helpful in dealing with it differently for different languages, like an exception is thrown in PHP. Different bindings might want to have different logic for this.

Also, JAVA and Python have already implemented it this way.

yipin-chen (approved):

sources/Valkey.Glide/ConnectionConfiguration.cs
currantw: It seems to me like this is exactly the sort of shared logic that we have decided as a team should be centralized in `glide-core`, precisely to prevent inconsistencies that result from each client "dealing with it differently".

My recommendation is to raise an issue to make this change in `glide-core`, and to update the other clients (including this one) for consistency. It should be a very simple change that shouldn't take more than a couple hours to implement and test across all clients.

If you disagree, can you please start a discussion in Slack?

tests/Valkey.Glide.IntegrationTests/ReadFromTests.cs
currantw: (2) Did you try `git blame`? What issue/pull request were they added as part of? Do any other clients implement Read Only functionality?

(3) Can you please raise an issue to fix this inconsistency, please? It should include adding corresponding integration tests across all clients, including this one, I would think.

currantw (approved):

sources/Valkey.Glide/ConnectionConfiguration.cs
prateek-kumar-improving: Good point on consistency. Keeping the binding-level trim here for now (it matches Node/Java and is low-risk), and raising the broader "normalize/validate the AZ in glide-core" question as a separate cross-client discussion so we don't settle it unilaterally in this C# PR.

Issue created: https://github.com/valkey-io/valkey-glide/issues/7171

tests/Valkey.Glide.IntegrationTests/ReadFromTests.cs
prateek-kumar-improving: (2) `git blame` traced the `ReadOnly` field to #219 ("Bump submodule (read only) and improve test reliability"), which intentionally added FFI support only — its description notes *"Add FFI support for read only. Feature itself not yet supported for C# client."*
Issue created: https://github.com/valkey-io/valkey-glide-csharp/issues/569

(3) Dug into this and it's standalone-only **by design**, not an inconsistency to fix: read-only mode is a standalone-only feature across all of GLIDE. Node exposes `readOnly` only on `GlideClient` config, Java only on `GlideClientConfiguration` (its `ConnectionManager` sets it `if (standaloneConfig.isReadOnly())`), and glide-core consumes `read_only` only in `standalone_client.rs`. Cluster clients can't set read-only, so there's nothing to validate on that path — hence no issue needed here.

